PR #13, opened 2020-10-18
removed CommunityAuthor module

smcgregor:
Example code that is not used.


smcgregor:
I am going to close this and take a more complete trip through the code to clean up things that are not used or needed.

PR #66, opened 2021-01-07
Fix ESLint errors and update Readme

### LINKED ISSUE #36: Add programatic style guide enforcement
The current version of the website was developed in plain text editors without a linter or enforcement of a style guide. Future merge requests will be enforced more zealously and settling all arguments with a programmatic enforcer is the easiest path forward.

jt-mchorse:
Closes #36


site/gatsby-site/src/templates/cite.js
```diff
@@ -183,68 +225,64 @@ export default class IncidentCite extends Component {
...
-      },
-      site: {
-        siteMetadata: { docsLocation, title },
-      },
+      // allMdx,
+      allMongodbAiidprodIncidents: { group },
     } = data;
 
-    const gitHub = require('../components/images/github.svg');
+    // const navItems = allMdx.edges
```
smcgregor, line 232: What is going on here?

jt-mchorse: This var has never been used, I kept it for now to have your confirmation on removal.

smcgregor: Ok, if it was never read go ahead and remove it. Is this the case in some of the other linted files?

jt-mchorse: Yes

jt-mchorse: Removed.

site/gatsby-site/pages/apps/submit.js
```diff
@@ -96,214 +98,240 @@ const dateRegExp = /^(\+?\d{4})?\s?-?\s?(\(?\d{2}\)?)\s?-?\s?(\(?\d{2}\)?)\s?-?\
...
+    console.log('---------------------------');
+  };
 
   handleOnError = (err, file, inputElem, reason) => {
-    console.log(err)
-  }
+    console.log(err, file, inputElem, reason);
+  };
 
-  updateRelated = (incident_id) => {
```
smcgregor, line 274: parens were added in another case above. Are parens needed or not needed per the style guide?

jt-mchorse: https://prettier.io/docs/en/options.html#arrow-function-parentheses

We did not overwritten any rules, so it must be using always.

smcgregor: Why this change then? Should it not have parens around `incident_id`?

smcgregor: This is my only lingering doubt about this merge request, but I think I am good with landing this as soon as we close out this discussion.

jt-mchorse: It's our prettier version, "prettier": "^1.19.1", which is using "avoid" as default. shall we upgrade package version to latest?

smcgregor: We might as well be at the latest since changing this again at a later time will be annoying. Regardless, I don't feel strongly about it so long as we are consistent. Would you like to proceed as-is?

jt-mchorse: I've updated the package versions and fixed all lint issues.

PR #67, opened 2021-01-07
Bugfix/wordcounts page loads slow

### LINKED ISSUE #56: SSR of Word Counts Produces Bloat
The [wordcounts](https://incidentdatabase.ai/summaries/wordcounts) page currently has long load times, likely because the server side rendering of the page produces a giant blob that should be made more efficient. The load time of the page after SSR should be nearly immediate.

jt-mchorse:
It preprocess the data before the static HTML generation, so the component consumes processed data only and just renders it. it prevents the page collect data on client's browser and process it there. We do processing only once on page build and that fixes the problem, The only concern is, the data won't be updated until the next build is triggered, but fortunately, we have a cron job scheduled in every hour and this fixes the problem. user will have maximum 1 hour old data on this page.

Closes #56


site/gatsby-site/gatsby-node.js
```diff
@@ -50,6 +51,77 @@ exports.createPages = ({ graphql, actions }) => {
     );
   });
 
+  const wordCounts = new Promise((resolve, reject) => {
```
smcgregor, line 54: Can we put this inside the wordcount component, or some other codepoint close to it and then require the functionality into here? I don't want to bloat gatsby-node or make it more difficult to maintain the component. I also want to establish a design pattern where we can do this design pattern many times.

jt-mchorse: I agree to not make the gatsby-node giant, let me make few functions to do this outside. Give me 30 mins, I'll update the PR.

smcgregor: Sounds good, thanks!

jt-mchorse: I'm going to create a `page-creators` directory at the top level, and will put `mdx.js`, `citiation.js`, `wordCounts.js` which returns a function accept `createPage` and `graphql` as props and return a promise. 

so createPages will look like

```
const createMdxPages = require('./page-creators/createMdxPages');
const createCitiationPages = require('./page-creators/createCitiationPages');
const createWordCountsPages = require('./page-creators/createWordCountsPage');

exports.createPages = ({ graphql, actions }) => {

  const { createPage } = actions;

  return Promise.all([
    createMdxPages(graphql, createPage),
    createCitiationPages(graphql, createPage),
    createWordCountsPages(graphql, createPage),
  ])
};
```

smcgregor:
I'll finish review on this as soon as we land your linter MR and make the diff cleaner.

smcgregor:
It looks like you may need to merge master into this branch before the diff in the GitHub interface will show clean for the linted files.

smcgregor:
I am going to merge this tomorrow morning, so if you are working on anything else you can base from this. In general this is looking like a great optimization. I think the design can be improved in the following ways:

1. All the "page creators" could be discoverable by graphql (maybe, LMK if I am mistaken) and avoid needing to register them individually. This would simplify the developer story.
2. The pages that are summaries and should have their data pre-computed should likely all be in the same sub directory of the pages directory. You pulled them up into the pages directory, but I like having all the database summaries (and components that may need some form of pre-computation) in their own directory.

None of these are blocking merge. I just don't want to merge at EOD on a Friday and potentially break something on deploy. :)

PR #70, opened 2021-01-11
remove UUID key assignees

### LINKED ISSUE #68: Remove UUID library references
There are many references where I solved the unique key requirement of lists by generating UUIDs. While this technically works, I don't believe it is good practice.

jt-mchorse:
Closes #68

### DIFF
--- a/site/gatsby-site/pages/apps/submitted.js
+++ b/site/gatsby-site/pages/apps/submitted.js
@@ -14,8 +14,6 @@ import Badge from 'react-bootstrap/Badge';
 
 import BSON from 'bson';
 
-import uuid from 'react-uuid';
-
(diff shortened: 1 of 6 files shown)


site/gatsby-site/pages/apps/submitted.js
```diff
@@ -284,6 +280,7 @@ export const pageQuery = graphql`
     allMongodbAiidprodQuickadd {
       edges {
         node {
+          id
```
smcgregor, line 283: This makes a lot of sense...

smcgregor:
@jt-mchorse can you merge `master` in? I think this is good to go, but I want to check the diff one more time before hitting merge. Also, I think you may be able to remove the UUID lib, but I am not sure whether something else will unravel when doing so.

jt-mchorse:
We can merge it now @smcgregor .

PR #71, opened 2021-01-11
Improve citation page v1

jt-mchorse:


site/gatsby-site/src/utils/getCanonicalUrlUrl.js
```diff
@@ -0,0 +1,9 @@
+import config from '../../config';
```
smcgregor, line 1: UrlUrl (twice) in the file name?

jt-mchorse: Mybad.

site/gatsby-site/src/components/ImageCarousel.js
```diff
@@ -0,0 +1,52 @@
...
+            }
+            alt={value['node']['title']}
+          />
+          <Carousel.Caption>
+            <Caption>
+              <a
+                href={value['node']['url']}
+                className="text-white"
+                target="_blank"
+                rel="noopener noreferrer"
```
smcgregor, line 39: No need to change this now, but I like the idea of telling the publications that we are sending traffic to them.

Issue #74, opened 2021-01-19
Duplicate Incidents


smcgregor:
[Incident 90](https://incidentdatabase.ai/cite/90) is a duplicate of [Incident 77](https://incidentdatabase.ai/cite/77). I had hoped to address this problem later, but this was my mistake and we now have to have a path to ensure the following,

1) People know they should cite 77 and not 90
2) People should not submit incident reports against incident 90, but should instead submit against 77
3) Incident 90 should continue to exist because incident identifiers are permanent, but when people follow the citation to the web address it should point people to the underlying incident 77

Towards this I created a new collection on the database, `duplicates`, that contains only the following document.

New duplicates collection,
```
_id: 600642cd866915d3c34392af
duplicate_incident_number:90
true_incident_number:77
```

When generating the citation pages, this should generate a placeholder page with the contents,

```
This incident is a duplicate of Incident (link). All new reports and citations should be directed incident (number). The reports previously found on this page have been migrated to the previously existing incident.
```

The [page listing](https://incidentdatabase.ai/summaries/incidents) can skip the duplicative number, but we may need to improve this later.

There is also potentially an issue where incident identifiers are auto incremented based on the currently maximum incident identifier, but if the maximum was removed from the incidents collection it will be re-granted. I will ensure this does not happen in the immediate future and open a new issue once this change over takes place.

PR #77, opened 2021-01-20
Make queue of reported incidents editable.

### LINKED ISSUE #51: Make Editing Reports in Queue Possible
Some incident report submissions have minor errors that should be correctable in the queue by administrators. Adding better UI will make it much easier to ingest cleaned up data without needing super user permissions.

jt-mchorse:
Closes #51


site/gatsby-site/src/components/forms/IncidentForm.js
```diff
@@ -0,0 +1,535 @@
...
+    setCsvIndex(newCSVIndex);
+  };
+
+  const onSubmit = async (values, { setSubmitting, resetForm }) => {
+    // When button submits form and form is in the process of submitting, submit button is disabled
+    setSubmitting(true);
+
+    if (incident) {
+      // Do update
+      // await user.functions.updateReportInReview(values)
```
jt-mchorse, line 250: @smcgregor , here we update the incident, can you please prepare the updateFunction so I can make a call here?

jt-mchorse:
Closes #74

site/gatsby-site/.eslintrc.json
```diff
@@ -16,6 +16,13 @@
       "alias": {
         "map": [
           ["@components", "./src/components"],
+          ["components", "./src/components"],
```
smcgregor, line 19: What is the effect of listing/not listing here?

smcgregor:
@jt-mchorse Please keep unrelated change sets separate from one another. I see there are changes for #74 and #51 here, which means the changes associated with the critical issue #74 can't land until #51 is mergeable.

site/gatsby-site/jsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
```
smcgregor, line 2: What is consuming this config file? I don't see it referenced in this merge request.

smcgregor:
I whitelisted the preview domain's origin in the backend to do functional testing. I get an error here:

https://deploy-preview-77--aiid.netlify.app/apps/quickadd

where `Report successfully added to review queue. It will appear on the review queue page within an hour.` displays even when I have not submitted anything.

I am also not sure what functionality is currently expected to work. After whitelisting and adding my admin token to the app I am not able to press the pen and paper icon in the lower left of the review queue. Should it be pressable? I know you need API support before the content will update on the backend, but I thought I would get to an error on posting updated content rather than a problem with editing.

smcgregor:
Things generally look great to me in looking at the various refactorings. I'll spend a bit more time understanding what is happening in this organization, but the most important next step is to get the preview working.

site/gatsby-site/.eslintrc.json
jt-mchorse: it allows eslint to detect alias imports. otherwise, eslint will give module not found error on alias imports.

site/gatsby-site/jsconfig.json
jt-mchorse: it helps VSCode to detect & provide alias imports.

jt-mchorse:
React-boostrap's alert was showing up even though the `show` prop was falsy value. seems they do strict value check. Set the default value as false and problem resolved.

site/gatsby-site/jsconfig.json
smcgregor: You point VSCode at this file, or it already knows about it?

jt-mchorse: Please reference this [link](https://code.visualstudio.com/docs/languages/jsconfig#:~:text=The%20presence%20of%20jsconfig.,by%20the%20JavaScript%20language%20service.)

smcgregor:
The form is always hitting the new endpoint regardless of whether it is editing an already existing incident. You have the update endpoint in a PR on your fork. I also chatted an update to the update endpoint that is currently on production fixing an issue with this PR. You should add edits to the Mongodb.md documentation and I will deploy them with this PR. Both the update endpoint and the new endpoint will need to update for your changes.

smcgregor:
It is logging in twice on every navigation, once for the anon user and once for the token user.

jt-mchorse:
The login issue has been fixed in this [PR](https://github.com/PartnershipOnAI/aiid/pull/109). 

https://user-images.githubusercontent.com/73718225/112172070-9e863580-8bb1-11eb-9f96-ee8ccdc121bb.mov

jt-mchorse:
> The form is always hitting the new endpoint regardless of whether it is editing an already existing incident. You have the update endpoint in a PR on your fork. I also chatted an update to the update endpoint that is currently on production fixing an issue with this PR. You should add edits to the Mongodb.md documentation and I will deploy them with this PR. Both the update endpoint and the new endpoint will need to update for your changes.

I believe the further updates on #109 fixes the issue. please check recording. I'm using the `updateOne` of the realm app directly, so if we update the mongodb.md, I want to remove existing user functions and use realm methods directly, so instead of calling `user.functions.deleteSubmittedDocument`, I want to call realmApp.collection.deleteOne. what do you think about this approach?

https://user-images.githubusercontent.com/73718225/112174395-9cbd7180-8bb3-11eb-9838-d091703c5b40.mov

smcgregor:
When navigating around the site I am still seeing a huge number of re-auth requests being sent server side. Does the userContext not cache the state of authentication when navigating around to different pages? Why not auth once then never again generate a request?

smcgregor:
Closing in favor of: https://github.com/PartnershipOnAI/aiid/pull/100

PR #78, opened 2021-01-24
redirect duplicated incidents to origin

### LINKED ISSUE #74: Duplicate Incidents

jt-mchorse:
Closes #74

### DIFF
--- a/site/gatsby-site/gatsby-config.js
+++ b/site/gatsby-site/gatsby-config.js
@@ -1,12 +1,14 @@
-require("dotenv").config();
-const queries = require("./src/utils/algolia");
-const config = require("./config");
+require('dotenv').config();
+const queries = require('./src/utils/algolia');
+
+const config = require('./config');
+
 const plugins = [
   {
-      resolve: `gatsby-plugin-netlify`,
-      options: {
-          headers: {}
-      }
+    resolve: `gatsby-plugin-netlify`,
+    options: {
+      headers: {},
+    },
   },
   {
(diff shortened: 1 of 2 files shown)


smcgregor:
This looks good to me. I will migrate the duplicate report to the older incident record and merge this.

Issue #86, opened 2021-02-06
Update 404 Page


smcgregor:
The current 404 page for the site looks like the following,

<img width="741" alt="Screen Shot 2021-02-06 at 9 22 46 AM" src="https://user-images.githubusercontent.com/64780/107125176-2364fc00-685d-11eb-8edd-97db14114a14.png">

This is the default 404 page for Netlify. An updated page should contain links back to the root page of the site and graphic.

rijusougata13:
is not back to our site link does the same job?

smcgregor:
Yes, but the screen shot above has completely different styling from the site and the messaging targets developers rather than end users. In short, it is not user friendly.

PR #96, opened 2021-02-24
Add custom 404 page

alexmcode:
Issue #86 

Let me know if we should customize it more in terms of UI and text.


smcgregor:
Nice how this gets picked up by the framework. A couple comments:

* Set a title on the page. Right now it is parroting back the address (in my browser at least). I suggest: "Page not found".
* Change the messaging a bit. 
```
<h4>Unknown page.</h4>
<h5>Please use the menus to navigate to an existing page.</h5>
```

More design would also be nice, but I think getting the minimum in here is good.

Issue #76, opened 2021-01-19
Convert datestrings to unix epoch


smcgregor:
The datestrings in the database are all in `YYYY-MM-DD` format and should be converted to their unix epoch for them to be more readily indexed and manipulated. This requires changes to the server side functions as well as a migration of all incident reports stored in the database.

PR #97, opened 2021-02-25
Add first verison of RSS Feed

alexmcode:
Hey @smcgregor so right now we are fetching the title, URL of the article referred in the incident and description for all incidents.

It looks like this:

image

Let me know if we should do something different with the formatting or with the query. From what I understood this RSS feed is supposed to return all incidents.

### DIFF
--- a/site/gatsby-site/gatsby-config.js
+++ b/site/gatsby-site/gatsby-config.js
@@ -109,7 +109,6 @@ const plugins = [
         {
           serialize: ({ query: { allMongodbAiidprodIncidents } }) => {
             return allMongodbAiidprodIncidents.edges.map((edge) => {
-              console.log({ edge });
               return Object.assign({}, edge.node.frontmatter, {
                 title: edge.node.title,
                 url: edge.node.url,
@@ -120,7 +119,7 @@ const plugins = [
           },
           query: `
             {
-              allMongodbAiidprodIncidents {
+              allMongodbAiidprodIncidents(sort: {fields: date_published, order: DESC}, limit: 100) {
                 totalCount
                 edges {
                   node {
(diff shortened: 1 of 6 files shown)


site/gatsby-site/gatsby-config.js
```diff
@@ -90,6 +90,54 @@ const plugins = [
...
+              site_url: siteUrl
+            }
+          }
+        }
+      `,
+      feeds: [
+        {
+          serialize: ({ query: { allMongodbAiidprodIncidents } }) => {
+            return allMongodbAiidprodIncidents.edges.map((edge) => {
+              console.log({ edge });
```
smcgregor, line 112: leftover log.

smcgregor:
Performance wise, any idea if it is better to subset the number of records? This will return 1400+...

Do you need this issue to get closed out first? https://github.com/PartnershipOnAI/aiid/issues/76

alexmcode:
>Performance wise, any idea if it is better to subset the number of records? This will return 1400+...

So this code would run only when we do a build. I did a test on my local machine building with 100 incidents VS all 1400 + and the difference was 1 second. This time can increase with time especially if we decide to pull more fields from DB in order to change the RSS feed template.

So I think we can limit to 100 newest incidents. Thoughts?


>Do you need this issue to get closed out first? #76

No, I would we don't need converting to unix epoch. [Gatsby relies on moment.js](https://www.gatsbyjs.com/docs/graphql-reference/#dates) and we can format in different formats and locals straight in the query.

smcgregor:
Thanks for the speedtest. No need to subset at the moment. I am good to merge if you are also good and you clear the log statement.

smcgregor:
I see you already took care of it :)

I'll merge in a moment...

smcgregor:
I asked the requestor of this feature for [feedback](https://github.com/PartnershipOnAI/aiid/issues/45). I also realized we need a link to the feed somewhere in the UI. Likely next to the GitHub link at the top of the page.

PR #100, opened 2021-03-05
Consolidate auth functions

### LINKED ISSUE #99: Consolidate auth functions into userContext
Consolidate auth APIs and user authentication status into single Context and wrap Layout with Provider. so we can access into userAuthentication status easily under any component. As auth status is delivered from context, we can reduce login attempts.

jt-mchorse:
Closes #99


smcgregor:
@jt-mchorse can you merge or rebase this to clean up the diff?

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
smcgregor, line 1: Why a yarn file?

site/gatsby-site/gatsby-browser.js
```diff
@@ -1,11 +1,11 @@
 export const onServiceWorkerUpdateReady = () => {
   const answer = window.confirm(
-    `This website has been updated. ` +
-      `Reload to display the latest version?`
-  )
+    `This website has been updated. ` + `Reload to display the latest version?`
```
smcgregor, line 3: It looks like linting was not quite aggressive enough here. No need to change this, but noting it should likely just be one string.

smcgregor:
## Blocking merge

* The header on the new discover app is duplicated in this merge request. I think this may be a merge issue as you merged master back in. See: https://deploy-preview-100--aiid.netlify.app/apps/discover
* I am not able to update the incident ID in the form. Do you think it is a permissions issue I should look into on the server side? 

## Discussion needed before merge

* From a code perspective, and what it means for the future, I am most concerned about the authentication and session management approach. It seems this solution relies on `gatsby-plugin-layout`, which appears to be core Gatsby functionality that was moved into a Plugin in V2. Should we be concerned about anything there?

## Good to fix, but not blocking merge

* The update form has a "review" link that when you click on it navigates beneath the modal, but otherwise does nothing. I think that may be needed when submitting new content but not when updating existing content.
* From a functional perspective, most of these changes are needed for a small number of trusted users in the near future, so even if the review and deduplication functionality is not 100 percent it makes sense to merge it and improve it through time. I have not seen an example of deduplication in play yet, but I think I am comfortable with that at the moment. A screen capture of the expected view for duplicated data would be useful here.

Other than these, I think this is likely going to be merged on next review. I will take one more thorough look through then hit the button.

site/gatsby-site/src/components/theme/themeProvider.js
```diff
@@ -28,7 +28,7 @@ class ThemeProvider extends React.Component {
   };
 
   render() {
-    const { children, location } = this.props;
```
smcgregor, line 31: Why is this being removed? I don't know what it does to begin with.

jt-mchorse: if you checkout Header component, it does not uses the location props at all. so I removed it.

jt-mchorse:
> * From a code perspective, and what it means for the future, I am most concerned about the authentication and session management approach. It seems this solution relies on `gatsby-plugin-layout`, which appears to be core Gatsby functionality that was moved into a Plugin in V2. Should we be concerned about anything there?

I think we are good, V2 introduced wrapPageElement and wrapRootElement instead of layouts, but they brings few issues on production, like having duplicated headers (like what we had in discover app), I've tried wrapPageElement and wrapRootElement in our project too and it also broke the header in production build. so we used this plugin.

smcgregor:
I am going to start a bunch of processing of incident data from the deploy preview and merge if everything holds together. I will also likely have a bunch of questions I will post inline to understand some of the changes better.

site/gatsby-site/src/components/forms/SubmitForm.js
```diff
@@ -0,0 +1,70 @@
+import React, { useEffect, useState } from 'react';
+import { Button, Container } from 'react-bootstrap';
+import { CSVReader } from 'react-papaparse';
+import cx from 'classnames';
+
+import RelatedIncidents from 'components/RelatedIncidents';
+import IncidentForm from 'components/forms/IncidentForm';
```
smcgregor, line 7: New issue: This is not really an incident form, it is an incident report form. The distinction is subtle, but incidents are collections of incident reports. I know the collection is currently titled "incidents", but this will need to change in the future because this distinction is likely to confuse people in the future. I think the best renaming is likely to change the react names to be "reports" rather than "incidents" and we will migrate MongoDB through time first by mirroring the incidents collection, then removing permissions on the incidents collection to confirm everything can work without it.

site/gatsby-site/src/components/styles/Form.js
```diff
@@ -0,0 +1,57 @@
+import styled from 'styled-components';
+import { Form, Button } from 'react-bootstrap';
+
+// Reference:
```
smcgregor, line 4: I think this has moved beyond the original reference and could likely be removed here.

site/gatsby-site/src/components/theme/themeProvider.js
```diff
@@ -37,11 +37,7 @@ class ThemeProvider extends React.Component {
     return (
       <div>
         <Global styles={baseStyles} />
-        <Header
-          location={location}
-          isDarkThemeActive={isDarkThemeActive}
-          toggleActiveTheme={this.toggleActiveTheme}
-        />
+        <Header isDarkThemeActive={isDarkThemeActive} toggleActiveTheme={this.toggleActiveTheme} />
```
smcgregor, line 40: New Issue: remove dark theme support. I don't want to support it because the templating is going to be far more difficult to support as the feature set expands.

site/gatsby-site/src/components/RelatedIncidents.js
```diff
@@ -0,0 +1,75 @@
+import React, { useEffect, useState } from 'react';
+import { ListGroup } from 'react-bootstrap';
+import Loader from 'components/Loader';
+
+import config from '../../config';
+import { useMongo } from 'hooks/useMongo';
+
+const RelatedIncidents = ({ incident = {} }) => {
```
smcgregor, line 8: New Issue (high priority): return related incidents under the following circumstances...

1) All incident reports with publication dates that are 2 weeks before or 2 weeks after the current publication date
2) All incident reports with dates that are 2 weeks before or 2 weeks after the current incident date

The newer `epoch_date_published` and `epoch_incident_date` fields should make this easier.

This is important because incident reports tend to land first with a single article then a blast of articles at a later date.

smcgregor:
New issue (high priority): the submit form says "update" even when you are submitted a new incident report that you are not updating.

smcgregor:
blocking merge: when you fill out the complete submission with all the fields, the submit button says "update" and when you click it it creates a new record in the review queue, locks the "update" button, and doesn't tell the user anything. The user will not know they successfully created a report into the review queue.

smcgregor:
New issue: The "Reject New Incident" button on the review queue comes up with a 502 error. Logs indicate "uncaught promise rejection: delete not permitted". I can delete manually after merge so we can see if this is really just a permissions issue, but I think it is likely to persist.

smcgregor:
I have not tested the CSV import functionality yet, but there is only one issue that is blocking merge at the moment (see above). @jt-mchorse please take a look at the blocking issue ASAP, then I will merge and you can start chewing on the issues I created in the comments above.

jt-mchorse:
> Even if you enter an incident number into the form, you are not updating an incident report. You are still submitting a new incident. It sounds like you need to track in the form whether it is a new incident report or an existing incident report that is being edited. Checking the incident number does not help with this.

Yes, it's calling `createReportForReview` always if user is on Submit page, I'm updating the submitting logic.
I'll update the submission logic

smcgregor:
> New issue: The "Reject New Incident" button on the review queue comes up with a 502 error. Logs indicate "uncaught promise rejection: delete not permitted". I can delete manually after merge so we can see if this is really just a permissions issue, but I think it is likely to persist.

This appears to be a more systemic issue related to the way this PR changes authentication. I pulled the code locally, and after merging master in I am still not able to either reject or promote reports. I'll work to debug it now.

smcgregor:
The issue appears to be a race condition where the anonymous user sets the state after the token user's state is set.

smcgregor:
I opened a merge request to your branch. It updates the branch to track master. https://github.com/jt-mchorse/aiid/pull/3

smcgregor:
Closing in favor of #163

PR #121, opened 2021-03-26
Removed algolia css dependency from new discover app

alexmcode:
The checking here should be to observe if we left out any Algolia styling


smcgregor:
I believe the pagination styling is missing, but it is otherwise looking good to me.

Screen Shot 2021-03-27 at 9 49 46 AM

smcgregor:
Something strange is happening to the `x` button on the search box when I hover my cursor over the search box. It looks like the highlight is not quite aligned with the symbol.

I am going to merge this because I believe that is an issue unrelated to the pagination styling.

PR #122, opened 2021-03-27
update package.json minor versions

smcgregor:
Changes to the package produced by `npm update`.


smcgregor:
@jt-mchorse @alexmcode Please don't look at this at depth unless you think it may cause problems. These are all minor version updates.

Issue #84, opened 2021-02-01
Update Incident Definition and Requirements


smcgregor:
Below is a draft update for the incident definition that goes into greater depth and nuance. This is a collaboration of Zachary Arnold at CSET and myself.

---

# What is an AI incident?

The AI Incident Database contains records of AI incidents. An “AI incident” is a situation in which AI systems _caused_, or _nearly caused_, _real-world harm_. In applying this definition, note that it’s meant to be broad. If, after careful inquiry, you’re not sure whether an incident qualifies, err on the side of submitting it.

## "AI"

For our purposes, “AI” means the capability of machines to perform functions typically thought of as requiring human intelligence, such as reasoning, recognizing patterns or understanding natural language. AI includes, but is not limited to, machine learning - a set of techniques by which a computer system learns how to perform a task through recognizing patterns in data and inferring decision rules, rather than through explicit instructions.

## "AI system"

By "AI systems," we mean technologies and processes in which AI plays a meaningful role. These systems may also include components that do not involve artificial intelligence, such as mechanical components.

> **Examples:** A self-driving car; facial-recognition software; Google Translate; a credit-scoring algorithm.

Algorithms that are not traditionally considered AI may be considered an AI system when a human transfers decision making authority to the system.

> **Example:** A hospital system selects vaccine candidates based on a series of hand tailored rules in a black box algorithm.

## "Caused"

When we say that an AI system “caused” harm, we mean that it played an important role in the chain of events that led to harm. The AI system doesn’t need to be the only factor, or even the major factor, in causing the harm. But it should at least be a “but-for” cause - that is, if the AI system hadn’t acted in the way it did, the specific harm would not have occurred.

We make no distinction between accidental and deliberate harm (i.e., malicious use of AI). For purposes of the AI Incident Database, what matters is that harm was caused, not whether it was intended.

## "Nearly caused"

When we say that an AI system “nearly caused” harm, we mean that it played an important role in a chain of events that easily could have caused harm, but some external factor kept the harm from occurring. This external factor should be independent of the AI system and should not have been put in place specifically to prevent the harm in question.

> **Example:** An industrial robot begins spinning out of control, but a nearby worker manages to cut power to the robot before either the robot or nearby people are harmed.

> **Counterexample:** An industrial robot begins spinning out of control, but its built-in safety sensor detects the abnormality and immediately shuts the robot down.

[...]

PR #123, opened 2021-03-28
Feature update incident def

smcgregor:
#84

* Todo before merging
  * Does not merge until https://github.com/PartnershipOnAI/aiid/pull/122 lands
  * Add additional examples to the table where it is currently marked as `TODO`
 
I had to edit the CSS of the discover application to make the surrounding app's blockquote look proper. I believe the build system is injecting the CSS files from the static folder throughout the app, at least on my local dev machine. I either need to figure out why it is getting picked up by the build system, or land https://github.com/PartnershipOnAI/aiid/pull/122 after removing the discover app from the static directory.


smcgregor:
Cleared all todos. Merging.

PR #125, opened 2021-03-28
Debugging Builds for PR 100

smcgregor:
See: #100 

It is working locally for @jt-mchorse, but not on the Netlify preview.


smcgregor:
From @jt-mchorse 

```
To reproduce in local, please run gatsby build and gatsby serve

it will host production build at localhost:9000 and you can see what’s happening
```

smcgregor:
This debugging PR is no longer needed.

Issue #115, opened 2021-03-20
Facet Search for Dates


smcgregor:
Two things on this,

1) The color of the facets that are selected are a bit close to the color of the header for the facet. Maybe make it lighter?
2) I added a field to the database for Unix epoch dates. This enables date range selection on an additional set of facets. Let's add facets for epoch_incident_date and epoch_date_published. Additional facets for epoch_date_downloaded, epoch_date_submitted, and epoch_date_modified would be useful for me and a few others, but I don't think they are broadly useful for the community. I think they should be displayed as facets if the user is authenticated.

alexmcode:
I think that what would work best would be just 2 date inputs (maybe datetime inputs?) for selecting the range. would you agree?

smcgregor:
The resolution of dates in the database are for the day even if the unix epoch is more specific. You can have a date picker, but not a datetime. I'll update the Algolia index today with these fields and fully merge my pending merge request.

smcgregor:
Algolia index is updated with the epochs and I added them to the supported facets list of the index. Should be ready to integrate with the UI.

Issue #113, opened 2021-03-20
Page auto scrolls to top when selecting a facet


smcgregor:
It is jarring when the UI shoots to the top following a click of one of the facets. I believe it should stay at the current scroll position when selecting facets because the user may be selecting more than one facet in that area of the page.

Issue #111, opened 2021-03-20
No Results Display is awkward


smcgregor:
Note: I will shortly be merging the update to the discover app that brings it into the Gatsby application from a separate static JS application. Initially both apps will be in place with one of them hidden from the nav structure, but as soon as all the functionality and UX from the old Discover app is migrated we will fully retire the old app. This is the first of a series of issues I am going to open for completing the switch over.

The no-results behavior has the right sidebar shooting over to the left hand side (see below). The expected behavior is that the facet bars will remain pulled to the right. Additionally, there should be a "no results" message displayed, such as "Your search returned no results. Please clear your search in the search box above or the filters".

<img width="790" alt="Screen Shot 2021-03-20 at 11 53 53 AM" src="https://user-images.githubusercontent.com/64780/111882476-1581bb00-8973-11eb-8662-87dde32d9001.png">

alexmcode:
@smcgregor if pr #126 is ok and merged I think this can be closed

PR #126, opened 2021-03-30
Discover app improvements

alexmcode:
List of improvements:
- Add date range facets
- Handle no-result incident case
- Fix scroll to top on search
- Scroll to the incident on citation page
- Fix click on incident number


alexmcode:
I also added some initial limitations (that I think we need to experiment around with) for instance the user should not be able to query when from date is smaller than to date

smcgregor:
Functional review:

* move the incident and publish date selectors to the spot above "flagged
* Don't include a submit button, just apply the facet immediately upon change
* Typing on the field produces very strange behavior for me. Try typing in a variety of dates in different positions and let me know if you don't experience an issue
* The publish date facet disappears when there are no results returned. Oddly, the Incident Date facet remains.
* I am OK with the user selecting data ranges where the end date is before the start date so long as the behavior is as expected. I think putting boundaries on this will likely produce counter intuitive behavior. I am good with merging this after you address the review. There are other items needing the attention more.

Code generally looks good. I'll be more thorough at merge time

smcgregor:
oh, and make sure you merge master in before final review/merge to master.

alexmcode:
>move the incident and publish date selectors to the spot above "flagged

done

>Don't include a submit button, just apply the facet immediately upon change

this it's a bit harder to do. basically, refine (rerender the page) will be trigger on every keypress or on every scroll in the years list. this makes the app very unstable (a lot of requests + crash). a possible solution would be to disable the pop-up date picker and allow only typing and debounce it... not sure if investing this time is worth it.

>Typing on the field produces very strange behavior for me. Try typing in a variety of dates in different positions and let me know if you don't experience an issue

this should be fixed now

>The publish date facet disappears when there are no results returned. Oddly, the Incident Date facet remains.

not really sure how to reproduce this. can you please what are the steps to reproduce?

>I am OK with the user selecting data ranges where the end date is before the start date so long as the behavior is as expected. I think putting boundaries on this will likely produce counter intuitive behavior.

so I think we do need to have some sort of limitations or the Algolia API will crash. maybe instead of putting a min/max, we could have an alert for letting the user know we can't process that query/request?

smcgregor:
> not really sure how to reproduce this. can you please what are the steps to reproduce?

Now both date pickers disappear when there are no results. Screen Shot 2021-04-01 at 12 01 03 AM

> this it's a bit harder to do. basically, refine (rerender the page) will be trigger on every keypress or on every scroll in the years list. this makes the app very unstable (a lot of requests + crash). a possible solution would be to disable the pop-up date picker and allow only typing and debounce it... not sure if investing this time is worth it.

I am not understanding the limitations, but I am OK with simplifying the datepicker if it allows for eliminating the submit button

> so I think we do need to have some sort of limitations or the Algolia API will crash. maybe instead of putting a min/max, we could have an alert for letting the user know we can't process that query/request?

I think this fits in with the "no results" behavior. If an unsatisfiable search is entered, the "no results" results should display. This means https://github.com/PartnershipOnAI/aiid/issues/111 is related here.

alexmcode:
Removed the submit button and the behavior seems ok now. Also, I needed to tackle issue #111. let me know if it was handled properly

smcgregor:
Any idea what is triggering all the auto scrolling to the top of the page? We need to have more control over that feature.

smcgregor:
Changes for #111 look good.

The behavior of the datepicker is still messy whey typing into the element. Unless you think it will be a quick fix, let's disable typing directly into the datepicker and require the user to use the selection popover you have.

The problems with the current form include:
* strange behavior when typing in pre-unix epoch dates
* when typing in the year it is possible to accidentally type in more than 4 digits in a strange way
* It is not clear which digit is being typed in when typing into the form

Also, when using the popover the selectable years should likely be in the range of the unix epoch to the present (or some near future date if a hard coded value is needed).

alexmcode:
@smcgregor disabled scroll to top. it works in 95-98% of the cases I would say.

I also disabled keyboard input for date facets. if we decide to put more time into this we can control most of the cases. because how the HTML input handles keyboard input - allows a lot of scenarios we need to handle each one individually if we want to control the entire UX.

>strange behavior when typing in pre-unix epoch dates

I'm not sure I understand what `pre-unix epoch dates` means

>Also, when using the popover the selectable years should likely be in the range of the unix epoch to the present

the `From` date pickers has a min - initial value from algolia and the `To` date picker has a max, the same from algolia. if the user selects a date outside the range algolia will throw an error that we catch and display `no incidents found` message. is this what you are referring?

smcgregor:
It looks like you are preventing it from scrolling to top after it starts to scroll?

`pre-unix epoch dates` are those dates before ~1970. They are negative numbers.

alexmcode:
>It looks like you are preventing it from scrolling to top after it starts to scroll?

What I'm doing is saving the current scroll position and updating it to the last saved value on when the number of hits changes - on every query. The reason page is scrolling to the top is that on search the hits container is rerendering with next hits - so the page changes dimensions

alexmcode:
@smcgregor did the latest changes that we discussed. please let me know if I missed something

alexmcode:
Should fix issues:
#128 
#115 
#113

smcgregor:
@alexmcode it looks like merging the other MR now conflicts here. Can you merge master in here and I will do a final review and merge?

site/gatsby-site/src/templates/cite.js
```diff
@@ -65,6 +65,20 @@ const IncidentCite = ({ data, ...props }) => {
     allMongodbAiidprodIncidents: { group },
   } = data;
 
+  const [hasScrolled, setHasScrolled] = useState(false);
+
+  const scrollToIncidentCard = () => {
+    console.log('scrollToIncidentCard');
```
smcgregor, line 71: remove debug logging here.

smcgregor:
It looks like the only thing left from this PR that did not land in prior PRs is the scrolling controls?

alexmcode:
@smcgregor hm so I see the cite page scrolls to the card clicked in the discover app and the page does not scroll to top on search (in the discover page). Am I missing something?

smcgregor:
This PR is called `Add date range facets`, but that is not the functionality that will land with this PR at merge time. Also, it looks like merging the sort PR brought this back into conflict.

alexmcode:
>This PR is called Add date range facets, but that is not the functionality that will land with this PR at merge time. Also, it looks like merging the sort PR brought this back into conflict.

I will rename this PR `Discover app improvements` and add a description with all things we add. does this sound good?

smcgregor:
Sounds good. I was mostly making sure I was not missing something.

PR #136, opened 2021-04-12
Add search box in langind page

alexmcode:


smcgregor:
Is there a Gatsby compliant way to navigate that won't trigger a full re-render?

smcgregor:
Also, put the following text beneath the search area: `Entering text above will search across more than 1200 incident reports`

alexmcode:
You mean by navigation back? or in what way re-render?

smcgregor:
When you type a search into the box and it navigates you via `history.go();`, it makes the header (maybe the whole page?) re-render. I think there should be a more friendly way to navigate in scripting.

alexmcode:
I've added a gatsby native navigation but this can be achieved with <Link> component -> a search button. i think it looks proper and by not executing the search automatically we will not take the user by surprise

site/gatsby-site/src/templates/docs.js
```diff
@@ -7,12 +7,39 @@ import Layout from 'components/Layout';
...
+const DiscoverAppSearch = () => {
+  const [searchTerm, setSearchTerm] = useState('');
+
+  return (
+    <>
+      <h1 className="heading1">Search for an incident</h1>
+      <InputGroup className="mb-3">
+        <InputGroup.Prepend>
+          <Button as={Link} to={`/apps/discover?s=${searchTerm}`} variant="outline-secondary">
+            Button
```
smcgregor, line 24: This should say "Search"

smcgregor:
This looks good to me functionality and visual wise. I'll merge shortly.

smcgregor:
Actually, one change is needed: when you press enter while the focus is on the search box it should issue the search.

smcgregor:

PR #137, opened 2021-04-13
Refactore leaderboards and add them in langing page also

alexmcode:
I\m not sure if the positioning in the landing page is optimal

### DIFF
--- a/site/gatsby-site/pages/summaries/leaderboard.js
+++ b/site/gatsby-site/pages/summaries/leaderboard.js
@@ -89,7 +89,7 @@ export const Leaderboards = ({ leaderboardsToGenerate, incidentData, limit, item
     <>
       {leaderboards.map((board) => (
         <div key={board.title}>
-          <h2>{board.title}</h2>
+          <h1 className="heading1">{board.title}</h1>
           <ul>
             {itemRender ? (
               <>{board.array.map((item, index) => itemRender(item, index))}</>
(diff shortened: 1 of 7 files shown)


smcgregor:
* These leaderboard blocks need some explanation for what information they are presenting. Maybe just a header for `Incident Report Submission Leaderboards`.

* Can you make the current leaderboard into search links as well?

* Let's also create a new `about` page that contains all the textual content from the current landing page, then let's make the landing page have exclusively the content you are developing. I will add a little bit of introductory content as well that links to the new `about` page.

alexmcode:
>Let's also create a new about page that contains all the textual content from the current landing page, then let's make the landing page have exclusively the content you are developing. I will add a little bit of introductory content as well that links to the new about page.

@smcgregor I think we should leave this point after we merge all #135 tasks. it will make the commit history much more readable and it will be also easier for us to develop. WHat do you think? does it make sense?

site/gatsby-site/pages/summaries/leaderboard.js
```diff
@@ -1,89 +1,15 @@
...
-        {content.map((value) => (
-          <li key={`${title}-${value[0]}`}>
-            {value[0]}: {value[1]}
-          </li>
-        ))}
-      </ul>
-    </>
-  );
-};
+import SubmittersLeaderboards from 'components/leaderboards/SubmittersLeaderboards';
```
smcgregor, line 7: `SubmittersLeaderboards` -> `SubmittersLeaderboard`

site/gatsby-site/src/components/leaderboards/Leaderboard.js
```diff
@@ -0,0 +1,37 @@
...
+
+  for (const item in dataHash) {
+    sortedArray.push({
+      label: dataHash[item].fieldValue,
+      value: dataHash[item].totalCount,
+      attribute,
+    });
+  }
+
+  sortedArray = sortedArray.sort((a, b) => {
```
smcgregor, line 15: Does graphql have an order-by property that would allow for avoiding this sorting?

smcgregor:
This looks great. I left 2 inline comments. One for consistency in naming and another potential tweak. Otherwise this looks good.

site/gatsby-site/src/components/leaderboards/Leaderboard.js
alexmcode: No, it doesn't seem to have an order bu for group aggregations. I think the performance is good enough as it is. If we really want to have them already sorted in the client we could add a new mongo collection

smcgregor:
Implementation is excellent. Can you put the blocks into a 3 column responsive layout? A basic card layout for visual separation would also be helpful. Once that is good I will compose this MR with the other one for the search box and the complete updating of the landing page and merge.

smcgregor:

PR #142, opened 2021-04-21
Add carousel in landing page

alexmcode:
@smcgregor should this landing page item have some sort of title or description?

current look:
image

### DIFF
--- a/site/gatsby-site/src/components/RandomIncidentsCarousel.js
+++ b/site/gatsby-site/src/components/RandomIncidentsCarousel.js
@@ -1,0 +1,83 @@
+import React from 'react';
+import { StaticQuery, graphql } from 'gatsby';
+import styled from 'styled-components';
+import md5 from 'md5';
+import Carousel from 'react-bootstrap/Carousel';
+import { Link } from 'gatsby';
+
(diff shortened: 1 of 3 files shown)


smcgregor:
Since some of the incidents have a lot more incident reports, the result of the current graphql query is that most of the returned incidents are associated with incident 1. It may be as much as 30 out of the 50. I think you should update the graphql query to something like the following,

```
query RandomIncidentsCarousel {
  allMongodbAiidprodIncidents(limit: 5, filter: {incident_id: {in: [INSERT_SELECTION_HERE]}}, sort: {order: ASC, fields: id}) {
    edges {
      node {
        id
        incident_id
        title
        image_url
      }
    }
  }
}

```
When your randomly sample 5 numbers in the 1 to 100 range and put them in `INSERT_SELECTION_HERE`, there will still be a bias towards incidents with many reports, but I think that is OK here.

The query also returns all the incident reports you want to render in a deterministic but random ordering since it sorts them based on the `id` field.

smcgregor:
I read more about static queries and see this won't work. I am simplifying and pushing an update with your other landing page changes.

smcgregor:

Issue #92, opened 2021-02-22
Georgetown CSET Taxonomy


smcgregor:
This issue is for mapping the CSET taxonomy to the AIID user interface and database. After adopting a design we will open one or more issues decomposing the design into a fairly large number of implementation steps and order them according to their time to delivery and importance for supporting general taxonomies. These issues blaze the trail for other taxonomies to follow, so solid documentation of the development process is important.

This is CSET's annotation guide [appendix](https://github.com/PartnershipOnAI/aiid/files/6019473/SubsetCSET.pdf) motivating the design. We also have CSET's spreadsheet collecting many of the classifications, but I want to concentrate on this document first.

Let's start with a discussion of the database model.

## New Table: Taxa

We will create a new collection, "taxa", within which there will be a collections of taxa schemas. This will be the programmatic enforcer of types and values that will be applied in tags on the incident reports. Supportable types include the [native types of MongoDB](https://docs.mongodb.com/manual/reference/bson-types/). Step one of the design is thus processing the annotation guide into a schema of a taxa table record. For example,

```JSON
{
"namespace": "CSET",
"description": "This should be a markdown document that will be rendered to HTML by the Gatsby build process",
"fields": {
  "field name 1": {"type": "short_string", "default": "", "mongotype": "typeval", "short description":"This goes in tooltips", "long description":"This goes in the documentation page"},
  "annotator": {"type": "string", "short description":"CSET researcher applying taxonomy", "long description":"The CSET taxonomy is presently a closed taxonomy, meaning only persons affiliated with CSET are permitted to act as editors within the CSET namespace. The person indicated by the annotator field is the one who applied the taxonomy to the incident."},
  "...": {},
},
}
```

Within this collection, the fields for each named field are the following,

* "type": {"text", "select", "date", "year", "enum", "bool"}
* "default": "" # the values that sit in the form as a starting value when applying a taxonomy 
* "mongo_type": "" # MongoDB datatype
* "short description": "This goes in tooltips"
* "long description":"This goes in the documentation page"
* "permitted_values": [] # optional, only for select and enum types

The description field should get picked up by a component rendering a collection of pages describing each taxonomy (i.e., something similar to the CSET document linked above).

## New Table: Classifications

A table for recording taxonomy values for each of the incidents.

```JSON
{
"1": {
 "CSET": {"annotator":"name here", "...":"..."}
},
"2": {
}
}
```

This will be rendered on incident citation pages and consumed or set by various applications built into the AIID.

[...]

PR #144, opened 2021-04-23
First taxonomy (CSET)

alexmcode:
For accessing the mock taxonomy please go to /taxonomy/CSET

@smcgregor  Let me know if we need to do any other preparations for going to the next steps or if you had something different in mind for this step


smcgregor:
You can now switch from mocks to real data sitting on the taxa collection.

See: https://github.com/PartnershipOnAI/aiid/pull/145
And my last comment on #92 

Let's presume the taxonomy page's intended audience will be end users that do not want to know about JSON formats or other details. You only need to provide the following information:

1. The namespace
2. The description for the taxonomy. This should render the text from markdown to HTML format
3. The field_list entries, sorted according to weight, and displaying only the long name and the long description

The rest of the details in the collection are meant more for future use after we get this fully rendered as a form of documentation for people looking to understand what the taxonomy is.

alexmcode:
Did the changes according to your details. For now, the list of taxonomies (page /taxonomies) is static - hardcoded - in order to generate the side menu. In the next iteration, I think we can generate this page also.

Let me know if the UI for the taxonomy page template works (i.e. taxonomy/cset)

smcgregor:
The classifications collection now contains two documents as shown on the issue: https://github.com/PartnershipOnAI/aiid/issues/92

alexmcode:
@smcgregor I've added the classifications fetching in the cite page. Let me know if the design is appropriate

Also if I'm not mistaken, classifications fields should be shown in weight order right? Still need to figure out an optimal way of doing that. coming in the next commit

smcgregor:
This is looking good. Comments below.

## Required for merge

* There are a bunch more fields I am going to add before all the data is imported. I don't want the entire citation page to be taken up by different taxonomies, so I think we should hide all the taxa fields that have a weight less than 50. Those fields should be shown after a user clicks "Show All Classifications" at the bottom of the taxonomy's card.
* Sort the taxa fields according to their weight.
* I need to specify the remaining entries for the taxa collection and get them into at least the first example.
* Move the `Tools` block to just under the `Reports` block
* The header of the `CSET Taxonomy Classifications` card can have a link that pulls to the right side that says "Taxonomy Details" and links to the taxonomy page you created.

## Nice to haves

* Place a tooltip on the taxonomy entries with their short descriptions from the taxa collection.
* boolean types can display as yes/no rather than true/false
* The name of the classification (e.g., Annotation_Status) should be separated by a space rather than the underscore
* When the classification has a display_type of `date`, the datetime should display as YYYY-MM-DD
* The incident reports lack a header or something differentiating their section in the page from the rest of the content. This is because they are all contained in their own cards, but they should really have something delineating them from the rest of the page content.
* The taxonomy page can summarize/list/count the the values taken by each of the fields.

alexmcode:
> The taxonomy page can summarize/list/count the the values taken by each of the fields.

Not sure I fully understand. You mean to display for every field how many incidents have values for that specific field? (i.e. 34 incidents specify `Annotator`)

smcgregor:
I think this suggestion is a bigger feature than makes sense at this
moment. For each field type, we can report different aggregate
statistics. String, enum, and list fields can report the counts for
the distinct string values. Date fields can render a timeline.
Anything you find quick and easy can be done now, but otherwise this
should wait.

On Thu, Apr 29, 2021 at 8:37 AM Alexandru Muscă
***@***.***> wrote:
>
> The taxonomy page can summarize/list/count the the values taken by each of the fields.
>
> Not sure I fully understand. You mean to display for every field how many incidents have values for that specific field? (i.e. 34 incidents specify Annotator)
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.



-- 
Sean McGregor

PhD, Machine Learning
Website: seanbmcgregor.com
Twitter: seanmcgregor
irc.freenode.net: smcgregor

alexmcode:
@smcgregor I've added the migration script.
Running locally the script:
1. export from google docs spreadsheets in csv format
2. copy the exported file in the scripts folder - same level as the script
3. add the absolute path for the csv file in the script at line 3
4. run script with command: `node ./src/scripts/migrateTaxaToClassification.js`

Let me know if I'm on the right track.

smcgregor:
@alexmcode this is the right track. Can you post the resulting JSON here and I'll update the classifications table and/or comment on the generated output? If you don't have this readily available I'll take care of it this weekend.

smcgregor:
I made the following changes locally,

```
--- a/site/gatsby-site/src/scripts/migrateTaxaToClassification.js
+++ b/site/gatsby-site/src/scripts/migrateTaxaToClassification.js
@@ -1,6 +1,8 @@
 const csv = require('csvtojson');
+fs = require('fs');
 
-const csvFilePath = 'rawClassifications.csv';
+const csvFilePath = 'MY PATH HERE';
+const outFilePath = 'MY PATH HERE';
 
 // i = 5 is the first row
 // i < 5 is for headers and subheaders
@@ -168,6 +170,7 @@ const main = () => {
 
       console.log('========================');
       console.log(nodes);
+      fs.writeFileSync(outFilePath, JSON.stringify(nodes, null, 4))
     });
 };
```

This then produced the results,
[classifications.json.log](https://github.com/PartnershipOnAI/aiid/files/6449855/classifications.json.log). GitHub made me add the `log` extension before it would allow me to post it here.

Comments:

1. We need to decide how we are going to handle the management of the taxonomies through the user interface, because that will dictate how we handle default values and validations. Many of the values in the current JSON are placeholders like empty strings, but we likely want to just not have those keys in the documents. The edit form we will develop on the classifications will fetch the schema from the taxa collection and drop in the values that are defined in the classifications document.
2. I like having the CSV parser written in Node because it will make it easier to plug into form validation code that works both with the script and a user interface.

Next Steps:

1. I will update the taxa collection to have all the columns that are being exported in this script
2. We are going to co-develop the CSV->JSON script and this pull request over the next week, meaning we are going to update the classifications table to match what the script is exporting and update both it and the pull request repeatedly until we are in good shape to merge to master. Please let me know if/when I should update the classifications table with the generated JSON. The goal will be to have all the annotated incidents render with their classifications on the citation page.
3. After we can merge to master, we will work to get the Algolia faceted search up.
4. Once the facets land, we can get the classifications form in place so we can edit the classifications on the app.
5. After we can edit the classifications, we will get form validations according to the taxa collection.
6. Finally we apply the validations to the CSV export script and update the source data to pass validation.

After step 6 I will likely have one or two additional taxonomy namespaces to add, but we may come back to them after other critical features.

alexmcode:
@smcgregor the changes we spoke of should be done now. The taxonomy and classification data from the incident page should now reflect MongoDB data. We need to be extra careful with this check since there is a lot of data/fields.

Also please notice that fields that have values: `null`, `""` - empty string or `[]` - empty array are NOT displayed in the cite page. Let me know if that's what we want

> No need to list the test taxonomies here. It should display in all the programmatic elements that pull from the DB, but v1 will be removed from the DB before this feature goes live in production.

Indeed, this was done for convenience sake. Listing the taxonomies it will be done programmatically

> Is listing the fields in the graphql needed? It is better to allow the DB to be the exclusive authority over the fields so the code does not need to update with changes to the schema.

This is a problem. I've just noticed this. So GraphQL need an explicit list of fields to specify for the selection. Because of this we need to list all the fields. But this is not good because we need to be able to display different types of classifications/taxonomies.

We need to change the structure of the data from the classifications collection: field `classifications` needs to be change from object to an array. 

So from:
```
"classifications": {
            "Annotation_Status": "6. Complete and final",
            "Annotator": "Zach Arnold",
            "Ending_Date": "1514808000301",
            "Beginning_Date": "1420113600301",
            ............
}
```

to:

```
"classifications": [
    {
            short_name: "Annotation_Status",
            value: "6. Complete and final",
            type: string,
            weight: 40,
            long_name: "....."
            short_description: "...."
            long_description: "...."
    },
    ..............
]
```

This means putting a bit more work into the migration script. Since we don't have much time until the presentation we can work with what we have now and finish the facetting phase and refactor after. Thought?

smcgregor:
I am not very experienced with GraphQL... I take it we can't GraphQL the
'classifications' then name reference fields within it from higher up?

alexmcode:
image

No, not really. GQL is built on the idea that you KNOW how your result should look and what you want to receive. So it's not suited for dumping everything from a certain location into the client. I think the workaround I described above it's good and it's a best practice. The only downside is that we don't have the time to put in the work in doing it right now

smcgregor:
I just did a bit of research on this topic and immediately landed in a locked discussion/flame war. I can see why the GQL people would limit this, but it does feel a bit hostile to users... Let's come back to making it not explicitly listed, but for now I think we are fine with the explicit list.

alexmcode:
@smcgregor I've added the downloadindex code here since it's linked with the classifications now.

To way I did the addition of the classification field way to just construct a new object with the sum of normal incident fields (`authors`, `source_domain`, etc) with the classification fields (`Near_miss`, `Full_Description`, etc). Is this what we want? who we will identity that those fields are for CSET taxonomy (or for other taxonomies in the future)?

smcgregor:
@alexmcode A bit more repacking of the data is needed. Add a `classifications` key to the document export as below, where the value is a list of short strings `NAMESPACE:FIELD NAME:FIELD VALUE`. This will allow the field to plug into Algolia's filtering and faceting system.

```json
[{"_id":"5d34b8c29ced494f010ed45a","authors":["Alistair Barr"],"description":"Child and consumer advocacy groups complained to the FTC that Google’s new YouTube Kids app contains “inappropriate content,” including explicit sexual language and jokes about pedophilia.","image_url":"http://si.wsj.net/public/resources/images/BN-IM269_YouTub_P_20150518174822.jpg","language":"en","source_domain":"blogs.wsj.com","text":"Child and consumer advocacy groups complained to the Federal Trade Commission Tuesday that Google’s new YouTube Kids app contains “inappropriate content,” including explicit sexual language and jokes about pedophilia.\nGoogle launched the app for young children in February, saying the available videos were “narrowed down to content appropriate for kids.”","title":"Google’s YouTube Kids App Criticized for ‘Inappropriate Content’","url":"https://blogs.wsj.com/digits/2015/05/19/googles-youtube-kids-app-criticized-for-inappropriate-content/","incident_id":1,"ref_number":0,"date_downloaded":"2019-04-13","date_modified":"2020-06-14","date_published":"2015-05-19","incident_date":"2015-05-19","submitters":["Roman Yampolskiy"],"date_submitted":"2019-06-01","report_number":1,"epoch_date_downloaded":1555113600,"epoch_date_modified":1592092800,"epoch_date_published":1431993600,"epoch_date_submitted":1559347200,"epoch_incident_date":1431993600,

"classifications": ["CSET:Named Entities:Google", "CSET:Named Entities:YouTube", "CSET:Named Entities:YouTube Kids", "CSET:Location:Global", "CSET:Intent:Accident"]

}
```

site/gatsby-site/page-creators/createTaxonomyPages.js
```diff
@@ -0,0 +1,41 @@
...
+              }
+            }
+          }
+        `
+      ).then((result) => {
+        if (result.errors) {
+          reject(result.errors);
+        }
+
+        result.data.allMongodbAiidprodTaxa.nodes.forEach((taxaonomy) => {
```
smcgregor, line 27: taxaonomy -> taxonomy

site/gatsby-site/pages/apps/discover.js
```diff
@@ -35,9 +35,18 @@ import '../../static/discover/src/app.css';
 import '../../static/discover/src/index.css';
 import { add, format, formatISO, isAfter, isBefore, isEqual } from 'date-fns';
 
-export const searchClient = algoliasearch('8TNY3YFAO8', '55efba4929953a53eb357824297afb4c');
+// export const searchClient = algoliasearch('8TNY3YFAO8', '55efba4929953a53eb357824297afb4c');
```
smcgregor, line 38: Remove comment.

site/gatsby-site/pages/apps/discover.js
```diff
@@ -964,6 +980,7 @@ const RenderCards = ({
   scrollTo,
   sortByDatePublished,
 }) => {
+  console.log(hits);
```
smcgregor, line 983: Remove console logging.

site/gatsby-site/pages/apps/discover.js
```diff
@@ -1304,7 +1323,8 @@ const DiscoverApp = (props) => {
           <>
             <Container>
               <InstantSearch
-                indexName="aiid-emergency"
+                // indexName="aiid-emergency"
```
smcgregor, line 1326: This should be set next to the API strings.

site/gatsby-site/pages/downloadIndex.js
```diff
@@ -0,0 +1,166 @@
...
+        ref_number
+      }
+    }
+    allMongodbAiidprodClassifications(filter: { namespace: { eq: "CSET" } }) {
+      nodes {
+        incident_id
+        id
+        namespace
+        classifications {
+          Annotation_Status
```
smcgregor, line 39: We will likely re-add some of these later, but I think only the following should be here at the moment:

```
Harm_distribution_basis
Intent
Lives_lost
Location
Named_entities
Near_miss
Severity
Technology_purveyor
AI_Applications
AI_techniques
Finacial_Cost
Harm_Distribution_Basis
Harm_Type
Infrastructure_Sectors
Level_of_Autonomy
Lives_Lost
Named_Entities
Nature_of_End_User
Near_Miss
Physical_System
Problem_Nature
Public_Sector_Deployment
Relevant_AI_functions
Sector_of_Deployment
System_Developer
Technology_Purveyor
```

note that Finacial_Cost is misspelled.

site/gatsby-site/src/templates/cite.js
```diff
@@ -244,5 +412,69 @@ export const pageQuery = graphql`
...
+          Harm_Distribution_Basis
+          Harm_Type
+          Infrastructure_Sectors
+          Laws_Implicated
+          Level_of_Autonomy
+          Lives_Lost
+          Named_Entities
+          Nature_of_End_User
+          Near_Miss
+          Notes
```
smcgregor, line 452: Remove notes.

alexmcode:
@smcgregor based on the last 2 commits we need to check:

- [ ] if stats data corresponds with the discover app incidents data - filtering (maybe postpone this after we redeploy the classifications data with the new CSET changes?)
- [ ] check (by using facet filtering) that the separator replacement for the facets was done correctly

smcgregor:
@alexmcode I updated the MongoDB data to conform to the changes that landed here https://github.com/PartnershipOnAI/aiid/pull/155

I have not updated the Algolia index yet, but you should be able to pull that branch to get this build passing again.

smcgregor:
Todo list:

1. Add field to the taxa collection indicating when the field is not supposed to be visible to the end user. In the future these fields will be used for the annotation process, but not presented to the final end user.
2. When `Where in the annotation process is this incident?` is on step 6, the classifications should display to the end user, otherwise the classification should be hidden. This should be done last after the final data update since so many of the incidents need to reach status 6 before we launch this feature. It makes the most sense to introduce our own field that indicates whether the classification should be published for end users rather than building on top of this process logic.
3. Let's improve the rendering of this `1 => 0 Incidents`. What is currently arrow notation could be cleaner.
4. Sort the listed/counted items on the taxonomy page by count (decreasing)
5. Restyle either `Searchable in Facet` or `Show more stats`. They should not both look the same when one is not clickable.
6. Render `location` display_type fields in their counts.
7. Improve the expression of the taxonomy in the code so that additional taxonomies are easier to add. For instance, maintain a single list of the fields that should be included in the taxonomy rather than having it in multiple places in the code.

Before Launch:

1. (Sean) Update the data one last time from the most up to date spreadsheet. We froze the data when we started and will return to the up to date set.

alexmcode:
> Add field to the taxa collection indicating when the field is not supposed to be visible to the end user. In the future these fields will be used for the annotation process, but not presented to the final end user.

@smcgregor let me know when I can address this in the code

alexmcode:
> Improve the expression of the taxonomy in the code so that additional taxonomies are easier to add. For instance, maintain a single list of the fields that should be included in the taxonomy rather than having it in multiple places in the code.

I've added a way to have a dynamic list of taxonomies in the `/taxonomies` page. My concern at this point is if we are going to add other classifications, the field list of those new classifications (and taxonomies) will be different right?

In this case, because we can't query a non-defined list of classifications (please see bellow error) we will not be able to display fields, stats, etc for those new classifications.

image

@smcgregor a solution for this would be to store the classifications as an array instead of an object. Is this scenario true or I'm misunderstanding something?

smcgregor:
> Add field to the taxa collection indicating when the field is not supposed to be visible to the end user. In the future these fields will be used for the annotation process, but not presented to the final end user.

`@smcgregor let me know when I can address this in the code`

When I am back "in the office" I'll add a field, `internal_process`, which when true should result in the field being hidden by default.

Regarding multiple taxonomies, for the purpose of shipping this first taxonomy please don't worry about the the graphql and related problems but keep the code and queries organized to the extent possible to make it easier to add another one later. If you have more time after reaching a feature complete status, then I think switching to arrays sounds like it should work. Please let me know if there is a tradeoff there that I have not thought of.

smcgregor:
One change requested:  "Show less stats" should be "Show fewer stats"

My todo list:

1. I need to get you the `internal_process` change to clear the last `todo` item there. I will do this in the next few days.
2. We need finalized data from CSET to get this to release candidate status.
3. I need to update some of the messaging around this, including details for the front page.

smcgregor:
Closing this MR in favor of https://github.com/PartnershipOnAI/aiid/pull/155

@alexmcode if you have anything else on this, please target the branch of #155 and we will merge that one to master. This is almost complete! It is getting reviewed now.

PR #151, opened 2021-05-10
Increase debounce time for search discover app

### LINKED ISSUE #149: Making fast changes to instant search form input confuses form
When you type quickly into the instant search form of the discover application, it looks like you sometimes end up in a race condition between different handlers. This is particularly true when you use backspace and pause randomly. The result is that you type into the form before the form programmatically updates to the state that was previously searched, and makes it difficult to efficiently move through the search space. The form should be updated so that it respects what has been entered into the form by the user and does not update to a previously issued search just because there was a delay in getting its response back.

alexmcode:
Tries to fix #149 

I've increased the time it waits for new input before sending the request to algolia. Not sure if it fixes the problem 100%. Please let me know

Also tested multiple scenarios and use cases in order to make the search more responsive but we might have reached the limits of algolia-react module. So if we decide to do more optimisation we need to invest more time into this

### DIFF
--- a/site/gatsby-site/pages/apps/discover.js
+++ b/site/gatsby-site/pages/apps/discover.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef, useMemo } from 'react';
+import React, { useState, useEffect, useRef, useMemo, useCallback } from 'react';
 import { navigate } from 'gatsby';
 import { StringParam, QueryParams, useQueryParams } from 'use-query-params';
 import algoliasearch from 'algoliasearch/lite';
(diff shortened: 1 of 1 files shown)


smcgregor:
Any idea whether it is possible to make the user input definitive rather than the last return from algolia?

alexmcode:
We don't populate the input with the Algolia return. The problem is that we a lot of UI on almost every keypress that needs to update. And the input/rendering remains behind. 

A possible solution would be to add a loader on top of the incident results that would take the rendering burden from the browser. It would still feel like a real-time search but also possibly improve performance. Thought?

Please test these new changes in order to try to replicate the problem again

smcgregor:
Closing in favor of https://github.com/PartnershipOnAI/aiid/pull/157

PR #153, opened 2021-05-13
Migrate download index func

alexmcode:
We need to add a way to control access to this path


smcgregor:
Making this an unlisted endpoint is good. All the Algolia data (and the majority of data in general) are available for download.

smcgregor:
Closing this since it was rolled in here https://github.com/PartnershipOnAI/aiid/pull/144

PR #161, opened 2021-06-08
Disable search on loading for discover app

### LINKED ISSUE #149: Making fast changes to instant search form input confuses form

alexmcode:
fixes #149

### DIFF
--- a/site/gatsby-site/pages/apps/discover.js
+++ b/site/gatsby-site/pages/apps/discover.js
@@ -712,28 +712,29 @@ const IncidentCard = ({
   </IncidentCardContainer>
 );
 
-const StyledSearchBox = ({ refine, customRef }) => {
+// eslint-disable-next-line react/display-name
+const StyledSearchBox = React.memo(({ refine, customRef }) => {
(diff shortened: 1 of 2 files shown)


smcgregor:
This causes the form element to lose focus and the user needs to click on it again. Can you reset focus after the return?

alexmcode:
@smcgregor it seems to be reproducing only on chrome. I've added an auto-focus and it seems it's working properly. Please let me know

smcgregor:
No change on chrome.

alexmcode:
@smcgregor did another change + some performance changes. let me know if it's still reproducible

smcgregor:
Works well on Chrome and Safari. Merging.

PR #164, opened 2021-06-22
Add update fct to mongo.md + modify create report

alexmcode:
@smcgregor let me know if I missed something


smcgregor:
I updated to match the functionality you have here and fixed some issues. It is up to date in production. See: https://github.com/PartnershipOnAI/aiid/pull/165

PR #168, opened 2021-06-28
fix key issues in citation page links

smcgregor:
This resolves some duplicate key issues that were popping up in the console, and resolves an issue where the citation page links were not clicking through to the content returned by algolia.


smcgregor:
@alexmcode can you take a look at why clicking on the citation page links were no longer scrolling to the report text? I spent a few hours trying to piece this together, but it looks to me like it should not have been working to begin with. The elements are not carrying the hash as their ID, so the `document.getElementById` selector should not have been working. At any rate, kicking people to the resource directly is a good stopgap.

PR #172, opened 2021-06-30
Fix incident report link to cite page

alexmcode:
@smcgregor the problem was that for scrolling to the card, in cite page, the app used attribute `_id` which got removed from what I can see. I changed that to use `mongodb_id`.

We need to take 2 decisions at this point:
1. do we continue with `mongodb_id`? - is this going to change/be removed in the future?
2. do we still move forward with the changes for generating the cite page static and removing algolia connection from this page?


smcgregor:
1. Either way. Since it is currently generating as the `downloadIndex` endpoint generates, I think we assume it will continue to be as it is now.
2. Let's switch to statically rendering the reports on the citation page.

So I think the choice here is to roll back the changes I made to this line: https://github.com/PartnershipOnAI/aiid/blob/f49263fbb0180c4f130830efda1cdd58bd57353d/site/gatsby-site/src/components/IncidentList.js#L22

Then to add on statically rendering the reports. Then this will be ready to go.

alexmcode:
@smcgregor should work now

smcgregor:
* Master needs to be merged in.
* This needs more testing. For instance, when you go to `/summaries/incidents` and scroll down to incident 41 and click through to the citation page for that incident, it will present the content that is far down the citation page rather than the content that is at the top. I can see the problem is that scrollTo is disabled for the citation page, but there are some instances where it is needed (as demonstrated for incident 41).

alexmcode:
@smcgregor taking care of the merge conflicts it seems there is no efficient way to generate statically all the classifications data. We have 2 major directions:
1. leave the classifications data to be dynamically calculated on page load - page load time seems fine
2. fetch all the classification data and taxonomy data for ALL incidents in the static build phase and generate the needed data. although this seems like a big task it will be handle by the build phase.

Should we go with solution no. 2?

smcgregor:
The build cycle is going to become multi-part and conditional in the future. We can always go back to dynamically populating. Let's go with (2) providing it does not have client-side impacts (I believe [page-creators](https://github.com/PartnershipOnAI/aiid/tree/master/site/gatsby-site/page-creators) takes care of this). The complete index is 6 MBs at the moment. When it is broken into 100+ incidents that is not bad.

alexmcode:
Should be good to go :)

site/gatsby-site/page-creators/createCitiationPages.js
```diff
@@ -8,19 +44,94 @@ const createCitiationPages = (graphql, createPage) => {
...
+            allMongodbAiidprodClassifications(
+              filter: { classifications: { Publish: { eq: true } } }
+            ) {
               nodes {
-                true_incident_number
-                duplicate_incident_number
+                incident_id
+                id
+                namespace
+                classifications {
```
smcgregor, line 76: I think you previously collected the classifications queries elsewhere? I know we are likely to change the classifications from objects to an array, so I am concerned by spreading this out.

Also, how did this change land in the fix to the links in the citation pages?

PR #190, opened 2021-07-15
Update a few files that weren't formatted/linted

cesarvarela:
Otherwise, they will come up every time someone formats/lints the project.


cesarvarela:
I'll look at why this one failed later is not that important anyway. It seems Netlify is acting up.

cesarvarela:
@smcgregor fixed

PR #191, opened 2021-07-16
Fix spacing issues on landing page cards

cesarvarela:


site/gatsby-site/src/pages/landingPage.js
```diff
@@ -37,7 +37,17 @@ const StyledCardParagraph = styled.p`
 `;
 
 const StyledRow = styled(Row)`
-  margin-bottom: 1em;
+  & + & {
```
smcgregor, line 40: What is this `& + &` syntax doing?

smcgregor:
I have one question regarding syntax, but this looks visually much improved and the implementation looks good. Merging now.

site/gatsby-site/src/pages/landingPage.js
cesarvarela: The `&` gives you access to the class name that styled-components will generate for the current component (kind of the `this` keyword). The `+` is just the adjacent sibling selector

cesarvarela: It does look cryptic tho

PR #193, opened 2021-07-19
Make the navigation collapse the nested items if a nested item is not navigated to

cesarvarela:


smcgregor:
You can't use `window` because it is not defined during server side rendering. I think you could solve this issue by counting the number of slashes in the path.

cesarvarela:
Yeah, I had it in the back of my mind and then forgot about it, fixed.

smcgregor:
I know this was my issue/suggestion, but I think it is a bit awkward when navigating the app. Any ideas? Maybe eliminate some of the pages to decrease the overall length of nav?

cesarvarela:
Yup, the actual gitbook behavior is slightly different. When you go to an URL, it only expands the items navigated to, but if you expand some other, it remembers the previously expanded ones: https://whitepaper.axieinfinity.com/axs

smcgregor:
Let's go with the gitbook way of doing it. If this looks like it will take more than 1.5 hours, I think we should just simplify the navigation structure of the site.

cesarvarela:
I gave it a try, but I would need to rewrite the nav component to get it working properly.

smcgregor:
Closing this since we decided it would be better to move away from the docs structure in the future, at least for the core path into the DB.

Issue #104, opened 2021-03-10
Integrate News Article Parsing Endpoint


smcgregor:
As a proof of concept I implemented an AWS lambda [function to parse news stories](https://github.com/aiincidentdatabase/aiid-aws-lambda/pull/1). It runs in a docker container with the ability to ship custom Python dependencies and is a way we can move much of the server-side dependencies into on-demand infrastructure that scales arbitrarily and is only paid when requested.

The initial use case for this functionality is to bridge the quick submit and submit applications. It should be possible to hit a button on the quick submitted incident report and pre-populate the submit form with the parsed results. This can save a tremendous amount of copy/paste time.

You might notice I have a tremendous number of commits on the pull request in aiid-aws-lambda. It took a long time to debug this one but it should be much easier to ship these server-side APIs in the future. I am also considering moving much of the Realm API into this repo because it is easier to version control the scripts into deployments.

smcgregor:
See also: https://github.com/PartnershipOnAI/aiid/issues/110

smcgregor:
This is done.

PR #196, opened 2021-07-21
Integrate news parsing endpoint

cesarvarela:
Deals with part of #104

### DIFF
--- a/site/gatsby-site/src/components/forms/IncidentReportForm.js
+++ b/site/gatsby-site/src/components/forms/IncidentReportForm.js
@@ -150,29 +150,21 @@ const IncidentReportForm = ({ incident, onUpdate, onSubmit }) => {
         label="Report Address :"
         placeholder="Report URL"
         addOnComponent={
-          !isEditMode ? (
-            <Button
-              className="outline-secondary"
-              disabled={errors.url || !touched.url || parsingNews}
-              onClick={() => parseNewsUrl(incident.url)}
-            >
-              {' '}
-              {!parsingNews ? (
-                <>Fetch info</>
-              ) : (
-                <>
-                  <Spinner
-                    as="span"
-                    animation="border"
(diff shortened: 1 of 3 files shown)


smcgregor:
* Good UI and progress, but needs more work before this checkpoint will be fully functional and mergeable.
*  This needs error handling before this checkpoint will be mergeable. A cold request to the API will often generate a 500 error (as it did for me)
* It looks like the maintext is not being placed into the form properly.
* Don't make the button disappear after fetching.
* Any affordance that lets the user know that the first request can take a while would be helpful. Subsequent requests are much faster.

cesarvarela:
Regarding the cold start, there are 2 toasts:

- if the function returns 500, it says: `Error fetching news info, please try again in a few seconds.` (there is no way to tell if the function crashed or it was a cold start)
- if the function takes > 20 seconds, a warning toast that says: `Sometimes fetching news info may take a while...`

Not sure if you have in mind something more involved than that?

smcgregor:
That should work.

site/gatsby-site/src/components/forms/IncidentReportForm.js
```diff
@@ -110,8 +110,74 @@ const IncidentReportForm = ({ incident, onUpdate, onSubmit }) => {
...
+      severity: SEVERITY.warning,
+    });
+  };
+
+  const parseNewsUrl = async (newsUrl) => {
+    setParsingNews(true);
+    const timeout = setTimeout(coldStartToast, 20000);
+
+    try {
+      const url = `https://z14490usg0.execute-api.us-east-1.amazonaws.com/default/parseNews?url=${encodeURIComponent(
```
smcgregor, line 129: @cesarvarela I am going to merge this PR, but let's move this connection string into the config file. I want to keep all the lambda functions and integrations in one place.

PR #199, opened 2021-07-23
Add people to governance page

smcgregor:


smcgregor:
@cesarvarela FYI, I added you here. :)

PR #202, opened 2021-07-27
Fix related incidents

### LINKED ISSUE #176: Refactor component RelatedIncidents
Original Incident: https://github.com/PartnershipOnAI/aiid/pull/100#discussion_r624888262

Component `RelatedIncidents` needs to return related incident under 2 following circumstances:
1 .All incident reports with publication dates that are 2 weeks before or 2 weeks after the current publication date
2.All incident reports with dates that are 2 weeks before or 2 weeks after the current incident date

The newer epoch_date_published and epoch_incident_date fields should make this easier.

This is important because incident reports tend to land first with a single article then a blast of articles at a later date.

alexmcode:
Fixes #176 

@smcgregor please let me know if the issue was understood correctly.

This PR will add the following behavior:
- when typing a published date it will search 2 weeks before and 2 weeks after that published date
- when typing an incident id it will fetch reports for that incident and use the `epoch_incident_date` of the incident to fetch all reports 2 weeks before and 2 weeks after the incident date of the searched incident.

### DIFF
--- a/site/gatsby-site/src/components/RelatedIncidents.js
+++ b/site/gatsby-site/src/components/RelatedIncidents.js
@@ -4,11 +4,22 @@ import Loader from 'components/Loader';
 
 import config from '../../config';
 import { useMongo } from 'hooks/useMongo';
+import { subWeeks, addWeeks } from 'date-fns';
+
+const useRunQuery = (query, callback) => {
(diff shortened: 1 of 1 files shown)


smcgregor:
I reviewed the code and it looks good to me.

PR #209, opened 2021-07-29
Add related incidents to submitted queue

alexmcode:
@smcgregor the default behaviour for showing related incidents for the submitted queue is exactly the same as the latest changes for related incidents in report submission. Do we want different criteria for suggesting possible similar reports in this case?


smcgregor:
Let's apply the same criteria in both places, but let's improve the layout and presentation of the information.

* No results behavior should indicate that "Preliminary checks failed to find incident reports with similar publication dates (+/- 2 weeks), similar incident dates (+/- 1 month), the same report URL, or the same authors."
* Let's add checks consistent with the above for both the submission form and this review area.
* Let's render the list of incident reports if the incident number is specified (as is currently done).

alexmcode:
@smcgregor 
> Let's render the list of incident reports if the incident number is specified (as is currently done).

Meaning that if the user did not input any incident id or id is 0 we are not running any query? Currently we run a query for each of the inserted criteria - incident id or submitted date (and in this PR URL and authors)..

Another question: I've noticed that in the submitted list, when checking for similar reports the code takes into account only one criteria (the first one in the code logic - incident id or submitted date if incident id is 0). This means that for any report in the queue it will never look for similar reports by URL (for instance). So I see 2/3 options:
1. we modify a bit the related incidents area and add a new subarea for every search criteria we use:
- incidents reports matched by id
- incidents reports matched by submitted date
- incidents reports matched by url
- incidents reports matched by authors

2. we execute a single query and put OR between all these conditions - so we will have a large group of possible reports - not ideal for the user IMO

3. we execute a single query with AND between conditions - this can narrow the search and possibly find more similar reports

Thought?

smcgregor:
Let's do (1). The ideal is to combine (1) and (2) by requesting it all in an `or` condition before separating results in the client. This would eliminate several DB requests. However, this is not a feature that is likely to hammer the DB so I am comfortable not doing it this way if it means moving faster to the next feature.

alexmcode:
@smcgregor it seems I spoke too early. When doing a MongoDB query using or operator it's not possible to trace back the source of the results to a specific condition. So the only way of doing point (1) is to execute 4 queries at once. This will only impact the edit of the submitted reports - expanding a submitted report will trigger all 4 queries.

But for the submit form it will be one query at a time, only for the value that changed. So I think it's a good trade-off

smcgregor:
Functionality looks solid. UI needs improvement at https://deploy-preview-209--aiid.netlify.app/apps/submit

* There is a spinner at the bottom even if nothing is pending
* The block doesn't have any padding/margin

Once those are good I think we merge this.

alexmcode:
@smcgregor ready to review

smcgregor:
@alexmcode I believe it is fixed on this branch because it does not have master merged in. The problem will persist.

Merging this one now.

smcgregor (approved):

Issue #61, opened 2020-12-31
Use Internet Archive Document Indexing for Permalinks


smcgregor:
Many of the linked reports have their address change or their content leaves the web entirely. The internet archive offers a variety of services for permanently indexing URLs. These services should be used to update URLs as they move about and to permanently link the content as stored by the archive.

https://blog.archive.org/2017/01/25/see-something-save-something/

https://blog.archive.org/2016/09/27/persistent-url-service-purl-org-now-run-by-the-internet-archive/

smcgregor:
@cesarvarela please check whether we can do a client-side URL for this without DB changes, otherwise links or example boilerplate for intergration server side would be useful.

smcgregor:
Addressed in https://github.com/PartnershipOnAI/aiid/pull/222

I am going to leave this issue open a bit longer since I would like to return to this after we test the feature in production. We should likely apply this to all source links and not just the links in the cards. I don't want to make the change globally until we know this is no issues with it.

smcgregor:
@cesarvarela I am taking you off this very old issue, but I am going to leave it in the backlog. The UX around links out to the web content could be greatly improved. Sometimes we link to the archive, sometimes to the original link, this should be improved in the coming UI update.

lmcnulty:
I'm inclined to think we should submit the link to the archive on report submission. 

When displaying links, I can think of a few options:

- Use a link-styled [split dropdown button](https://getbootstrap.com/docs/4.2/components/dropdowns/#split-button) to allow switching between a default and alternative destination.
  - Probably not many people would think to use it.
  - Should we default to the original source, or the archive?
- Link to the original, but display an archive.org icon beside it linking to the archived version
  - Seems like the option that discloses the choice most clearly, but only to those who already know about the Internet Archive. 
  - Would also add visual clutter.
- Add a build-time option to query all the report urls, and display the archive url if the response code is non-200.
  - Most helpful to the user
  - Also most complicated.
  - Could be combined with the first option, affecting the default but still including both links.

smcgregor:
I think I prefer the first option here, with the default being the original
source. Tough call though. I don't think we should put in all the build
time mechanics yet.

PR #222, opened 2021-08-02
Add a web archive link component for incident cards

cesarvarela:
The API I'm using is here: https://archive.org/help/wayback_api.php. The component uses the URL and the `date_submitted ` item property to get the Wayback Machine's closest snapshot. If this fails, it falls back to the submitted URL.

It makes the discover page make a lot of requests though, we might want to use an intersection observer to check the link is in view but don't want to over-engineer it either.

This deals with #61


smcgregor:
Build failed so I can't check the deploy preview.

What is the API's latency? We could capture the click or hover and use the JS API rather than swapping the link target, but this would require fast feedback.

cesarvarela:
Fixed the build. It's quite fast but not fast enough for capturing before the click, I just looked and there won't ever be more than 30 cards per page? Maybe it's not that bad?

What takes way longer is opening the web archive links, but I guess that's expected.

smcgregor:
@cesarvarela please merge master in and I will review the code/merge this. Functionally it is working pretty well.

cesarvarela:
Merged and now uses `window.open` as we discussed.

Another option would be to plug into Gatsby's `onCreateNode` and add a new field on build time to Incident nodes with the report's Web Archive URL.

smcgregor:
Opening the window pending a response from the API is clever. Can you test this in Safari, Chrome, and Firefox to make sure it doesn't run afoul of any security restrictions?

cesarvarela:
Tested:
Windows: Firefox, Chrome, Edge
Ubuntu: Firefox, Chrome
macOS: Safari, Chrome
iOS: Safari, Chrome

Didn't run into any problems  🎈

smcgregor (approved):

PR #234, opened 2021-08-11
i18n support

cesarvarela:
i18n support in Gatsby is not fully-fledged yet, there are only a few official Gatsby themes that add some support for it, but there is nothing definitive.

This implementation is split into the main i18n theme and 3 child themes that implement a specific localization library. In this PR, I chose to use the `Lingui` library, which seems to be the most complete (you don't need anything else to do the full translation workflow)

With the current implementation, the default locale is `en` and is mapped to the `/` path (so current routes remain unchanged), other locales will be mapped as `/{locale}/path,` so for the `es` locale,  the paths get mapped to `/es,` `/es/some-path,` etc.

https://deploy-preview-234--aiid.netlify.app/es/

The workflow for translating the UI is:

- Replace HTML text with the `<Trans>` macro (and use the appropriate ICU format if necessary: https://lingui.js.org/ref/message-format.html)
- extract messages `npm run extract`
- translate the appropriate `.po` files
- compile the translations, so they get picked up by Gatsby `npm run compile`

More on this here: https://lingui.js.org/tutorials/react.html#introducing-internationalization

(I've excluded the `/es/**` paths from the sitemap to make it possible to commit/merge new translations without them being visible to users)

Main official Gatsby theme:
https://github.com/gatsbyjs/themes/tree/master/packages/gatsby-theme-i18n

Child theme:
https://github.com/gatsbyjs/themes/tree/master/packages/gatsby-theme-i18n-lingui

Ongoing Gatsby discussion:
https://github.com/gatsbyjs/themes/discussions/116

(This only deals with the translation of UI elements, the blog, everything coming from the DB, and auto-generated pages from .mdx files need to be updated to handle the proper locale)

### DIFF
--- a/site/gatsby-site/.babelrc
+++ b/site/gatsby-site/.babelrc
@@ -1,0 +1,12 @@
+{
+    "plugins": [
+        [
+            "@babel/plugin-proposal-export-default-from"
+        ]
+    ],
+    "presets": [
+        [
+            "babel-preset-gatsby"
+        ]
+    ]
+}
--- a/site/gatsby-site/.linguirc
+++ b/site/gatsby-site/.linguirc
@@ -1,0 +1,24 @@
+{
+    "locales": [
+        "es",
+        "en"
+    ],
+    "catalogs": [
+        {
+            "path": "<rootDir>/i18n/lingui/{locale}/messages",
+            "include": [
+                "<rootDir>/src"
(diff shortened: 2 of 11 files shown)


smcgregor:
Closing since other lines of translation are farther along and likely to be the base for i18n support.

PR #244, opened 2021-08-20
Editable classifications

alexmcode:
This would be a first version for classifications edit form. Saving the form doesn't do anything at this point. Until we have a backend function to handle this should we do something more in this PR?

### DIFF
--- a/site/gatsby-site/src/components/TaxonomyForm.js
+++ b/site/gatsby-site/src/components/TaxonomyForm.js
@@ -74,8 +74,6 @@ const EditTaxonomyForm = ({ namespace, incidentId, setIsEditing, setShowBanner }
 
   const [fieldsWithDefaultValues, setFieldsWithDefaultValues] = useState([]);
 
-  const [listTypeFields, setListTypeFields] = useState([]);
-
   const useRunTaxaQuery = () => {
     return new Promise((resolve, reject) => {
       const { runQuery } = useMongo();
@@ -107,6 +105,7 @@ const EditTaxonomyForm = ({ namespace, incidentId, setIsEditing, setShowBanner }
           {
             // 'classifications.Publish': true,
             incident_id: incidentId,
+            namespace,
           },
           (res) => {
             resolve(res);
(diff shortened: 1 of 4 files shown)


smcgregor:
Can you add a function to [here](https://github.com/PartnershipOnAI/aiid/blob/master/realm.md)? I'll drop it into the backend and test it with your deploy preview.

smcgregor:
Looking good, remaining items are:

* Add the ability to create classifications for incident records according to taxonomies in the taxa collection
* Add support for multiple taxonomies displaying on the citation page to edit
* Add Realm function for editing the taxonomy to the [documentation for Realm functions](https://github.com/PartnershipOnAI/aiid/blob/master/realm.md). I'll add and test it on the backend.

smcgregor:
@alexmcode `Add Realm function for editing the taxonomy to the documentation for Realm functions. I'll add and test it on the backend.` is blocking merge here. The other two items can be on a separate pull request.

alexmcode:
@smcgregor the first 2 points were already done. Just needed to double-check the backend function. All is needed is testing

smcgregor:
I created the realm function per this pull request and will merge shortly.

site/gatsby-site/src/components/TaxonomyForm.js
```diff
@@ -0,0 +1,518 @@
...
+
+      try {
+        runQuery(
+          {
+            namespace,
+          },
+          (res) => {
+            resolve(res);
+          },
+          'mongodb-atlas',
```
smcgregor, line 89: These should be pulled from site config, not inlined.

site/gatsby-site/src/components/TaxonomyForm.js
```diff
@@ -0,0 +1,518 @@
...
+        runQuery(
+          {
+            // 'classifications.Publish': true,
+            incident_id: incidentId,
+            namespace,
+          },
+          (res) => {
+            resolve(res);
+          },
+          'mongodb-atlas',
```
smcgregor, line 113: Same here.

smcgregor:
@alexmcode I left two inline comments about moving the query config out of the component. I am going to merge this now, but please address that comment in a follow up pull request.

smcgregor (approved):

PR #249, opened 2021-08-28
Delete old discover files

cesarvarela:
Seems this isn't used anymore


smcgregor:
Some external links are still hitting this, which is more of a reason to remove it. Once you add [a redirect](https://github.com/PartnershipOnAI/aiid/blob/master/site/gatsby-site/gatsby-node.js#L28), I am good with merging this. Note that the redirect will not work in your development environment and will need to be tested on the deploy preview.

cesarvarela:
Updated

smcgregor:
Something strange is happening in the styling of the discover app on this MR when compared to the production version. The headers of the filters are not highlighted in blue.

cesarvarela:
Fixed

smcgregor (approved):

PR #256, opened 2021-09-01
Replace hardcoded db name

alexmcode:
Fix the last review from #244


smcgregor:
Will merge after `aiidprod` is also pulled from config.

smcgregor (approved):

PR #245, opened 2021-08-23
Remove lfs

cesarvarela:
I've uploaded all the lfs banners to Cloudinary, so once #235 is merged, this one should be good to merge.
They are in a folder called `legacy`:

image

image

I guess they'll need to be deleted from Netlify's lfs too?

PR #258, opened 2021-09-02
Consolidate pages on default pages folder

cesarvarela:
Also removes unnecessary pages creator plugin.

This will make i18n setups easier in the future too.

should be merged after #245 and #249


cesarvarela:
Fixed conflicts

smcgregor (approved):

PR #270, opened 2021-09-08
Classification Workflow Support

alexmcode:
- [x] Add notes fields in the schema
- [x] Add notes form fiels
- [x] Render the CSET taxonomy with react-table
- [x] support filtering
- [x] support sorting
- [x] render incident id column in the classifications table
- [ ] Render the incident data with react-table. Initially just do the incident number, title of one of the reports, and the incident date. We will refine this through review


alexmcode:
@smcgregor please let me know if this is enough to drop the new field in the backend

smcgregor:
Looks good to me. You can alternatively consider it a classification field so you can use the same forms and rendering, but I don't feel strongly about it.

alexmcode:
It will be in the same place as the rest of the classifications in the UI. the only difference is that I don't consider this to be a classification. I see it more as metadata for classifications

smcgregor:
I'm going to merge this, but it needs the following changes:

1. The scroll bars should be shown regardless of whether you are scrolling. This is needed to indicate there is more content and how to get to it. This is how Google Sheets handles wider spreadsheets.
2. The table view is useful both for internal process needs (i.e., the needs of the people that are applying the taxonomy) and for external data exploration needs. However, the "internal process" fields should not be displayed to anonymous external users.
3. Links in the table to the citation page can reference the taxonomy's anchor and scroll the page.
4. There should be a spinner when the taxonomy is initially loading into the table.
5. The taxonomy select dropdown at the top of the page currently has two values in it, but both appear to be `CSET`. Maybe one should be eliminated? Is this a backend issue?
6. The taxonomy selection drop down needs contextualization. Maybe put it in a sentence that says, "Showing the DROPDOWN taxonomy".

All these can come later since the table view is not presently linked to in any of the pages.

smcgregor (approved):

smcgregor:
Also, let's set the pagination at 500 entries.

alexmcode:
hey @smcgregor 
>The table view is useful both for internal process needs (i.e., the needs of the people that are applying the taxonomy) and for external data exploration needs. However, the "internal process" fields should not be displayed to anonymous external users.

what fields do we want as internal and also how do we determine the user type? - authenticated and non-authenticated users?

smcgregor:
Public and `isAdmin` are good for now. I am going to merge https://github.com/PartnershipOnAI/aiid/pull/275 this week, at which point you will hook this into the user accounts defined there. Admin users will retain their ability to edit.

You will also need to hide entire rows if the classification is not published. You can see the logic on the CSET taxonomy component.

PR #274, opened 2021-09-12
Staging system

cesarvarela:


smcgregor:
@cesarvarela this is now highest on the merge priority list, but I think the Realm files will need updating from production again since a few functions and schemas have changed. Do I just run the Netlify CLI again on this branch?

cesarvarela:
@smcgregor yes please do

smcgregor:
@cesarvarela can you remind me of the command sequence? I am not seeing where you shared that with me the last time.

cesarvarela:
@smcgregor sure, it is: `realm-cli pull --remote=aiidstitch2-fuwyv` (from inside the realm folder of this branch)

smcgregor:
I set all environmental variables as specified, but I also added a change to default to the production realm settings (in the config.js). This is needed because it will otherwise pickup the staging realm configuration when building master.

Let's continue discussing here, but #288 is the definitive branch to work from: https://github.com/PartnershipOnAI/aiid/pull/288

smcgregor:
I added the production details to the production GitHub environment.

smcgregor:
I was not seeing the branch deploy preview connecting to the staging version of the database, so I changed the config to pull in the netlify environment variable regardless of which branch it is, or even if it is defined on `process.env`. Now the deploy preview is complaining properly [0] rather than picking up the production DB. I don't see why it is not picking up the environment variables from netlify in the config.js, but that is what is happening at the moment. Are you sure your config.js is not referencing production?

[0] https://app.netlify.com/sites/aiid/deploys/61540c745bee150008e1ed54

smcgregor:
Production Realm integration broke. It was picking up the staging environment config since the environment variables were defined for them. I prepended "STAGING_" to the realm environment variables for the time being, which are the staging environment.

cesarvarela:
I've got the deployments working on my fork, and effectively we'll need to set up 2 Netlify sites, one for production and another for staging.

The staging site should take care of deploying the `staging` branch and its deploy previews (PRs into `staging`)

The production site should take care of deploying the `master` branch and its deploy previews (PRs into `master`)

The only difference is the Branches section on Netlify's UI, where we must set the production branch to `master` for the production site and `staging` for the staging site.

image

PR #279, opened 2021-09-23
Incidents table view

alexmcode:
Includes also changes from #270 

For accessing incidents table navigate to: /apps/reports

Only the most recent commit on this branch  - 9d61c41 is relevant for this PR


smcgregor:
I am going to merge this now, but like the classifications table there needs to be changes before we will integrate it into the navigation and make it more outward facing.

* `Incident Table View` -> `Incident Report Table`
* Let's increase the default pagination to 200
* Let's add the dates submitted, modified, downloaded, author, submitter, domain and flagged as columns. Clicking on these values should add their values to the filters of the columns.
* The incident id column needs something to indicate that clicking it takes you to that incident. I know you have set the cursor to a pointer, but I think the text should likely also be a link. Maybe adding a clickable icon for the citation view would be helpful, then clicking elsewhere on the incident id could set a filter on the column.

smcgregor (approved):

Issue #243, opened 2021-08-18
César Work Items


smcgregor:
- [x] organize the structure of the application
- [x] pull the social links to the right
- [x] create independent staging system for development and write directions for doing so
- [x] user auth story
- [x] Update donut visualization to new library (e.g., https://naver.github.io/billboard.js/ ). The library is flexible.
- [x] Update react bootstrap version
- [x] Timeline rendering (visualization component that will be capable of showing all reports on a timeline along and on another timeline all incidents by incident date)
- [x] Programmatically updating Algolia
- [x] improve discover application
- [x] improve the landing page
- [ ] geographic rendering of data (create a map)
- [ ] Programmatic translation of reports to other languages

smcgregor:
Most of these are done or in review. Closing this in favor of individual issues.

PR #285, opened 2021-09-25
Replace react-c3 with react-billboard

cesarvarela:
Part of #243


smcgregor:
I think maybe you need to merge master into this branch?

cesarvarela:
@smcgregor updated, the build is failing but I think the reason it is not related to the PR.

smcgregor:
Build is passing at the moment on master. Does it work for you on local? My best guess is that the package lock needs to be updated after merging master in since both involve package updates.

cesarvarela:
@smcgregor it's working now 🎉

smcgregor:
Deploy preview shows a blank screen and `TypeError: Cannot read property 'width' of undefined` in the console.

cesarvarela:
There were some facets with an empty label/value. I'm setting the blank labels to `N/A` lmk if you want to keep that or filter them out.

image

cesarvarela:
Just noticed that it looks empty here too:

image

smcgregor:
"NA" means "not applicable", which may not be the case in some instances. I think this should be fixed in the backend, or labeled as "not assigned". I am not sure why that one case is rendering with zero incidents. It should only render things that have values/counts.

cesarvarela:
Ok, I've added a filter to remove empty facets so we can merge this and fix the backend stuff later.

PR #288, opened 2021-09-29
Add Staging Environment

smcgregor:
Slightly updated version of #274 from @cesarvarela


cesarvarela:
My idea was to have a `staging` branch, merge this PR into it and get the staging env working first.

Then all development could be done against staging and only push to `master` from `staging`.

PR #290, opened 2021-10-06
Add timeline component to show incident reports

cesarvarela:
WIP


smcgregor:
This looks great. One functional comment, multiple popovers can be open at one time at the moment. See: https://deploy-preview-290--staging-aiid.netlify.app/cite/1

And click on the two available popovers.

PR #294, opened 2021-10-11
Upload Algolia index on post build

cesarvarela:
To get this working, the appropriate environment variables need to be set on Netlify:

```
GATSBY_ALGOLIA_APP_ID
GATSBY_ALGOLIA_SEARCH_KEY
ALGOLIA_ADMIN_KEY
```


cesarvarela:
Just noticed that some attributes neeed to be set to searchable when uploading them to Algolia, can you check if I got them all?

This is the UI:
image

Those settings should match these:
https://github.com/PartnershipOnAI/aiid/pull/294/files#diff-641a4bec23ef30314893374b4b34afc0b3ce96d6575bd0c005309fe1a0454d50R154

PR #296, opened 2021-10-14
Improve discover

cesarvarela:
WIP


smcgregor:
This is looking great. The only problem I see is in the dismiss button for the modal popups, which is not rendering properly.

PR #300, opened 2021-10-20
Move Discover filters to the top

cesarvarela:
This will save some space to show more cards.

Builds on top of #296


cesarvarela:
@smcgregor should be ready now

smcgregor:
Mostly good. My quick take feedback is below.

* The split in the filters when the screen is wider looks awkward. Maybe it would be better to make the whole filter set fluid?
* There should likely be a "clear filters" button since the filters are not exposed by default.
* There are a few rough areas in the filters themselves, but I think that was an issue in the old filters as well or at least it is unchanged.

PR #302, opened 2021-10-20
Fix Timeline component crash

cesarvarela:
It also improved how some timelines are rendered because, in some instances, dots overlapped.


smcgregor:
Or a quarter? Regardless, a good conditional render.

PR #316, opened 2021-11-02
Fix incident table

alexmcode:
- [x]     Incident Table View -> Incident Report Table
- [x]     Let's increase the default pagination to 200
- [x]     Let's add the dates submitted, modified, downloaded, author, submitter, domain and flagged as columns.
- [x]     Clicking on these values should add their values to the filters of the columns.
- [x]     The incident id column needs something to indicate that clicking it takes you to that incident. I know you have set the cursor to a pointer, but I think the text should likely also be a link. Maybe adding a clickable icon for the citation view would be helpful, then clicking elsewhere on the incident id could set a filter on the column.

### DIFF
--- a/site/gatsby-site/gatsby-browser.js
+++ b/site/gatsby-site/gatsby-browser.js
@@ -9,6 +9,7 @@ export const onServiceWorkerUpdateReady = () => {
 };
 
 import 'bootstrap/dist/css/bootstrap.min.css';
+import 'bootstrap-daterangepicker/daterangepicker.css';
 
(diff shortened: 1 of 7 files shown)


smcgregor:
@alexmcode We verbally discussed a set of improvements that you will be layering on. Do you want me to merge this in now or wait until the subsequent changes are in? I am good with either.

alexmcode:
@smcgregor I actually would like to do the changes if there is no urgency

alexmcode:
@smcgregor it should be ready for review

PR #324, opened 2021-11-07
added query-string dependency

smcgregor:
This appears to be missing in the builds


smcgregor:
Closing this in favor of https://github.com/PartnershipOnAI/aiid/pull/325 since I had failed to update master prior to applying the fix.

PR #326, opened 2021-11-07
Deploy to Production

smcgregor:


smcgregor:
It looks like the problem was that production was using a different node version from staging. I pinned them both to NPM_VERSION 8.1.0 by setting it as an environment variable on netlify.

PR #328, opened 2021-11-09
Improve Discover filters

cesarvarela:
- Displays refinement list in an overlay instead of just a dropdown 
- Clear button/list
- Better indication of which filters the user applied 
- Prevent refinement items from jumping around when clicking on them


smcgregor:
It looks like if you set the filters from a querystring parameter or by clicking on the incident number of one of the cards, it does not necessarily set the state of the filter in the UI.

cesarvarela:
Yup, that's a side-effect of this one: `Prevent refinement items from jumping around when clicking on them, because you have to provide a sorting function to avoid that, but if the selected item is outside the first N items that we ask to Algolia it won't be listed on the list.

I just pushed a minor update that appends such items at the end, it should work, but it will be missing the count value. lmk if that's ok. If not, I'll remove the sorting algorithm, and it will work as before.

smcgregor:
Still some weirdness if you have an incident ID selected then click the incident ID on one of the cards. It will deselect all incident IDs in the filter.

cesarvarela:
I see the same behavior on prod, not sure when the issue was introduced or if it was always the case. 

 I think it should just set the filter to that incident ID?

cesarvarela:
Pushed a fix for the filters issue

smcgregor:
While looking at this, I noticed the filters on classifications on *production* are not working. @cesarvarela can you take a look?

smcgregor:
* DownloadIndex on production is not giving the classifications, whereas on staging it is.
* Production Algolia thus does not have its classifications
* I have confirmed that the classifications collection is present on both staging and production
* I am not sure why the classifications are absent on export
* I am going to test it out locally against the staging and production backends to see if the behavior differs there. If it does not differ, then there must be a hidden code difference. If it does differ, then something diverged in the backend.

smcgregor:
Fixed here: https://github.com/PartnershipOnAI/aiid/pull/332

smcgregor:
I see what you mean by the counts now. For this to ship, the currently selected filters must be at the top and the counts must reflect how many will be displayed after clicking the filter.

cesarvarela:
Updated to not sort refinements.

smcgregor:
Regarding: `...counts must reflect how many will be displayed after clicking the filter.` If this is not true, you don't know how many reports are associated with each of the potential filters. You may be clicking through to a collection of filters that is not matching any reports.

cesarvarela:
I think what's on this branch already mirrors the production behavior regarding refinement counts. I'm not sure I follow your comment.

smcgregor:
Replicating the bug:

1. In the filters, select incident `34` (I selected this incident arbitrarily).
2. Look at "source" column. In production the maximum count is `3`, but in the deploy preview it is `54`

This is happening because the counts resulting from one filter are not reflected in the other filters.

cesarvarela:
Thanks, I think it does work now.

smcgregor:
Works for me.

This PR increases the importance of #340 ... The default display should either be the details view #340 should land so people will get the details when they click through to the citation record.

Finally, the big block of text when clicking "more..." on the details view either needs whitespace or it needs to be truncated earlier and linked to the full citation record at the end of the text.

PR #331, opened 2021-11-10
Discover view modes

cesarvarela:
Should be ready now, tell me what you think. @smcgregor 

It probably needs to be merged after #328


smcgregor:
Does this still need to be merged after #328 ? It looks like it doesn't build on top of everything 328 has, but your note above mentions it.

cesarvarela:
I think it will conflict but yeah we can merge this one first.

PR #333, opened 2021-11-11
Fix classification workflow support

alexmcode:


smcgregor:
This is looking good. From a functional perspective I am good with merging this, although I would like to improve it more. CSET will be using it at the end of the month and I will test it out as soon as you integrate it with the user auth changes that are landing in the next few days.

**Minor bugs**
* If you scroll to the far right, it displaces the header and you have to scroll to the far left to correct it
* `Reset Filters` is not showing if you are not authenticated

**Improvements**
* The dates can be rendered YYYY-MM-DD and drop the more specific hours/minutes information.
* The cells are almost flush with the left side of the page. If it could go up to the edge it would look a bit better.
* Create anchors to specific taxonomies on the taxa page and link to them.
* Editing directly from the spreadsheet without clicking through to the citation page.

I may have more thoughts after using it on production.

alexmcode:
Hi @smcgregor 
> Reset Filters is not showing if you are not authenticated

Not sure how to reproduce this. On my side it looks like this:

image

> If you scroll to the far right, it displaces the header and you have to scroll to the far left to correct it

The same, can't reproduce

PR #340, opened 2021-11-16
WIP: Fix citation anchors

cesarvarela:


site/gatsby-site/gatsby-browser.js
```diff
@@ -13,29 +13,26 @@ import 'bootstrap-daterangepicker/daterangepicker.css';
 
 import '@fortawesome/fontawesome-svg-core/styles.css';
 
-const ROUTES_WITH_RULES = ['/cite/', '/apps/discover'];
-
 export const shouldUpdateScroll = ({ routerProps: { location } }) => {
   const { pathname, hash } = location;
 
-  if (ROUTES_WITH_RULES.includes(pathname)) {
```
cesarvarela, line 21: To clarify why I'm changing this, this line never worked.

Inside citation pages, `pathname` equals something like `/cite/10` and `ROUTES_WITH_RULES.includes(pathname)` will never match that because it is doing array level matching not string matching.

PR #343, opened 2021-11-19
WIP: Fix flagging

### LINKED ISSUE #301: Flagging Endpoint is Broken
You cannot currently flag incident reports since the backend rejects it with a 500.

smcgregor:
Closes #301 

This is a work in progress. It looks to me like the flagging form may be completely disconnected from any backend endpoint and the flag was previously being set by MongoDB queries. So the next steps are to define a Realm function for flagging content and make that function open to all anonymous users.


smcgregor:
Other pull requests have been reworking this functionality.

PR #346, opened 2021-11-21
Deploy to Production

smcgregor:


site/realm/auth/providers.json
```diff
@@ -14,8 +14,8 @@
         "type": "local-userpass",
         "config": {
             "autoConfirm": true,
-            "resetPasswordSubject": "Password Resets are not Currently Supported",
-            "resetPasswordUrl": "http://aiid.partnershiponai.org/",
+            "resetPasswordSubject": "",
```
smcgregor, line 17: @cesarvarela this should be something like "Your AI Incident Database Password Reset Request"?

cesarvarela: I see that this was merge, but just in case, the message you get when this is set to empty:
image

smcgregor: Thanks! Looks good.

PR #350, opened 2021-11-22
Landing tweaks

cesarvarela:
Some general cleanup of the landing page design/markup.


smcgregor:
Initial feedback:

* I think this may take a few times back and forth, but there are a few ideas in this design we should work to adopt.
* Remove "✌️" from the leaderboards
* Put the counts on the left side so they will all be consistently aligned
* I think it may look better with the header of the boxes more cleanly separated.
* "The Database in Print" either needs its own box or it needs to be at the bottom of the scroll. I think keeping the articles higher up is better.
* Not everyone will know what to search from the landing page. In addition to the search, there should be a button for "Discover Incidents".

Should we adopt a design language like material design and build around it?

cesarvarela:
I pushed a few updates based on your comments.

Material UI could work, yeah, the official react library is https://github.com/material-components/material-components-web (supported by Google), this unofficial one https://mui.com/ and the design guidelines: https://material.io.

It is more explicit than Bootstrap, so there is less room for doing "ugly" stuff.

It will mean reworking most of the UI, though.

smcgregor:
Let's open an issue around adopting MDB, but let's not work this at the moment. It is better to get this incremental improvement landed than engage in a complete redesign.

cesarvarela:
updated with:
- consistent size gutter
- consistent borders for each card
- equal height 'About' cards (it doesn't look pretty, but I couldn't find a way to display them better). Maybe making them horizontal like this?
image

smcgregor:
I think this looks good for the update and will move this through to production after some other items land. Please resolve the conflict from the quickadd fixes.

PR #355, opened 2021-11-25
Map rendering of Location field

cesarvarela:
There needs this env variable to be added with the appropriate value to Netilify:
`GOOGLE_MAPS_API_KEY`

It needs more work to handle `Global`,  `Twitter`, and some other locations, but some early feedback wouldn't hurt.


cesarvarela:
Once everything is fixed, it should look like this: 

image

smcgregor:
It looks like this needs backend changes?

cesarvarela:
Yes needs a  `GOOGLE_MAPS_API_KEY`env var on Netlify. This key should have access to the  Geocoding API.

smcgregor:
You can presume in the future incidents will either be global or have a specific geographic location (city, country, etc.) that can be geocoded. This visually looks good to me, but I have not been able to play with it live yet.

smcgregor:
Closing in favor of https://github.com/PartnershipOnAI/aiid/pull/394

PR #365, opened 2021-11-30
Push translated indexes to Algolia

cesarvarela:
~~This does what it is supposed to do but takes way too long.~~

~~We'll have to choose one of these, in order of complexity (low to high) and speed (slow to fast):~~

- ~~leave it as it is~~
- ~~move this to a separate build process (like how the Realm app is deployed)~~
- ~~move index generation/translation entirely to GCP~~

Another issue I came across is that this requires both GCP and Algolia paid subscriptions. On GCP, you need an account with billing enabled, and on Algolia, you get the `Error: Record quota exceeded. Change plan or delete records.`

Needs a new env var to be added: `GOOGLE_TRANSLATE_API_KEY`


smcgregor:
How long does it take? Either way, don't worry about it at the moment. We are good to pay for the paid account features and can optimize how/when this runs after it works with longer build times.

cesarvarela:
Adding a couple of concurrent queues made it run fast enough to keep it as it is.

cesarvarela:
The total sum of chars of all incident reports: 4549950
Number of languages: 111
Translation cost: 20usd/1000000 chars

Total cost: 4549950/1000000 * 20 * 111 = 10.100,889 USD


https://cloud.google.com/translate/pricing

cesarvarela:
To create a google translate API key just follows this: https://neliosoftware.com/blog/how-to-generate-an-api-key-for-google-translate/

Then this key needs to be set to the `GOOGLE_TRANSLATE_API_KEY` environment variable on Netlify.

To prevent someone from nuking his credit card on GCP, I've also added an extra environment variable to do a dry run of the translation process. Specifically, you have to set `TRANSLATE_DRY_RUN=false` to enable calling the real translate API. If this variable isn't set, it will mock the API call and return `translated-to-{language}` instead.

cesarvarela:
closing this in favor of #376

PR #366, opened 2021-11-30
Show horizontal scroll

alexmcode:
This PR makes horizontal and vertical scrollbars show on any kind of scroll.
On Firefox it looks like this:

image

Chrome does a poorer job at this but it can still be used - it will display the scroll bar on hover:

image

Let me know if you see the same thing in testing

### DIFF
--- a/site/gatsby-site/src/pages/apps/classifications.js
+++ b/site/gatsby-site/src/pages/apps/classifications.js
@@ -39,6 +39,9 @@ const Container = styled.div`
 const TableStyles = styled.div`
   padding-top: 1rem;
 
+  width: calc(100vw - 2.7rem);
+  height: calc(100vh - 220px);
+
   table {
     border-spacing: 0;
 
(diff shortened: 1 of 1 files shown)


smcgregor:
I don't see the horizontal scrollbar on my laptop's Chrome, Safari, or Firefox unless I scroll horizontally on the trackpad first. Can you confirm that you did not accidentally scroll horizontally on your computer? I don't see how adding the lines below may solve the problem.

```
  width: calc(100vw - 2.7rem);
  height: calc(100vh - 220px);
  ```

**Comment by alexmcode** [comment]
I've managed to make scrollbars always visible on chrome and safari. Firefox seems to be less flexible. We might need to use a 3rd party lib: https://stackoverflow.com/questions/6165472/custom-css-scrollbar-for-firefox

**Comment by smcgregor** [comment]
You can't reference `navigator` in SSR since it is not defined.

Would it be better to have the page actually be much much wider and use the natural scrollbars? Does that not make a difference?

**Comment by alexmcode** [comment]
Making the page wider would make the vertical scroll bar disappear to the end of the table. That is the purpose of:
```
 width: calc(100vw - 2.7rem);
 height: calc(100vh - 220px);
```

to make the scroll bars stay at the "ends of the screen"

alexmcode:
@smcgregor managed to go around the need to identify browsers. Scroll should be visible with the mention that in firefox both scrollbars are visible ONLY when any scrolling in any direction. let me know if it's working

PR #368, opened 2021-11-30
Fix boolean values

### LINKED ISSUE #362: Classifications form introduces type inconsistency
When you edit a boolean field and set it to empty, it changes the value to the empty string rather than unsetting the value. This produces a type inconsistency that breaks the build.

```
10:58:30 AM: mongodbAiidprodClassifications.classifications.Lives Lost:
10:58:30 AM:  - type: boolean
10:58:30 AM:    value: true
10:58:30 AM:  - type: string
10:58:30 AM:    value: ''
```

alexmcode:
fixes #362

### DIFF
--- a/site/gatsby-site/src/components/TaxonomyForm.js
+++ b/site/gatsby-site/src/components/TaxonomyForm.js
@@ -358,11 +358,6 @@ const EditTaxonomyForm = ({
         }
       }
 
-      //Convert string values into date
-      if (f.display_type === 'date') {
-        newValues[f.short_name] = `${values[f.short_name]}T00:00:00.000Z`;
-      }
-
       //Convert string into boolean
       if (f.display_type === 'bool') {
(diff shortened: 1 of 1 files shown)


site/gatsby-site/src/components/TaxonomyForm.js
```diff
@@ -362,6 +362,15 @@ const EditTaxonomyForm = ({
       if (f.display_type === 'date') {
         newValues[f.short_name] = `${values[f.short_name]}T00:00:00.000Z`;
       }
+
+      //Convert string into boolean
+      if (f.display_type === 'bool') {
+        if (values[f.short_name] === '') {
+          newValues[f.short_name] = undefined;
```
alexmcode, line 369: @smcgregor if an empty value is set for boolean fields - unsettling - it will add undefined. this line dictates it

PR #376, opened 2021-12-04
Add Language switcher to Discover 

cesarvarela:
Adds a language switcher to the discover app and points the Algolia client to the relevant index.

Builds on top of #294


cesarvarela:
@smcgregor  

Screenshot:
image


This branch deployed to my Netlify env:

https://deploy-preview-12--cesar-aiid-staging.netlify.app/apps/discover?display=details

I've added another env variable to allowlist languages so devs can work on the free tier of Algolia (the max number of indices is 7).

smcgregor:
Much progress here!

Work needed prior to merge:
* Resolve the merge conflicts
* Add details to the readme about translations, including the expected cost of running it, how it decides if it has already been translated, etc. I will use this in my functional review.
* Algolia is more an index than the persistent state of the database. We don't have a backup strategy for Algolia and we are not likely to manage the data on their servers. I think we should post the translations to the MongoDB database rather than Algolia. We may need to discuss this, but I think having a design pattern in the Mongo documents that translated fields have their translated language prepended to them would be good (e.g., `es:title`). Note that the `language` field in the incidents collection should then be interpreted as the original untranslated language, and we will want to replicate incidents that are currently in English to `en:title`/`en:text` fields so when people are viewing English they read the `en:` prepended versions of things. Essentially, this makes the database schema more language agnostic.

Needed before or shortly after merging:
* I have not run the code live yet, but I don't see anything for making the translation sticky in the URL. We should maintain any app state that should be shareable in the URL.
* Anything that is showing translated text should be labeled as showing translated text. This mitigates this potential incident: https://incidentdatabase.ai/cite/72

Needed in the future (don't work on it now):
* Adding support for correcting bad translations.
* Extending this to all 100+ languages.
* We don't know the performance of the Google translate API across all potentially translatable languages. In some cases we will want to avoid presenting poorly supported languages. This may require a simple way of privately previewing languages up until we get someone to 👍  the translation for a language of fluency.
* We need to support entry of data _from_ languages other than English to the other languages.

cesarvarela:
Updated:
- docs explaining the translation process
- persist language in URL via query string (this will need some work in the future once the final i18n implementation is done, because we will probably need to include the current language in the URL path i.e. `/en/apps/discover` because of the static nature of the site.)

smcgregor:
In discussion, it looks like we are going to do the following,

1. The `incidents` collection will contain only the information as it pertains to the incident.
2. Report-specific data will all be moved to a language-specific `incident_reports` collection (e.g., `incident_reports_en`).
3. @cesarvarela is going to look into the best compromise between duplicating fields between the translated incident_reports (e.g., the publication date), versus expressing the database in a more normalized manner similar to a SQL database. Both have their advantages/disadvantages. The current starting point is duplicating the fields, which are lightweight compared to the text of the translations. Data bloat and throughput is not a concern here. Implementation time and minimizing technical debt accumulation is more important.

cesarvarela:
Updated the translation process so no fields are duplicated, the translations db will only store the translated texts:

image

smcgregor:
* Add issue for changing algolia language settings to match the language of the index
* Hide the language switcher, but allow for changing the language in the address bar
* Let me know I should merge this to staging and we will test it out over the next week with the other changes before pushing to production

cesarvarela:
Updates:

Sets `indexLanguages` and `queryLanguages` per index based on the appropriate language:
-  https://www.algolia.com/doc/api-reference/api-parameters/queryLanguages/,
- https://www.algolia.com/doc/api-reference/api-parameters/indexLanguages/

Language switcher is hidden by default, to show it run in the dev tools console:

`localStorage.setItem('i18n', 1)` 

to hide it run 

`localStorage.removeItem('i18n')`

cesarvarela:
Tests have proven useful while wrapping up work in this branch. 

It should be safe to merge; if the `i18n` storage value isn't set, the old Algolia index is used.

These env variables need to be set too:

```
GATSBY_AVAILABLE_LANGUAGES=es,en, ...
MONGODB_TRANSLATIONS_CONNECTION_STRING= (could be the same value as MONGODB_CONNECTION_STRING)
```

smcgregor:
Readme needs details on `MONGODB_TRANSLATIONS_CONNECTION_STRING`, otherwise this is good to go. I'll set staging and production to the following:

```
GATSBY_AVAILABLE_LANGUAGES=en,es
MONGODB_TRANSLATIONS_CONNECTION_STRING=A new user with read/write permissions
```

cesarvarela:
Updated readme

site/gatsby-site/gatsby-node.js
```diff
@@ -178,6 +182,34 @@ exports.onPostBuild = async function ({ graphql, reporter }) {
...
+    config.mongodb.translationsConnectionString &&
+    config.google.translateApikey &&
+    config.google.availableLanguages &&
+    config.header.search.algoliaAdminKey &&
+    config.header.search.algoliaAppId
+  ) {
+    try {
+      if (process.env.TRANSLATE_DRY_RUN !== 'false') {
+        reporter.warn(
+          'Please set `TRANSLATE_DRY_RUN=false` to disble dry running of translation process.'
```
smcgregor, line 195: typo `disble` -> `disable`

smcgregor (approved):

PR #378, opened 2021-12-08
Update to Gatsby v4

### LINKED ISSUE #348: Resolve Dependabot issues
See: https://github.com/PartnershipOnAI/aiid/security/dependabot

cesarvarela:
Closes #348


cesarvarela:
After a smoke test this seems to be working, supposedly the build should be around 40% faster, not sure it is happening.

smcgregor:
Deploy time more than doubles for this, to ~14 minutes from ~6. Is this related to caching and it should speed up in subsequent deploys?

cesarvarela:
The cache wasn't being used because we were missing the `@netlify/plugin-gatsby` package. 

It's there now, but you need to add the environment variable `AWS_LAMBDA_JS_RUNTIME` set to `nodeJS 14.x` on Netlify.

smcgregor:
OK, it is building now, but it is still taking longer. Can you push an update to the readme with the details

`AWS_LAMBDA_JS_RUNTIME`: `nodejs14.x`

and some minor change that will cause the site to re-build? We can confirm that the cache is used via the time being reduced.

cesarvarela:
Pushing to the README file didn't trigger a commit; can you start a redeploy from the UI?

cesarvarela:
Updated the Cypress branch to this one and this appeared on the Netlify log:
image

 So I'd say this is working now.

(it took around 3 mins)

smcgregor:
Excellent, good to squash and merge this now?

cesarvarela:
Clicking around I found that the headings look kind of weird (big):

image

I think this is because I fixed how the styles are applied there, and those should be h2s o h3s instead of h1s. 

Other than that this looks good to merge

PR #384, opened 2021-12-17
Add array editing ui form element

### LINKED ISSUE #375: Improve Array Entry and Other UX Improvements to Forms
Particularly on the classifications edit form.

alexmcode:
Closes #375 

Toggle edit mode on a classifications form in order to see the changes

image

### DIFF
--- a/site/gatsby-site/src/components/TaxonomyForm.js
+++ b/site/gatsby-site/src/components/TaxonomyForm.js
@@ -6,6 +6,8 @@ import { useMongo } from 'hooks/useMongo';
 import { Formik } from 'formik';
 import Loader from 'components/Loader';
 import config from '../../config.js';
+import { faCheck, faTrashAlt } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 
 const ClassificationContainer = styled.div`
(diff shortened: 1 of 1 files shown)


smcgregor:
Hi @alexmcode two comments on the image:

1. It looks like the delete button is only available on the last array item. Is this the case? It should be on all the items.
2. The check mark for adding array items should say "add item" (or something like it). It should also only be clickable after there is input in the form.

If these are both good to go, I will test it behaviorally then look at the code.

alexmcode:
> Hi @alexmcode two comments on the image:
> 
>     1. It looks like the delete button is only available on the last array item. Is this the case? It should be on all the items.
> 
>     2. The check mark for adding array items should say "add item" (or something like it). It should also only be clickable after there is input in the form.
> 
> 
> If these are both good to go, I will test it behaviorally then look at the code.

image

Can't replicate 1. Can you please send a video or more detailed steps?

Fixed 2

smcgregor:
I started the functional testing. When you hit enter while in an array edit form, it submits the entire form rather than adding a single array entry. This is unexpected behavior. It should either block the `enter` key input and not process it, or it should add an item to the array.

cesarvarela:
Closing in favor of #475

PR #395, opened 2022-01-09
Basic JSON API

smcgregor:
This pull requests adds support for JSON clients on observablehq ([example](observablehq.com/@df23e92fd6fce069/zoomable-timeline-of-the-ai-incident-database-v1-0)).

### DIFF
--- a/site/gatsby-site/cypress/integration/submit.js
+++ b/site/gatsby-site/cypress/integration/submit.js
@@ -10,6 +10,8 @@ describe('The Submit form', () => {
   });
 
   it('Should fetch an article data and fill forms accordingly', () => {
+    cy.visit(url);
+
     cy.intercept('GET', parserURL).as('parseNews');
 
     cy.get('input[name="url"]').type(articleURL);
@@ -33,6 +35,8 @@ describe('The Submit form', () => {
   });
 
   it('Should fetch an article data and fill forms accordingly', () => {
+    cy.visit(url);
+
     cy.intercept('GET', parserURL).as('parseNews');
 
     cy.get('input[name="url"]').type(articleURL);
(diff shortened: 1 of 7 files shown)


cesarvarela:
After playing a bit with the observable example, I've managed to query the graphql directly on this fork:

https://observablehq.com/d/7b1a720acfda6a80

The relevant code to create the graphql client:

```
apollo = import("https://cdn.skypack.dev/@apollo/client@3.4.17?min")

Realm = import('https://cdn.skypack.dev/realm-web@1.5.1?min')

client = {  
  const APP_ID = 'aiidstitch2-vsdrv'
  const app  = new Realm.App({id: APP_ID})
  
  async function getValidAccessToken() {
    // Guarantee that there's a logged in user with a valid access token
    if (!app.currentUser) {
      // If no user is logged in, log in an anonymous user. The logged in user will have a valid
      // access token.
      await app.logIn(Realm.Credentials.anonymous());
    } else {
      // An already logged in user's access token might be stale. To guarantee that the token is
      // valid, we refresh the user's custom data which also refreshes their access token.
      await app.currentUser.refreshCustomData();
    }
    return app.currentUser.accessToken
  }
  
  const client = new apollo.ApolloClient({
    link: new apollo.HttpLink({
      uri: `https://realm.mongodb.com/api/client/v2.0/app/${APP_ID}/graphql`,
      // We define a custom fetch handler for the Apollo client that lets us authenticate GraphQL requests.
      // The function intercepts every Apollo HTTP request and adds an Authorization header with a valid
      // access token before sending the request.
      fetch: async (uri, options) => {
        const accessToken = await getValidAccessToken();
        options.headers.Authorization = `Bearer ${accessToken}`;
        return fetch(uri, options);
      },
    }),
    cache: new apollo.InMemoryCache()
  });

  return client
}
```

And to query the graphql endpoint:

```
reports = { client.query({query: apollo.gql`{
    incidents {
      authors
      source_domain
      title
      url
      incident_id
      ref_number
      date_downloaded
      date_modified
      date_published
      incident_date
      submitters
      date_submitted
      report_number
      text
    }
  }`})  
}
```
The good thing is that users will be able to query the database in any way they want, without writing new functions.

The bad thing is that Observable's domain needs to be whitelisted, which is done per user, i.e. `https://cesarvarela.static.observableusercontent.com` (this would be an issue with the JSON API too).

We could remove all the whitelisted domains from the Realm config to deal with this.

smcgregor:
@cesarvarela I checked whether Realm supports wildcards in the cross origin whitelist, but that didn't seem to help. After clearing the whitelist on staging, I can authenticate with your code, but I am not getting the reports to return. Can you share your Observable fork? I think it is set to private.

cesarvarela:
Sorry thought was public by default:

https://observablehq.com/@cesarvarela/zoomable-timeline-of-the-ai-incident-database-v1-0

Didn't check the logic of the chart itself but it is pulling from the graphql endpoint.

PR #407, opened 2022-01-25
Feature timeline date

### LINKED ISSUE #131: Timeline Rendering
A component that can display timelines of incident reports would help in deduplicating incidents, seeing the sequence of reporting about an incident, and track the history of incidents added to the database. We would like to display timelines on several pages.

The proof of concept on this is to show the information pictured below and linked [here](https://incidentdatabase.ai/cite/1) in a timeline. We can look to other use cases and ways of rendering timelines later, but an initial implementation serving the citation page is a good starting point.  
Screen Shot 2021-04-08 at 11 28 51 PM

cesarvarela:
closes #131 

Found this incident with the wrong date (or very good reporters):

image

### DIFF
--- a/site/gatsby-site/src/components/Timeline.js
+++ b/site/gatsby-site/src/components/Timeline.js
@@ -55,7 +55,9 @@ const Line = styled.line`
   stroke-dasharray: 2;
 `;
 
-const Point = styled.circle``;
+const Point = styled.circle`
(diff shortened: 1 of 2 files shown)


smcgregor:
@cesarvarela it is actually very good reporters! In this case Microsoft fired reporters, which set the conditions necessary for the algorithmic reporter to misbehave. Some background reading can be published before the incident happens.

Issue #385, opened 2021-12-18
Add tests to role-based features


cesarvarela:
It should complement #383 but is a little more complex, including setting a testing environment.

Following what's being described here but for Realm auth:

https://docs.cypress.io/guides/testing-strategies/auth0-authentication#Auth0-Application-Setup

smcgregor:
Closing this in favor of https://github.com/PartnershipOnAI/aiid/issues/443

PR #393, opened 2022-01-09
Feature e2e login roles

cesarvarela:
To get this to work 2 new env variables need to be added:

`E2E_ADMIN_USERNAME` and `E2E_ADMIN_PASSWORD`

As the name implies this should belong to a user with the admin role.

builds on top of #390  and #389

PR #415, opened 2022-01-27
Add resources taxonomy

cesarvarela:
TODO:

- [x] Handling errors on form submission
~- [ ] Tests~ needs #385  first
- [x] Standardize `notes` field?
- [x] `Publish` field?
- [x] Custom display of `resources` taxa so it shows the `long_description` field


cesarvarela:
@smcgregor  

There are 2 `notes` fields inside CSET classifications documents. One is a direct property and another inside the `classifications` property (as `Notes`). Are they different? If not, which one should we use?

```
{
  incident_id: 1,
  notes: 'something',
  classifications: {
    Notes: 'something'
  }
}
```

There is also this `Publish` field that is not editable through the `TaxonomyForm` because it is not listed in the Taxa `field_list` property. Should it be added there? When/Where is it edited?

smcgregor:
`Publish` is something that, when set to `true`, should make the classification available to the public. If it is not set to true, or is missing, then the classification should only be available to editors. The field should be available in any form permitting users to write. See: https://github.com/PartnershipOnAI/aiid/issues/392

The notes field is something that is used by taxonomy editors that want to maintain notes that are not displayed to general users. I think it would be best for this field to be a taxonomy field, just like the others, but with a flag saying it is for editors only. However, it was implemented previously as a special case outside the classifications object. I recommend not changing it -- there should only be a `notes` field outside the classifications object. The others were created erroneously.

cesarvarela:
To get this working, this new taxonomy needs to be added to the `taxa` collection:

```
{"_id":{"$oid":"61f158f4c19c105af2f3d6af"},"namespace":"resources","weight":{"$numberInt":"50"},"description":"# What are these resources?\n\nThe following resources have been associated with incidents in the database to provide tools and processes to persons and companies looking for best practices in the prevention or mitigation of similar incidents in the future.","field_list":[{"short_name":"Datasheets for Datasets","long_name":"Datasheets for Datasets","short_description":"Document the datasets use would mitigate or prevent the incident","long_description":"__Lorem ipsum__ dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.","display_type":"bool","mongo_type":"bool","default":null,"placeholder":null,"permitted_values":null,"weight":{"$numberInt":"70"},"instant_facet":true,"required":false},{"short_name":"Publish","display_type":"bool","mongo_type":"bool"}]}
```
And a dummy resource classification needs to be added to the `resources` collection:

```
{"_id":{"$oid":"61f433b874fa62e773d7b764"},"classifications":{"Datasheets for Datasets":false,"Publish":false},"incident_id":{"$numberInt":"0"},"notes":"dasdasd"}
```
An alternative to the dummy would be to just have a real classification, Gatsby needs at least one item to be able to query the collection and build its internal schema.

cesarvarela:
@smcgregor This is ready for you to give it a try

site/gatsby-site/page-creators/createCitiationPages.js
```diff
@@ -49,172 +50,180 @@ const getClassificationsArray = (incidentClassifications, taxonomy) => {
...
-                    mongodb_id
-                    text
-                    authors
-                  }
-                }
+const createCitiationPages = async (graphql, createPage) => {
+  const result = await graphql(
+    `
+      query IncidentIDs {
+        allMongodbAiidprodIncidents(filter: { incident_id: { ne: 0 } }) {
```
smcgregor, line 57: Why `ne: 0`?

site/gatsby-site/src/graphql/classifications.js
```diff
@@ -0,0 +1,122 @@
...
+`;
+
+export const FIND_CSET_CLASSIFICATION = gql`
+  query FindCSETClassifications($query: ClassificationQueryInput) {
+    classifications(query: $query) {
+      _id
+      incident_id
+      notes
+      classifications {
+        AIApplications
```
smcgregor, line 38: I have usually seen this with underscores separating the words. Is this the more common pattern for keys with multiple words in graphql?

smcgregor (commented):
There is a lot to test here. @cesarvarela how confident/tested are you in this change set? I can either concentrate on reviewing this in `staging`, or review it on the branch after whitelisting it.

site/gatsby-site/page-creators/createCitiationPages.js
cesarvarela: that's to ignore a "dummy content" incident. It is probably a leftover, I'll delete it

site/gatsby-site/src/graphql/classifications.js
cesarvarela: Graphql convention is camel case, but since there are no spaces it leaves it as it is.

Fields with spaces are joined and camel-cased: `Datasheets for Datasets` => `DatasheetsForDatasets`

cesarvarela:
After some smoke testing, I'd say it is safe to merge. 

I would like to add some tests but this requires #393  and having a reproducible test environment since this mutates the database.

PR #418, opened 2022-01-29
Show taxonomy data to anonymous users

cesarvarela:
This partially fixes taxonomy visibility but could use a little bit more work.


smcgregor:
@cesarvarela merged to staging instead of master, I'll deploy if this looks good on staging

PR #438, opened 2022-02-07
Fix scroll to contents

### LINKED ISSUE #422: Scroll-to is broken on Contents
Anchors like this `https://incidentdatabase.ai/research#citingthedatabaseasawhole` are not scrolling to show the expected content at the top.

Screen Shot 2022-02-02 at 9 43 38 PM

cesarvarela:
I don't know when this broke, probably in between Gatsby upgrades.

There is an official plugin: `gatsby-remark-autolink-headers` that solves this better, but the link ids that it generates replace spaces with dashes so `Something Interesting` becomes `something-interesting` instead of `somethinginteresting` which is what the actual Table of contents component uses.

closes #422


smcgregor:
@cesarvarela merging. Make sure you update your local staging to match the current staging since I recreated it from master.

PR #441, opened 2022-02-08
fix scroll test

smcgregor:
This changes the scroll heights to exactly those as found when running it on my laptop. Builds are currently failing on production because of these tests. If they still fail I will likely update the data on the staging server to match the current production since there is a cross dependency there.


smcgregor:
FYI @cesarvarela

smcgregor:
@cesarvarela I increased the tolerance on the test so both production and staging would work.

PR #447, opened 2022-02-11
Add instructions on how to add new taxa

### LINKED ISSUE #338: Make Adding Taxonomies to UI Easier
There are currently hard coded values for the CSET taxonomy that should first be collected into a single location in the graphql, then (ideally) removed from graphql hardcoding entirely. The first stage on this issue is to collect it so that it is hard coded in only one location, then we can look to eliminating it from the code entirely.

cesarvarela:
closes  #338

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,9 +270,9 @@ test:e2e:ci
 ### To add new taxonomies, follow these steps:
 Let's say you want to add the `CTECH` Taxonomy.
 
-1. Create a new collection using the lowercased Taxonomy name: `ctech`
+1. Create a new MongoDB collection using the lowercased Taxonomy name: ctech.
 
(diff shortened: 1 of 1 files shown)


smcgregor:
I think this also needs pointers to the GraphQL API in the codebase?

cesarvarela:
Yes, you are right, I added a 4th step, which could be automated by building the graphql queries with data pulled from the taxa collection, but it is probably not worth it.

smcgregor (approved):

smcgregor:
@cesarvarela The collection also needs to be listed in `gatsby-config.js`?

```
    resolve: 'gatsby-source-mongodb',
    options: {
      dbName: 'aiidprod',
      collection: [
        'incidents',
        'submissions',
        'quickadd',
        'duplicates',
        'taxa',
        'classifications',
        'resources',
      ],
```

PR #456, opened 2022-02-16
UI Improvements

cesarvarela:
Includes these two, tests, and a few structural improvements.

-  In /apps/classifications, it’s a bit annoying to have two clicks to edit (edit button -> edit); it would be better to go straight to full edit mode. If the idea is to let people view the classification, could add a "view" button to give access to the main modal.
-  In apps/classifications, could there be a "submit" (or "save") button at the top of the edit interface as well as the bottom?


cesarvarela:
@smcgregor ready for review

site/gatsby-site/src/components/taxa/Taxonomy.js
```diff
@@ -0,0 +1,187 @@
...
+                {taxonomy.classificationsArray
+                  .filter((field) => {
+                    if (showAllClassifications) return true;
+                    if (!showAllClassifications && field.weight >= 50) {
+                      return true;
+                    }
+                    return false;
+                  })
+                  .filter((field) => {
+                    if (field.name === 'Datasheets for Datasets' && field.value == 'No') {
```
smcgregor, line 126: We are going to be adding 10s to 100s of the resources in the next few months. I don't think we can have these taxonomy-specific references sprinkled in the codebase. If some special logic is needed, it should likely be added to the taxa collection defining the specific taxonomy.

site/gatsby-site/cypress/fixtures/classifications/upsertCSET.json
```diff
@@ -0,0 +1,94 @@
+{
+    "data": {
```
smcgregor, line 2: What is the lifecycle of the fixtures when running tests locally versus in the build system? This may need to be added to the README.

cesarvarela: In this case, it is just a fancy way to load a JSON file; it will always do the same thing no matter the env.

site/gatsby-site/src/components/taxa/Taxonomy.js
cesarvarela: Then we have to add another property that allows us to overwrite the way classifications are rendered.

Maybe adding an optional `render` property that if defined lets you set a different rendering method:

```
 {
         "short_name":"Datasheets for Datasets",
         "long_name":"Datasheets for Datasets",
         "short_description":"Document the",
         "long_description":"Lorem ipsum.",
         "display_type":"bool",
         "mongo_type":"bool",
         "default":null,
         "placeholder":null,
         "permitted_values":null,
         "weight":{
            "$numberInt":"70"
         },
         "instant_facet":true,
         "required":false         
         
        "render": "long_description" | "some other render method"
}
```
If omitted the actual value will be used, which is the current method.

What do you think?

smcgregor: This is the intention of the `display_type` field.

cesarvarela: Yeah but this is a special case where the `display_type` is different when editing than when displaying the field:

If editing, it is presented as a yes/no select.
If displaying the value, the `long_description` field is shown.

Maybe use a `display_type: "long_description"`?

Issue #458, opened 2022-02-16
Many reports have the same report number


smcgregor:
`{report_number: 1335 }` on the incidents collection produces 187 results. This value is supposed to be the global auto-incrementing primary key for reports and needs to be migrated to unique values. It looks like the problem is the report promotion function which is not incrementing the value -- it is getting the last value and repeating it.

PR #459, opened 2022-02-16
increment report number

smcgregor:
Addresses the problem for future promoted reports, but still need to address the legacy issue. #458


smcgregor:
@cesarvarela do you want to do this migration with a new strategy, or should I do this as another Python script? https://github.com/PartnershipOnAI/aiid/tree/master/scripts/db_migrations

PR #461, opened 2022-02-18
Migrations

### LINKED ISSUE #458: Many reports have the same report number

cesarvarela:
Adds `umzug` to Gatsby's build process and runs any migration found in the `/migrations` folder.

Adds a new migration `2022.02.18T16.29.14.increment-report-number.js`, which goes through the list of incidents sorted by `report_number`, and when it finds a duplicate, it updates the `report_number` field of the remaining incidents incrementally.

closes #458


smcgregor:
This looks chaotic to me when moving through staging and production. Maybe we check the [context](https://docs.netlify.com/configure-builds/environment-variables/#build-metadata) to see whether the build is on the production branch and only run the migration then?

cesarvarela:
Updated to check for `CONTEXT=='production'`

cesarvarela:
Just noticed that migrations must be called on Gatsby's `onPreBootstrap` build step so the changes it makes are available to the actual build.

README.md
```diff
@@ -357,6 +357,34 @@ Let's say you want to add the `CTECH` Taxonomy.
...
+## Database Migrations
+Migration files are stored in the `/site/gatsby-size/migrations` folder and their executions are logged in the `migrations` collection.
+
+### Configuration
+Please add a connection string with read/write permissions to the mongo database:
+```
+MONGODB_MIGRATIONS_CONNECTION_STRING=
+```
+### Execution
+Migrations run on Netlify after the build process finishes, but it is also possible to run them from the command line for debugging or testing purposes.
```
smcgregor, line 369: This documentation line is no longer accurate.

site/gatsby-site/migrations/2022.02.18T16.29.14.increment-report-number.js
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ *
+ * @param {{context: {client: import('mongodb').MongoClient}}} context
+ */
+
+exports.up = async ({ context: { client } }) => {
+  await client.connect();
+
+  const collection = client.db('aiidprod').collection('incidents');
```
smcgregor, line 11: In the future, these should likely be loaded from config. No need to change them at the moment.

cesarvarela:
Updated.

PR #462, opened 2022-02-19
removed defunct summaries contents

smcgregor:


smcgregor:
A test is failing, but I believe it is unrelated to this fix and manual testing of the behavior checks out.

Issue #455, opened 2022-02-16
UI improvements


cesarvarela:
- [x] Would be helpful to be able to navigate to /apps/classifications from the website (after logging in)
- [x] In /apps/classifications, it’s a bit annoying to have two clicks to edit (edit button -> edit); it would be better to go straight to full edit mode. If the idea is to let people view the classification, could add a "view" button to give access to the main modal.
- [x] In apps/classifications, could there be a "submit" (or "save") button at the top of the edit interface as well as the bottom?
- [ ] It would be helpful to be able to select multiple incidents and change certain things simultaneously - most obviously "stage"
- [x] For the fields where multiple choices are possible, would it be possible to use checkboxes (rather than command/control to select multiple), to ensure annotators know that they can choose multiple?
- [ ] After logging into the database, there is no button to log out in the UI

PR #473, opened 2022-02-23
Replace Select component with a checkbox group in Taxonomy Form

cesarvarela:
image

The list sometimes gets a little long but gives all the available items visibility. It could be made scrollable with a max height if that's better

part of #455


smcgregor:
Looks good, except the prompt is no longer accurate (don't use the ctrl key).

cesarvarela:
Removed

cesarvarela:
Maybe do the same but for enum types and using radio buttons instead?

smcgregor:
Yes, sounds like a good change. Should I hold this PR or merge this one first?

cesarvarela:
Yes, let's merge this one as it is.

PR #484, opened 2022-03-01
Move parsing endpoint to a Gatsby function

### LINKED ISSUE #467: Move parsing news endpoint to a Netlify function
Considering that everything is already set up to use Netlify functions through Gatsby, the only thing left would be to find a library that extracts article data.

Something like https://github.com/ageitgey/node-unfluff.

This would fix the cold start problems of the current endpoint.

cesarvarela:
Closes #467


cesarvarela:
It seems to perform similarly; sometimes, it succeeds in some attributes where the old one fails and vice versa, but overall I think this is a net positive because it seems much more stable (I had the old one crash while testing a couple of times when it should have been wide awake already.)

cesarvarela:
Needs to be whitelisted it seems.

site/gatsby-site/src/api/parseNews.js
```diff
@@ -0,0 +1,21 @@
...
+import { format, parseISO } from 'date-fns';
+
+export default async function handler(req, res) {
+  const { url } = req.query;
+
+  const article = await Mercury.parse(url, { contentType: 'text' });
+
+  const response = {
+    title: article.title,
+    authors: [article.author],
```
smcgregor, line 11: What if there are multiple authors? I think the form is still input as a CSV here?

smcgregor:
One comment on the code side. If you are good with the behavioral testing, I'll merge after you reply and test this and other items in staging tomorrow.

site/gatsby-site/src/api/parseNews.js
cesarvarela: Good question. Do you have an example?

smcgregor: Two that may get treated differently

https://www.wsj.com/articles/ukraine-war-leads-to-more-global-business-disruption-11646173915

https://www.theguardian.com/world/2022/mar/01/russias-war-in-ukraine-complete-guide-in-maps-video-and-pictures

PR #491, opened 2022-03-02
Fix Submit form

### LINKED ISSUE #490: Submit form is Broken
It looks like there is a problem in the submit form in staging. I found it after merging to production, but I cancelled the build so the issue is not live.

It looks like there is now an inconsistency in the values being pushed to the backend with the update. Specifically,
```
Version currently deployed to staging and sitting on master:
      "authors": "Test Author", # this is not an array

Current working version,
      "incident_id": 0, # this is missing on staging
      "tags": [], # this is missing on staging
      "authors": [ # this is an array
        "Beth Mole"
      ],
```

I think the form submission (preferred) or the server side function needs updating.

cesarvarela:
closes #490 

This happened because I'm mocking the backend response, and mocks tend to drift away from actual implementation.

Eventually would be nice to have a full e2e test environment.

### DIFF
--- a/site/gatsby-site/cypress/integration/submit.js
+++ b/site/gatsby-site/cypress/integration/submit.js
@@ -41,12 +41,13 @@ describe('The Submit form', () => {
       expect(xhr.request.body.arguments[0]).to.deep.include({
         title: 'YouTube to crack down on inappropriate content masked as kids’ cartoons',
         submitters: 'Something',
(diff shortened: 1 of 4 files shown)


cesarvarela:
Also tweaked the form behavior a little bit, so it is easier to trigger the form's validation and prevent the state from being cleared on error.

PR #496, opened 2022-03-08
Normalize incidents collection

cesarvarela:
@smcgregor please take a look


smcgregor:
I'll review, please update for the conflict as appropriate. Note to self: the build will not pass until staging is migrated according to this change set.

site/gatsby-site/migrations/2022.03.04T15.44.49.incidents-normalization.js
```diff
@@ -0,0 +1,107 @@
...
+        title,
+        url,
+        tags,
+        flag,
+      } = report;
+
+      const newReport = {
+        _id,
+        incident_id,
+        authors: authors || [],
```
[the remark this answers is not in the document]
cesarvarela: I couldn't find any, but this is a failsafe to prevent mongo to store `null` values which break the Graphql endpoint.

site/gatsby-site/cypress/integration/downloadIndex.js
```diff
@@ -0,0 +1,92 @@
...
+
+  const downloadsFolder = Cypress.config('downloadsFolder');
+
+  it('Should return a properly constructed Algolia index', () => {
+    cy.visit(url);
+
+    cy.query({
+      query: gql`
+        query {
+          reports(limit: 9999) {
```
smcgregor, line 15: It may be better to set this low for tests so the test will run faster.

site/gatsby-site/cypress/integration/downloadIndex.js
```diff
@@ -0,0 +1,92 @@
...
+            date_submitted
+            description
+            epoch_date_downloaded
+            epoch_date_modified
+            epoch_date_published
+            epoch_date_submitted
+            epoch_incident_date
+            flag
+            image_url
+            incident_date
```
smcgregor, line 30: incident_date is still on the reports collection? It should be joined in to the rows of the reports, but it should not be in the same collection as reports post-normalization.

site/gatsby-site/cypress/integration/downloadIndex.js
```diff
@@ -0,0 +1,92 @@
...
+          }
+        }
+      `,
+    }).then(({ data: { reports } }) => {
+      cy.get('[data-cy=download]').click();
+
+      cy.readFile(path.join(downloadsFolder, 'index.json'), { timeout: 15000 }).then((index) => {
+        expect(index).to.have.length(reports.length);
+        expect(index.find((r) => r.report_number == 19)).to.deep.equal({
+          objectID: '5d34b8c29ced494f010ed46c',
```
smcgregor, line 48: Maybe you need to get the whole index to ensure this record is in the set?

site/gatsby-site/migrations/2022.03.04T15.44.49.incidents-normalization.js
```diff
@@ -0,0 +1,107 @@
...
+        date_downloaded,
+        date_modified,
+        date_published,
+        date_submitted,
+        description,
+        epoch_date_downloaded,
+        epoch_date_modified,
+        epoch_date_published,
+        epoch_date_submitted,
+        epoch_incident_date,
```
smcgregor, line 53: This should be on the incident collection, not the reports collection.

smcgregor (commented):
* I will need to trigger a backup from the workflow before merging this since it is deleting many records.
* We need to address the locking/mutation issue since we will soon have multiple editors in the system. Please open an issue for this and we will talk on Wednesday about when we schedule it for work.

After we close the discussion threads on this, let's merge to staging and do a thorough functional evaluation there. Too much downstream code is depending on this at the moment to block with a non-staging branch review. Time to get this to staging then production!

site/gatsby-site/cypress/integration/downloadIndex.js
cesarvarela: Tested just in case, but it doesn't seem to make that much of a difference, maybe once there are more reports.

site/gatsby-site/cypress/integration/downloadIndex.js
cesarvarela: Yeah, I was thinking that it could be done in another PR since it should also use MongoDB's Date data type,  and that would require more changes yet. But if you think it's better to do it all at once I can add that change here too.

site/gatsby-site/cypress/integration/downloadIndex.js
cesarvarela: Not sure I follow. This tests downloads the entire index JSON file and  checks that the report number 19 has proper structure and values.

site/gatsby-site/migrations/2022.03.04T15.44.49.incidents-normalization.js
cesarvarela: Yes, same as my comment above, I was thinking to update it in another PR, too many changes at once otherwise.

cesarvarela:
Add some comments

site/gatsby-site/cypress/integration/downloadIndex.js
smcgregor: If you set the number of records to return to a small number, then this specific record might not be returned.

smcgregor:
Squashing and merging to staging. This will blog deployments to production until everything is tested thoroughly in user space, but it is time we start that testing.

Issue #414, opened 2022-01-26
Incident has wrong occurance date


smcgregor:
https://incidentdatabase.ai/cite/2?lang=en

It should be December 5th for all the reports.

smcgregor:
Updated. It will change in the next deploy.

PR #502, opened 2022-03-19
Set epoch vals on update

smcgregor:
Related to #414


smcgregor:
Merging this now since it won't deploy the realm function to staging without this sitting on staging.

PR #510, opened 2022-03-25
Add resources taxa to Algolia index

### LINKED ISSUE #487: Non-CSET Facets are not Pushed to Algolia
When `"instant_facet":true,` on the taxonomy collection, it is expected that that resource would be searchable in the Algolia index, but I don't believe it is getting updated in the index.

cesarvarela:
- Adds `resources` taxa to Algolia index
- Adds new field `MSFT_AI_Fairness_Checklist`
- Adds unit test to the Algolia index generator util

image

closes #487 

builds on top of #496


smcgregor:
I changed the base from staging to master and back and it is still showing a diff on files that are unchanged after I merged the upstream. Can you merge staging into your branches to clear the diff?

cesarvarela:
Update the missing test fails because it needs a Realm schema update.

PR #526, opened 2022-04-13
Support markdown formatting on reports' text property

### LINKED ISSUE #519: Generate Markdown in Reports
When you use the fetch endpoint for parsing news articles, the returned text is a single big block of text.

cesarvarela:
This turned into adding support for markdown on the report's text property.

- `reports.text` and `submissions.text` is now stored as markdown
- `plain_text` field added to `submissions` and `reports` collections
- `text` field is pushed as plain text to the Algolia index
- translations store `text` as markdown since from my tests Google's Translate API can handle it well
- a migration updates every `submission` and `report` to normalizes the newline handling

- added a markdown editor 
image

(the custom resolver option was trickier because realm doesn't handle external npm dependencies very well)

_tests wont's pass until migration runs_

closes #519


smcgregor:
Blocking

* Merging this: https://github.com/responsible-ai-collaborative/aiid/pull/535 it is fixing a current issue in production so I won't add something that could block deploying from staging.

Not blocking

* We can't use the Google API to translate Markdown. I don't believe it will process links properly. For instance, how will the translation API know which word or words in the translation corresponds to the link? This may be an issue we need to open rather than blocking this from rolling forward in the short term. We can choose to strip links, etc. up until we have a solution.
* It looks like it is reinterpreting the currently plain text database as markdown, which should generally work but there is a possibility of something hiding among the reports that is not intended to be markdown. We will need to check for weird things and weird formatting across the whole text database.
* People are consuming the text of the project for natural language processing (NLP) purposes. Those models should not be processing markdown content. Can the GraphQL endpoint optionally return markdown or plaintext?
* People need to be told in the submission form that they can submit markdown
* At some point in the future we may want to re-parse the historical reports to have its original formatting, this should look significantly better.

Other Comments

* Displaying non-markdown content in the discover app makes sense. People will click through to view with formatting.
* This should merge after addressing the one blocking issue.
* This comment is formatted with Markdown ;)

cesarvarela:
Updated

smcgregor:
@cesarvarela I see this is marked for review. Should I review it? Builds are currently failing.

site/gatsby-site/src/components/taxa/Taxonomy.js
```diff
@@ -57,10 +57,6 @@ const Taxonomy = ({ taxonomy, incidentId }) => {
   const canEdit =
     isRole('taxonomy_editor') || isRole('taxonomy_editor_' + taxonomy.namespace.toLowerCase());
 
-  if (!canEdit && taxonomy.classificationsArray.length < 1) {
```
smcgregor, line 60: Why did you remove this? I think this makes the taxonomy not show up in the event no values are associated with the taxonomy. This is the desired behavior.

site/gatsby-site/src/global.css
```diff
@@ -911,3 +911,16 @@ img {
     padding-left: 0px;
   }
 }
+
+/* disable Github Permalink */
+.bytemd-toolbar-right [bytemd-tippy-path='5'] {
```
smcgregor, line 916: What is this doing?

site/gatsby-site/src/utils/AlgoliaUpdater.js
```diff
@@ -59,9 +63,12 @@ class AlgoliaUpdater {
       for (const report_number of incident.reports) {
         const report = reports.find((r) => r.report_number == report_number);
 
+        const text = await remark().use(remarkStrip).process(report.text);
```
smcgregor, line 66: Should use the plain_text field instead, no?

site/gatsby-site/src/utils/Translator.js
```diff
@@ -83,9 +85,15 @@ class Translator {
...
       .collection(`incident_reports_${language}`);
 
-    const fields = [...keys, 'report_number'];
+    const translated = [];
 
-    const translated = items.map((item) => lodash.pick(item, fields));
+    for (const item of items) {
+      const { report_number, text, title } = item;
+
+      const plain_text = (await remark().use(remarkStrip).process(text)).contents.toString();
```
smcgregor, line 93: Why not using the plain_text field here? Also, I thought translations used the markdown...

site/gatsby-site/src/utils/discoverIndexGenerator.js
```diff
@@ -146,13 +150,16 @@ module.exports = async ({ graphql }) => {
         }
       }
 
+      const text = await remark().use(remarkStrip).process(report.text);
```
smcgregor, line 153: Same here, why not plain_text?

site/gatsby-site/src/components/submissions/SubmissionForm.js
```diff
@@ -110,17 +113,9 @@ const SubmissionForm = () => {
 
   const [parsingNews, setParsingNews] = useState(false);
 
-  const coldStartToast = () => {
-    addToast({
```
smcgregor, line 114: Cold starts are no longer a problem then?

smcgregor:
A few questions/comments inline, otherwise 👍

site/gatsby-site/src/global.css
cesarvarela: The Markdown Editor has a button that links to the GitHub project, and that's the only way to hide it at the moment.

site/gatsby-site/src/components/taxa/Taxonomy.js
smcgregor: We are going to have a few of these taxonomies through time and we don't want to show them if they are blank and the user can't fill them in. I would take a re-render over displaying it to all users.

cesarvarela: Oh got it 👍

site/gatsby-site/src/utils/AlgoliaUpdater.js
cesarvarela: Checked again and no, because that `report` object is built from the `reports` collection + the translated `text`. This will need to be updated later to take into account the `report.language` field though.

site/gatsby-site/src/utils/Translator.js
cesarvarela: That is storing the already translated text

site/gatsby-site/src/components/submissions/SubmissionForm.js
cesarvarela: Yup

site/gatsby-site/src/utils/discoverIndexGenerator.js
cesarvarela: updated

site/gatsby-site/src/components/submissions/SubmissionForm.js
cesarvarela: I mean, it when from the 20 to 30 secs to only a few.

site/gatsby-site/src/components/taxa/Taxonomy.js
cesarvarela: Updated, refactored it a little bit, the re-render will only happen to authenticated users.

cesarvarela:
Updated. Tests won't pass until the migration is run.

smcgregor:
@cesarvarela It looks like submit.js and submitted.js conflict now. Can you resolve and we will merge Wednesday?

PR #552, opened 2022-04-23
Fix wrong edit incident link

cesarvarela:
Makes sure there is a default value for `incindetId` and `reportNumber`. 

Using Gatsby's dynamic routing might be more elegant but this would do for now.

### DIFF
--- a/site/gatsby-site/src/templates/cite.js
+++ b/site/gatsby-site/src/templates/cite.js
@@ -181,7 +181,7 @@ function CitePage(props) {
                   <Button
                     variant="outline-primary"
                     className="me-2"
-                    href={'/incidents/edit?incident_id=' + incident.incident_id}
+                    href={'/incidents/edit?incidentId=' + incident.incident_id}
                   >
                     Edit Incident
                   </Button>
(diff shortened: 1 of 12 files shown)


site/gatsby-site/src/templates/cite.js
```diff
@@ -181,7 +181,7 @@ function CitePage(props) {
                   <Button
                     variant="outline-primary"
                     className="me-2"
-                    href={'/incidents/edit?incident_id=' + incident.incident_id}
+                    href={'/incidents/edit?incidentId=' + incident.incident_id}
```
smcgregor, line 184: Parameters should mirror their expression in the field definition. This should remain `incident_id` and this should be `const [incidentId] = useQueryParam('incident_id', withDefault(NumberParam, 1));` no?

cesarvarela: Sure, it is my js mind that doesn't like snake_case, updating.

smcgregor: https://github.com/responsible-ai-collaborative/aiid/issues/554

cesarvarela:
Ready. Also updated the `report_number` field for consistency.

Issue #558, opened 2022-04-25
Refactor client-side data access to only use Realm's  graphQL endpoint


cesarvarela:
The MongoDB database is being accessed in 3 different ways: 

- Realm's client-side MongoDB client (the submissions component does this)
- Realm's custom functions (`createNewReportforReview`, `quickAdd`, etc.)
- Realm's graphQL endpoint

Ideally, all data access should be done through one option, the 3rd option, Realm's grapqhQL endpoint.

PR #575, opened 2022-04-28
Use submissions graphql instead of custom context

### LINKED ISSUE #571: Related Reports Broken
See "Incident reports matched by ID". It is empty, when there should be one report there.

Screen Shot 2022-04-27 at 10 53 46 PM

CC @cesarvarela -- I think you are likely to be resolving this when you harmonize the APIs.

cesarvarela:
It's part of #558 

closes #571

The test is failing because there are 2 reports with the same `report_number`:

image

The empty message mentions something about incidents with a +/- month of difference but the code does an ID check, looks like this text is out of date?

image


cesarvarela:
Updated, the diff is a little bit dirtier than it could, but at least everything is green.

smcgregor:
You have a few PRs open at the moment, is there are review order I should follow?

cesarvarela:
It probably doesn't matter. I also fixed the test that was letting a submission slip through.

cesarvarela:
Now that I think about it, the Realm stuff will probably conflict. 

I'll make one PR with everything.

PR #588, opened 2022-05-05
Add a test to check db integrity

cesarvarela:


cesarvarela:
@smcgregor not sure it is a good idea to merge this, but it helped uncover one of the causes of the broken build: report `1563` is not linked to any incident.

smcgregor:
I think that is the report that had a duplicate report_id and I updated it on the backend during our call. If I add it to the appropriate incident collection array, it should be good?

cesarvarela:
hopefully yes

smcgregor:
A test fixture landed in the staging DB, it is this one: https://github.com/responsible-ai-collaborative/aiid/blob/5400eecdf5a91663090850356e313e85e79c338a/site/gatsby-site/cypress/fixtures/submissions/submitted.json

I deleted the record in the DB and will confirm it does not get re-created when merging the other PRs to staging.

PR #590, opened 2022-05-07
Add custom resolvers to link/unlink reports

### LINKED ISSUE #522: Create a custom mutation resolver to promote submissions to new incidents
We want to run as many operations as possible in one graphql call to avoid possible race conditions

cesarvarela:
closes #522


site/realm/functions/unlinkReportsFromIncidents.js
```diff
@@ -0,0 +1,8 @@
+exports = async (input) => {
+
```
smcgregor, line 2: Let's keep the implementation, but raise an error at present since we don't want to have orphaned reports.

site/gatsby-site/src/components/ReportedIncident.js
```diff
@@ -79,100 +61,109 @@ const ReportedIncident = ({ incident: report }) => {
...
-    delete newReport.mongodb_id;
-
-    delete newReport._id;
-
     const {
       data: {
-        lastIncident: [{ incident_id: lastIncidentId }],
-        lastReport: [{ report_number: lastReportNumber }],
-        refsNumbers,
+        promoteSubmissionToReport: { 0: incident },
```
smcgregor, line 79: Why is this `{ 0: incident }`? Is this something the Realm function unpacks as an input variable?

smcgregor:
@cesarvarela I have one question on this inline, but I am going to merge it now. There is a lot to unpack in this one, great work/improvements!

site/gatsby-site/src/components/ReportedIncident.js
cesarvarela: The resolver can handle promoting a submission to a report that's linked to many incidents,  but at the moment we are only promoting to only, so it's getting that one.

PR #593, opened 2022-05-11
Trigger render on mount to update citation date

cesarvarela:


cesarvarela:
@smcgregor I think you need to trigger a build here

smcgregor:
I turned off the ability to run builds on anything except master for the product build server, and staging won't run if you target master. Can you move this through staging?

cesarvarela:
Now that I think about it, this isn't happening on `staging so this PR does nothing :/

cesarvarela:
Sorry for the back and forth, this might fix it. The theory is that the build starts failing once a day has passed, and a build is triggered from something that doesn't come from a code change and doesn't cause Netlify's cache to be rebuilt, like the hourly cron.  So we have 2 options:

- merge this into master and wait for a day
- merge this into staging, and once staging is into master, wait a day

smcgregor:
Sounds good. Merging!

PR #594, opened 2022-05-11
Change hits per page to 28

### LINKED ISSUE #538: Tiling on Discover App
- Tiling is 4 per row and then the last row only has 2, I would either load 28 instead of 30, or change to 3/row max.

Issue from Joshua Lugo

cesarvarela:
closes #538


smcgregor:
It looks like a test needs an update here.

cesarvarela:
Ready

PR #595, opened 2022-05-11
Remove incident_date from edit form

### LINKED ISSUE #548: Remove incident_date from edit report form
Considering that incidents have now their form and eventually incidents and reports will have a many-to-many relationship.

image

cesarvarela:
closes #548


smcgregor:
@cesarvarela I think this needs a bit more work. I still need to trace the flow of assignment on `incidents` and `submissions`, but this is likely the last changeset required to properly normalize the incidents and reports collections. The problem this changeset seems to introduce is ambiguity on the properties of the incident when adding a report for a new incident. Right now, the submitter determines the incident_date and the incident_date is set when migrating from `submission` to `incidents` and `reports`. However, if the incident_date is not in the submissions form, then there are not properties the submitter can set associated with new incidents. I think we need to have separate components for the incident form and the report form and both forms should be presented to the user when making a submission.

cesarvarela:
There is still the `incident_id` field in the report's schema that is used in a couple of places too (nothing too critical I believe).

 I guess I could remove it in this PR too, alongside using separate component for submissions.

smcgregor:
I am good with increasing the scope of this PR to include that.

Note: I plan on deploying to production this weekend and ingesting a bunch of reports. I think we are good to do this after?

cesarvarela:
yup 👍

smcgregor:
It looks like this issue is currently breaking staging, which is blocking deployment. I added a new incident in staging, and the incident_date is not set. The `app/incidents` table also does not allow for setting the date on incident 180, which is the offending incident. This makes the build fail.

Deploy: https://app.netlify.com/sites/staging-aiid/deploys/62807791dc5fe76862046f9b

cesarvarela:
https://github.com/responsible-ai-collaborative/aiid/pull/610 would prevent newly created issues to have  `null` values (eventually #611 would totally avoid this) and also allow you to set the `date` field on incident 180.

Technically this PR is not needed to move forward with #599 (though it is my curren working item)

smcgregor:
@cesarvarela should we migrate all the missing fields to have empty values? Half the incidents don't have fields like the alleged developer/deployer fields.

cesarvarela:
Note to self: undo #615

cesarvarela:
> @cesarvarela should we migrate all the missing fields to have empty values? Half the incidents don't have fields like the alleged developer/deployer fields.

yeah either that or set defaults everywhere in the UI, The migration will be probably easier and #611 would prevent for it to happen again

site/gatsby-site/src/utils/AlgoliaUpdater.js
```diff
@@ -46,30 +46,37 @@ class AlgoliaUpdater {
...
-        ...report,
-      };
+    const downloadData = [];
 
-      if (classificationsHash[report.incident_id]) {
-        finalDataNode.classifications = classificationsHash[report.incident_id];
-      }
+    for (const incident of incidents) {
+      for (const report_number of incident.reports) {
+        const report = reports.find((r) => r.report_number == report_number);
```
smcgregor, line 60: This looks a bit wasteful. I think it may be worth opening an issue for it so we can land this.

site/gatsby-site/src/pages/summaries/incidents.js
```diff
@@ -1,105 +1,97 @@
...
-      allMongodbAiidprodReports: { group },
-    } = data;
+export default function Incidents({ data, ...props }) {
+  const incidents = data.allMongodbAiidprodIncidents.nodes.map((incident) => {
+    const reports = incident.reports.map(
+      (report_number) =>
+        data.allMongodbAiidprodReports.nodes.find((r) => r.report_number == report_number) || {
+          report_number,
+          title: `Missing Report ${report_number}`,
+        } // there are missing reports, remove once that's fixed
```
smcgregor, line 52: What is happening here?

cesarvarela: I forgot to flag this.

I'll have to look a little more into this, but there seem to be a couple of incidents that contain reports that do not exist:

- `report_number` is present in the `incidents.reports` array 
- there is no report in the `reports` collection with that `report_number.`

It is probably because of the inconsistency bug we had a while ago; it might be good to add a test for it next to the other integrity tests.

site/gatsby-site/src/components/leaderboards/UniqueSubmittersLeaderboard.js
```diff
@@ -19,9 +27,12 @@ const UniqueSubmittersLeaderboard = ({ limit, className }) => {
 
   const submitters = {};
 
-  for (const report of nodes) {
+  for (const report of reports) {
+    const { incident_id: id } = incidents.find((incident) =>
+      incident.reports.includes(report.report_number)
```
smcgregor, line 32: The use of `find` throughout this PR is problematic and gets worse through time.

site/gatsby-site/src/components/submissions/SubmissionForm.js
```diff
@@ -0,0 +1,282 @@
...
+
+// set in form //
+// * title: "title of the report" # (string) The title of the report that is indexed.
+// * text: "Long text for the report" # (string) This is the complete text for the report in the MongoDB instance, and a shortened subset in the Algolia index
+// * date_downloaded:`2019-07-25` # (Date) Date the report was downloaded.
+// * submitters: Array(string) # People that submitted the incident report
+// * authors: Array(string) # People that wrote the incident report
+// * date_published: `2019-07-25` # (Date or null) The publication date of the report.
+// * image_url: "http://si.wsj.net/public/resources/images/BN-IM269_YouTub_P_2015051817" # (string) The URL for the image that is indexed. This will be stored on the server as a hash of the URL.
+// * url: "https://blogs.wsj.com/digits/2015/05/19/googles-youtube-kids-app-criti" # The fully qualified URL to the report as hosted on the web.
```
smcgregor, line 24: Also `incident_date`

smcgregor: and tags

smcgregor:
A few inline comments, but I don't see any issues overall. I am going to merge this to staging and do a full interactive review of core functionality. I think we need to open issues for the following, @cesarvarela do you agree?

* Replace the usage of `find` with data layer lookups or with indices
* Resolve the missing reports question

I'll add the docstrings at merge time when resolving the BibTex conflict.

smcgregor:
Closed in favor of https://github.com/responsible-ai-collaborative/aiid/pull/650

PR #598, opened 2022-05-12
Improve incidents table view

### LINKED ISSUE #584: Improve Incident Table App
#541 introduces a table view for the incidents collection. I am going to merge it to staging, but I am not going to place it in the navigation yet.

Improvements to be made,

(blocking general availability)
- [x] When applying a tag in the form, the tag is not applied if you do not terminate the tag by pressing enter
- [x] Make the incident number a link to the citation page
- [x] The loading spinner should present a more friendly loading picture
- [x] The `Actions` column should be hidden if you are not logged in
- [x] Zebra the rows so it is easier to read

(good to have but not blocking)
- [x] The table header should float as you scroll down the page
- [x] Should have the filtering functionality of `/apps/classifications`
- [ ] Expand the row to display more text after clicking on the row
- [x] While a record is updating, the modal's controls should be disabled
- [x] Toast indicating the record has been updated should trigger the modal to disappear
- [x] The `save` and `close` button should be on the same line, with one pulling left and the other pulling right. It may also be better to label the `save` button as `update`.

Screen Shot 2022-04-29 at 5 58 02 PM

cesarvarela:
Tackles most UX improvements from the issue except the expanding rows.

Eventually there should be feature parity with the other table apps  (classifications)

closes #584

### DIFF
--- a/site/gatsby-site/cypress/integration/apps/incidents.js
+++ b/site/gatsby-site/cypress/integration/apps/incidents.js
@@ -4,7 +4,7 @@ import incident from '../../fixtures/incidents/incident112.json';
 
 import updateOneIncident from '../../fixtures/incidents/updateOneIncident112.json';
 
-describe('Classifications App', () => {
+describe('Incidents App', () => {
   const url = '/apps/incidents';
 
   it('Successfully loads', () => {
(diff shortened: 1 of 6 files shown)


smcgregor:
Code review looks good. I don't know all the APIs you are using, but I don't see any issues. I am going to merge to staging and test drive the functionality along with the other changes that have landed recently. If the play testing goes well, we may add this app to general availability.

PR #615, opened 2022-05-17
add date to incident collection on promotion

smcgregor:
This is a stopover until #595 lands


cesarvarela:
Looks good, added a note to self to remove this once #595 lands

cesarvarela (approved):

Issue #607, opened 2022-05-16
(Issue Ladder: 3) Write a Cypress Test for Existing Functionality


smcgregor:
note: This is an issue that should not be closed. The purpose of the issue ladder is to provide a set of activities that bring new developers up to speed on the codebase and collaborating with current developers.

**Why**

Programmatic testing ensures that the build system does not produce a non-functional website. The purpose of this ladder issue is to teach you about how to do test driven development with the AIID, and in particular the role served by Cypress as a test runner.

**What**

Pick a currently untested page of the application and create a new test suite for it. See [here](https://github.com/responsible-ai-collaborative/aiid/tree/master/site/gatsby-site/cypress/integration) for what is currently tested. Open a pull request with the new test coverage. We recommend you run the Cypress tests locally (the browser interface is very fancy) before pushing to run on AIID infrastructure.

PR #619, opened 2022-05-18
Add wordcount.js integration test

lmcnulty:
Fulfilling [#607](https://github.com/responsible-ai-collaborative/aiid/issues/607), tests that:
- table and word cloud both exist 
- words in table have length > 2 and count > 10.

### DIFF
--- a/site/gatsby-site/cypress/integration/wordcounts.js
+++ b/site/gatsby-site/cypress/integration/wordcounts.js
@@ -5,9 +5,18 @@ describe('The Word Counts Page', () => {
 
   it('word count table should exist', () => {
     cy.visit('/summaries/wordcounts');
-    cy.get('.row .list-group');
-    cy.get('.row .list-group > div');
-    cy.get('.row .list-group > div > span');
+    cy.get('[data-cy=wordlist-container]').should('exist').and('be.visible');
+
+    // Check that there are multiple rows
+    cy.get('[data-cy=wordlist-container] > * > *').its('length').should('be.gt', 1);
+  });
+
+  it('word cloud should exist', () => {
+    cy.visit('/summaries/wordcounts');
+    cy.get('[data-cy=wordcloud]').should('exist').and('be.visible');
+
+    // Check that there are multiple text nodes
+    cy.get('[data-cy=wordcloud] text').its('length').should('be.gt', 1);
   });
 
   it('words should have length > 2 and count > 10', () => {
@@ -15,18 +24,19 @@ describe('The Word Counts Page', () => {
 
     // Pull out the words and their counts and then check
     // that the minimum word length and occurence count are accurate
-    let counts = [];
+    const counts = [];
 
-    let words = [];
+    const words = [];
 
-    cy.get('.row .list-group > div')
+    cy.get('[data-cy=wordlist-container] > * > *')
       .each((row) => {
         words.push(row.text().trim());
+
+        const countText = row.find('span').text().trim();
+
+        const spanIsNumeric = /\d+/.test(countText);
+
         let count = 0;
-
-        let countText = row.find('span').text().trim();
-
-        let spanIsNumeric = /\d+/.test(countText);
 
         if (spanIsNumeric) {
           count = parseInt(countText);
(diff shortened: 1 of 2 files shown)


site/gatsby-site/cypress/integration/wordcounts.js
```diff
@@ -0,0 +1,47 @@
+describe('The Word Counts Page', () => {
+  it('successfully loads', () => {
+    cy.visit('/summaries/wordcounts');
+  });
+
+  it('word count table should exist', () => {
+    cy.visit('/summaries/wordcounts');
```
smcgregor, line 7: These CSS-based selectors should detect whether the table as it currently exists is present or not, but if/when the display of the table changes this test will naturally fail. I think this test as a basic "did it load" test would be fine just checking for the header text of `Word Counts`. Your next test, which looks at the contents of the table, is then looking at the more specific question of the contents of the table.

site/gatsby-site/cypress/integration/wordcounts.js
```diff
@@ -0,0 +1,47 @@
...
+      })
+      .then(() => {
+        const ascending = (a, b) => a - b;
+
+        expect(words.map((word) => word.length).sort(ascending)[0] > 2).to.be.true;
+        expect(counts.sort(ascending)[0] > 9).to.be.true;
+      });
+  });
+  it('word cloud should exist', () => {
+    cy.get('.col-8 svg');
```
smcgregor, line 44: These CSS selectors are likely an indication that the word clouds should have a handle that can be checked for that is not related to the templating. Maybe a CSS class that has no associated styling (at present) would be good to add?

smcgregor (commented):
Two inline comments. @cesarvarela do you agree with my comments? Please feel free to review this one as you have written a lot more Cypress than I have.

site/gatsby-site/cypress/integration/wordcounts.js
```diff
@@ -0,0 +1,47 @@
+describe('The Word Counts Page', () => {
+  it('successfully loads', () => {
+    cy.visit('/summaries/wordcounts');
+  });
+
+  it('word count table should exist', () => {
+    cy.visit('/summaries/wordcounts');
+    cy.get('.row .list-group');
+    cy.get('.row .list-group > div');
```
cesarvarela, line 9: I agree with the comment above. Also, this seems to be missing assertions? It is getting the DOM nodes but not doing anything with them:

Something like `cy.get('.row .list-group').should('exist');`

cesarvarela: https://docs.cypress.io/api/commands/should

site/gatsby-site/cypress/integration/wordcounts.js
cesarvarela: Yes, cypress recommends using the `data-cy="something"` attribute to keep tests separate from CSS. It is not expected to add this to every component, but only the top ones.

https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

site/gatsby-site/cypress/integration/wordcounts.js
```diff
@@ -0,0 +1,47 @@
...
+        let count = 0;
+
+        let countText = row.find('span').text().trim();
+
+        let spanIsNumeric = /\d+/.test(countText);
+
+        if (spanIsNumeric) {
+          count = parseInt(countText);
+        }
+        counts.push(count);
```
cesarvarela, line 34: Looks weird to me to have this much logic in a test, but maybe there is no better way to do it. Ideally, it should be easier to tell what the test it is doing.

nitpick: variables that don't change should be declared as `const` instead of `let`.

site/gatsby-site/cypress/integration/wordcounts.js
cesarvarela: Yeah, the test will fail, but leaving it like that, it is not clear what you are testing. Because that test will pass no matter how many items there are or even if they are visible or not, as long as there is at least one present on the DOM, but that's a little bit open-ended.

You probably want something like `cy.get('.row .list-group').should('have.length', 2).and('be.visible')` (didn't test it).

smcgregor (approved):
Approved, but @cesarvarela is the strongest in Cypress so I will let him do a final review.

smcgregor:
@lmcnulty feel free to resolve review discussion points in the event you believe it likely that the person opening the discussion is satisfied with the changes you introduced to address the discussion.

cesarvarela (approved):

Issue #613, opened 2022-05-17
Missing and broken links


lmcnulty:
In addition to [#537](https://github.com/responsible-ai-collaborative/aiid/issues/537), there are several other places in the UI where elements that seem like they should be clickable are not:

- [x] The header title → Should link to the root page

  <img src="https://user-images.githubusercontent.com/25443411/168822650-59c6d56d-d8a3-4420-87e5-d8123cda29c4.png" width="300"/>

- [x] Blog post table of contents items → Should scroll to the corresponding heading. This only occurs on the blog posts, not the other pages. The cause seems to be that the headings don't have IDs. #672
  
  <video src="https://user-images.githubusercontent.com/25443411/168826150-14d3a7c3-25ae-4097-a722-45725736ddbd.mp4">

---
- The wordcount items → Should link to Discover with the search term
  
  <img src="https://user-images.githubusercontent.com/25443411/168823756-fdafa652-55e1-4440-a561-6edc65e1f66f.png" height="400"/>

  This probably requires first that wordcount words are lemmatized instead of stemmed.

PR #621, opened 2022-05-18
Add headerTitle to link

lmcnulty:
This fixes the first item of [#613](https://github.com/responsible-ai-collaborative/aiid/issues/613) where the site title in the header was not a link to the homepage.

The divider is changed from a `<li>` to a `<div>` because when inside of a `<Link>` it was displaying with a bullet to the left, and in any case it doesn't seem like a list item semantically.


smcgregor:
Deploy preview looks good in Chrome, FF, and Safari. Merging to staging...

Issue #627, opened 2022-05-20
Date picker for Date fields in report submission


LKchemposer:
Please provide a date picker for date fields (i.e., `Incident Date`, `Date Published`, and `Date Downloaded`) instead of an input text. It'll help reduce time spent on manually converting various datetime systems (e.g., American vs. others) into `YYYY-MM-DD` format, and errors associated with it.

smcgregor:
Just merged this to staging (thanks @lmcnulty). It should land in production in about a week. Closing this now, but if you experience issues after the deployment please reopen.

LKchemposer:
The input for `Date Published` and `Date Downloaded` is not a date picker, still 😞 

Date picker for `Incident date` works wonders! 😄

lmcnulty:
Once again, this should be solved once #664 makes it into production.

PR #629, opened 2022-05-20
Add type="date" to date fields on IncidentReportForm.js

lmcnulty:
Per #627, adds the browser's default date picker (as used in IncidentForm.js) to text fields on the submission form.


smcgregor:
Tested on the deploy preview and it looks like it is working. I am a bit surprised since the presentation order of the date changes.

Merging, Khoa should give feedback when it lands in production and decide whether to close #627

Issue #633, opened 2022-05-23
Disable Scroll on Incident Submission Form incident_id


smcgregor:
When you are editing the incident number on an incident, if you scroll while your cursor is on the incident_id field, it will scroll the numeric field. This is unexpected and should be disabled.

PR #638, opened 2022-05-24
Make id input non-scrollable

lmcnulty:
This solves #633. An "id" input type is added for non-scrolling numeric inputs. The "number" input type is left alone, as the default browser behavior is potentially desirable when the values represent quantities rather than IDs.


cesarvarela (approved):
I think this is making the `onWheel` event unavailable for any other component type? It is rarely used, but who knows.

I would use a `disableWheel` boolean flag that doesn't tie the type `id` with wanting to disable this behavior.  

But I'm okay with merging as it is.

smcgregor:
Ok, @lmcnulty please ping me after you resolve the merge conflict.

Issue #635, opened 2022-05-23
Default image on report submission appears unexpectedly


lmcnulty:
In addition to #624, there are some UX issues with submitting a report
The default report image doesn't show up until after you unfocus the image url input field. This is awkward as it's very large and pushes the other elements out of the way. It's also unlabelled, so it's not necessarily clear what happened. This is especially jarring when using the tab key to advance to the next field, as shown in the video below:

  https://user-images.githubusercontent.com/25443411/169911491-8bf3566d-7671-48a1-a5b3-31017dafd02e.mp4

I think the default image should be small, labelled, and visible on the initial page load.

smcgregor:
Agreed. I think shrinking the image and having the placeholder displayed until there is a new candidate image sounds good to me.

PR #640, opened 2022-05-24
Auto-update image preview, add label, and set max size

lmcnulty:
Solves #635. Shows fallback image by default, also resulting in image preview updating as url is typed rather than after focus exits in the input. Sets a max-height on the preview and adds a label:

image


cesarvarela (commented):
I think the issue is somewhere else because the fallback handling is done here: https://github.com/responsible-ai-collaborative/aiid/blob/fba1626eceeaeac22e5fba1c25b4a82e76a5304b/site/gatsby-site/src/utils/cloudinary.js#L26.

Looks more like a render cycle problem, but maybe it is not a rabbit hole worth getting into.

I would like to see a couple of tests checking each outcome of this change though (the `integrations/submit.js` page seems a good candidate).

smcgregor:
LGTM

@cesarvarela : you have been spending time in the submit testing code. Can you take final review on this one?

smcgregor:
@lmcnulty can you resolve the merge conflict then ping @cesarvarela for final review?

site/gatsby-site/cypress/integration/submit.js
```diff
@@ -486,6 +486,29 @@ describe('The Submit form', () => {
...
+
+    const newImageUrl = 'https://' + suffix;
+
+    const cloudinaryImageUrl =
+      'https://res.cloudinary.com/pai/image/upload/d_fallback.jpg/f_auto/q_auto/v1/reports/' +
+      suffix;
+
+    cy.visit(url);
+    cy.get('input[name=image_url]').type(newImageUrl);
+    cy.wait(1000);
```
cesarvarela, line 508: This `wait` might introduce some flakiness, have you tried something else? Especially on CI where everything runs slower. 

I'm assuming the test fails without it?

(If there is no better way let's go with it)

cesarvarela (approved):

site/gatsby-site/cypress/integration/submit.js
lmcnulty: Ah! Good catch. That was leftover from a previous state and is no longer necessary. I've removed it in the last commit.

smcgregor:
I started the deploy preview and will merge if it passes.

Issue #617, opened 2022-05-18
Alignment and Spacing


lmcnulty:
- [x] In the header, the GitHub star button is not vertically aligned with the RSS and Twitter buttons.
  https://github.com/responsible-ai-collaborative/aiid/pull/648
  
  <img src="https://user-images.githubusercontent.com/25443411/169071020-048843df-35d9-4921-a467-f0e63e63de82.png" height="40"/>

- [x] When scrolled down, there's an awkward space between the footer and bottom of the sidebar. The sidebar should go down at least as far as the footer, and possibly even all the way to the bottom of the page. 
  https://github.com/responsible-ai-collaborative/aiid/pull/646
  
  ![scrolled-sidebar-space svg](https://user-images.githubusercontent.com/25443411/169076649-69a78721-acbc-422b-99e8-573c328d9360.png)

- [x] In the sidebar, there's space below the "Discover · Submit" that looks somewhat awkward.  
  https://github.com/responsible-ai-collaborative/aiid/pull/623
  
  <img src="https://user-images.githubusercontent.com/25443411/169075861-5a73e453-647b-4b78-90f1-03dedb42f791.png" height="300">

- [x] In the sidebar, the external link icon is spaced too far from the link text. It's also not aligned vertically with the text, although that could be a valid design choice if it's being treated like a superscript – provided it's consistent throughout the site. In such case, it should be 1ch to the right of the link text.  
  https://github.com/responsible-ai-collaborative/aiid/pull/647

  <img src="https://user-images.githubusercontent.com/25443411/169071088-847966c2-60bf-43e5-8776-258668b35bcf.png" height="40"/>

---

- ~~On Firefox, the "Discover · Submit" buttons render incorrectly, with the text running too close to the left edge.~~ **Nevermind**: I'm not sure what causes this, but it doesn't happen on a clean Firefox profile.

  <table>
  <tr>
    <td>Firefox</td>
    <td><img src="https://user-images.githubusercontent.com/25443411/169080847-38472606-c561-4b2e-ac91-825c283ae175.png"  width="150"></td>
  </tr>
  <tr>
    <td>Chromium</td>
    <td> <img src="https://user-images.githubusercontent.com/25443411/169080836-705fe270-69db-4dc8-a582-c3ec53eaf102.png" width="150"></td>
  </tr>
  </table>

PR #648, opened 2022-05-25
Header icon alignment

lmcnulty:
To solve the first item of #617, it was easier to re-implement the `react-github-btn` than to get it to correctly align with the fontawesome icons.

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>


smcgregor:
Visually this looks good.

One implementation change: the ground truth on the GitHub Url should live in only one place. Right now it is in the config file and also this change set. I recommend either adding a new config value for the repo's API address, or re-processing the GitHub URL as it is currently configured into the API address.

smcgregor (approved):

PR #649, opened 2022-05-25
Add BibTeX copy button

### LINKED ISSUE #644: BibTeX copy button
If a user is viewing the BibTeX citation, it seems like their intention will almost always be to copy and paste it into their bibligraphy. It would be nice to have a button to that would [`writeText()`](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) to the clipboard.

<table>
<tr><th>Current</th><th>With copy button</th></tr>
<tr><td></td>
<td></td></tr>
</table>

Another option would be to make the citation text a `<textarea>` so that `ctrl+a` `ctrl+c` would be a quick way to copy it.

lmcnulty:
Resolves #644. Toast z-index set to 1 greater than the modal overlay.

https://user-images.githubusercontent.com/25443411/170362857-f3c8c305-583c-41e9-bf5d-9440f57f45d2.mp4

### DIFF
--- a/site/gatsby-site/src/contexts/ToastContext.js
+++ b/site/gatsby-site/src/contexts/ToastContext.js
@@ -10,6 +10,7 @@ const ToastsWrapper = styled.div`
   bottom: 2em;
   left: 2em;
   max-width: 100% !important;
+  z-index: 1056;
 `;
 
 const CloseButton = styled(Button)`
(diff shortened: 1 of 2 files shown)


site/gatsby-site/src/contexts/ToastContext.js
```diff
@@ -10,6 +10,7 @@ const ToastsWrapper = styled.div`
   bottom: 2em;
   left: 2em;
   max-width: 100% !important;
+  z-index: 1056;
```
smcgregor, line 13: Ha. Good fix.

smcgregor (approved):

Issue #630, opened 2022-05-20
Discover filter UX


lmcnulty:
- [x] [theverge.com](https://theverge.com) and www.theverge.com should be treated as the same source. The solution might be to just strip `^www\.` from domains names, and possibly other domain prefixes that are usually not contented-related, such as `^m\.`

  <img height="200" src="https://user-images.githubusercontent.com/25443411/169567159-9c164626-4be6-49f5-b2fe-4d62fdf3e57d.png">

- [ ] Ideally, the taxonomy selection should be tree-based. Until then, it should at least be ordered alphabetically so that all the possibilities for e.g. CSET:Intent: are grouped together. The current top categories, like "Lives Lost:False," seem like the least interesting ones.
  <table>
    <tr><th>Current</th><th>Tree-based</th>
    <tr>
      <td><img src="https://user-images.githubusercontent.com/25443411/169568233-4e8e3888-b2d1-4150-b1fd-c7ce8e5048dd.png"></td>
      <td>
<video width="250" src="https://user-images.githubusercontent.com/25443411/172436662-140c313b-af07-4a63-8b32-567bbed7610c.mp4">
</td>
    </tr>
  </table>

- [ ] The classification selection UI shows categories like "CSET:Sector of Deployment:" with no value after the last colon. This seems to pick out items where the "CSET:Sector of Deployment" is not set – but it's not clear from the UI that this is what it does. In fact, I initially assumed that it was the reverse, selecting the items for which it *is* set.

- [ ] It seems unlikely to me that end users would want to find an incident by ID, and if they did, it might be more natural to order them numerically by the ID rather than by the number of reports. When I saw this the first time, I didn't understand why there would be multiple reports for the same incident, since I assumed that I was searching through incidents rather than reports.

  <img height="200" src="https://user-images.githubusercontent.com/25443411/169566457-656ca5e0-2d23-4624-8de5-0ea4750a5599.png"/>

- [ ] Not a high priority, but it can be nice to select a date range with sliders
  <table>
  <tr><th>Current</th><th>Example</th></tr>
  <tr>
  <td><img height="200" src="https://user-images.githubusercontent.com/25443411/169565455-081ab5ae-4ade-4eba-bd9c-3fa390979236.png"></td>

  <td><img height="200" src="https://user-images.githubusercontent.com/25443411/169565837-69ffe729-4637-4f9a-bc32-98d9c7178da3.png"></td>
  </table>
[...]

PR #655, opened 2022-05-31
Discover ux hit actions

### LINKED ISSUE #537: Actions for search results are below the fold
On the search results/discover page, the titles and header images of the results do not link to the associated incident pages, as one might expect:

image

To reach either, a user needs to scroll to the bottom of the result card, which, because of the long descriptions, can be an entire screen's height:

image

To go to the original source article, the user has to click an unlabelled icon, the meaning of which is not obvious:

image

This issue could be resolved if the title, image, and domain indicator for each card were links, like so:

expectation

---

Similarly, on the incident page, the article title is a link to the ID of the card containing it, which is not particularly useful, and the domain label is non-interactive.

image

To reach the source link, the user once again has to scroll to the bottom of the card, which in this case means scrolling through the full article text:

image

lmcnulty:
Resolves #537. Changes:

- Makes result header images and titles into links to the report text on our site
- Make the subtitle containing the domain into a link to the original source
- Removes blockquotes from descriptions
  - These are usually redundant and take up vertical space.
  - In compact view, they also sometimes push the titles outside of the card
- Filters options are start out collapsed under a "filter search" `<summary>`.
  - This adds enough vertical space that the result titles appear above the fold.
  - On mobile, the modal filter dialogue is still used
  - If we want to highlight the filter options more, another possibility would be to put them in a sidebar
    - It might be better to de-emphasize the filters until we've solved some of the other UX issues (see #630).
    - I think that the "Classifications" filter is much more interesting than any of the others. Perhaps we could put it in the first row but hide the others behind the click?

## Before

The result titles are mostly below the fold, and they are not clickable.

https://user-images.githubusercontent.com/25443411/171270128-e700e2db-7751-4c4c-b542-2234638a4c6b.mp4

## After

Titles are above the fold, and header images, titles, and source domains are links.

https://user-images.githubusercontent.com/25443411/171270335-786d281f-282c-4cf5-85fb-7d5aaaedf0b8.mp4

### DIFF
--- a/site/gatsby-site/src/components/discover/hitTypes/Details.js
+++ b/site/gatsby-site/src/components/discover/hitTypes/Details.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import { Card } from 'react-bootstrap';
-import { Highlight, Snippet } from 'react-instantsearch-dom';
 import { Image } from 'utils/cloudinary';
 import styled from 'styled-components';
 import { fill } from '@cloudinary/base/actions/resize';
@@ -8,6 +7,8 @@ import { fill } from '@cloudinary/base/actions/resize';
 import md5 from 'md5';
 import { navigate } from 'gatsby';
 import Actions from '../Actions';
+
+import { SourceDomainSubtitle, HeaderTitle } from './shared';
 
 const IncidentCardImage = styled(Image)`
   height: ${({ height }) => height};
@@ -19,10 +20,6 @@ const StyledLabel = styled.p`
   margin: 0.6em 0;
 `;
 
-const TitleHighlight = styled(Highlight)`
-  font-weight: bold;
-`;
-
 export default function Details({
   item,
   authorsModal,
@@ -30,30 +27,22 @@ export default function Details({
   flagReportModal,
   toggleFilterByIncidentId,
 }) {
+  console.log(item);
   return (
     <Card className="h-100" data-cy={item.mongodb_id}>
-      <IncidentCardImage
-        className="card-img-top"
-        publicID={item.cloudinary_id ? item.cloudinary_id : `legacy/${md5(item.image_url)}`}
-        alt={item.title}
-        height="240px"
-        transformation={fill().height(480)}
-      />
+      <a href={'/cite/' + item.incident_id + '#r' + item.objectID}>
+        <IncidentCardImage
+          className="card-img-top"
+          publicID={item.cloudinary_id ? item.cloudinary_id : `legacy/${md5(item.image_url)}`}
+          alt={item.title}
+          height="240px"
+          transformation={fill().height(480)}
+        />
+      </a>
       <Card.Body className="d-flex flex-column ">
-        <Card.Title>
-          <TitleHighlight hit={item} attribute="title" />
-        </Card.Title>
+        <HeaderTitle item={item} />
 
-        <Card.Subtitle className="mb-2 text-muted">
-          {item.source_domain} &middot;{' '}
-          {item.date_published ? item.date_published.substring(0, 4) : 'Needs publish date'}
-        </Card.Subtitle>
-
-        {item._snippetResult.text.matchLevel === 'full' && (
-          <blockquote className="mt-4">
-            <Snippet attribute="text" hit={item} />
-          </blockquote>
-        )}
+        <SourceDomainSubtitle item={item} className="mb-2 text-muted" />
 
         <Card.Text className="flex-fill">{item.text.substr(0, 400) + '...'}</Card.Text>
 
@@ -63,7 +52,7 @@ export default function Details({
               type="button"
               className="btn btn-secondary btn-sm w-100"
               onClick={() => {
-                navigate(`/cite/${item.incident_id}#r${item.mongodb_id}`);
+                navigate(`/cite/${item.incident_id}`);
               }}
             >
               <StyledLabel>Show Details on Incident #{item.incident_id}</StyledLabel>
--- a/site/gatsby-site/src/components/discover/Filters.js
+++ b/site/gatsby-site/src/components/discover/Filters.js
@@ -2,10 +2,7 @@ import React from 'react';
 import REFINEMENT_LISTS from 'components/discover/REFINEMENT_LISTS';
 import { Row, Col } from 'react-bootstrap';
 import Filter from './Filter';
-import Stats from './Stats';
 import styled from 'styled-components';
-import ClearFilters from './ClearFilters';
-import DisplayModeSwitch from './DisplayModeSwitch';
 
 const StyledFilter = styled(Filter)`
   .dropdown-toggle {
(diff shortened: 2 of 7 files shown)


smcgregor:
Design review: the filters should be expanded on page load if a filter is applied. I'll look closer at the code after that fix is in, and play with it interactively.

smcgregor:
Build is failing. Can you try merging staging into this branch to see if that fixes it? I plan on playing with the functionality then returning to the code review.

site/gatsby-site/src/components/discover/hitTypes/Details.js
```diff
@@ -19,41 +20,29 @@ const StyledLabel = styled.p`
...
-`;
-
 export default function Details({
   item,
   authorsModal,
   submittersModal,
   flagReportModal,
   toggleFilterByIncidentId,
 }) {
+  console.log(item);
```
smcgregor, line 30: Leftover logging?

smcgregor:
I played with the deploy preview and there appears to be a bug. Replicating:

1. Open the discover application
2. Apply a filter
3. Refresh the page
4. (bug 1): the filters are not expanded and visible
5. Expand the filters
6. (bug 2): the color of the applied filter is not changed, but the number of filters applied is properly indicated

site/gatsby-site/src/components/discover/hitTypes/Details.js
lmcnulty: Good catch. I wonder if it would be worth it to add `no-console` to the eslint config for src/components and src/pages?

lmcnulty:
The "[Generate a Algolia index](https://app.netlify.com/sites/staging-aiid/deploys/62a3351f87e31b0009385fd3#L640)" test is now failing... I think that's caused by recent, unrelated changes, since #685 failed it as well.

If that's not an issue, I think this is ready for another review / merge.

smcgregor:
1. `Controls` should be implemented in the `Filters` component. You moved some of the functionality into `Discover` page.
2. Rather than list out every refinement to check for, you should rely on a single source of truth and pull from `import REFINEMENT_LISTS from 'components/discover/REFINEMENT_LISTS';` as is done to render the `StyledFilter` in Filters.js. ([example](https://github.com/responsible-ai-collaborative/aiid/blob/master/site/gatsby-site/src/components/discover/REFINEMENT_LISTS.js#L24)).

Please let me know if I am missing something with (1) and (2) above. Once they are addressed, please ask @cesarvarela to review and provide an explanation for how you landed on using a `click` event on first render. He may have ideas there.

lmcnulty:
Regarding the feedback:

1. I somewhat disagree that `Controls` belongs in `Filters`. The parts that I moved to `Discover` were `Stats`, `DisplayModeSwitch`, and `ClearFilters`. The first two are basically unrelated to the filters – they would work just as well if there were no filtering functionality. `ClearFilters` is logically filter-related, but it's displayed independent of the filters. At least for now, it's positioned in the row above the filters, and it's shown greyed-out even when the filters are not rendered. So none of those components feel like they're part of the filters component.
  
    However, I think the underlying issue you were getting at was that `discover.js` had grown too long. So I moved `Controls` to its own file rather than to adding it to `Filters`. Is that ok?

2. Thanks, I missed that the refinements were coming from there. I've made the change.

Regarding the use of the `.click()` event: After I had switched to looking at the `query` rather than the `searchState`, the filters were correctly displaying on page load when one or more was applied. However, they were displaying in a single column regardless of the viewport width. I didn't look into it carefully enough to figure out for sure why that was occurring, but hiding and then showing the filters again caused them to display correctly. I'm guessing that the responsive design depended on the js `window` object, which was unavailable in server-side rendering, so it would only display correctly when a client-side render was triggered. `.click()` seems to accomplish that, although perhaps there's a better way?

site/gatsby-site/src/components/discover/Controls.js
```diff
@@ -0,0 +1,72 @@
...
+      let filter = false;
+
+      for (let refinement of REFINEMENT_LISTS) {
+        if (query[refinement.attribute]) {
+          filter = true;
+          break;
+        }
+      }
+      if (filter) {
+        document.querySelector('#expand-filters').click();
```
cesarvarela, line 38: Direct DOM manipulation is a big no-no! Worst case use a ref, but here seems like you would want to use a state flag.

cesarvarela (changes requested):

smcgregor:
It should likely be filters are in controls which is referenced in discover. I think you did that here. :)

It looks like @cesarvarela has some suggestions on the click.

site/gatsby-site/src/components/discover/Controls.js
lmcnulty: I've changed it to use `setExpandFilters(true)`.

site/gatsby-site/src/components/discover/Controls.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+      setFirstRender(false);
```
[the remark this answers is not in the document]
lmcnulty: Done.

site/gatsby-site/src/components/discover/Controls.js
```diff
@@ -0,0 +1,67 @@
...
+  color: inherit;
+  background: none;
+  border: none;
+`;
+
+const Controls = ({ query }) => {
+  const [expandFilters, setExpandFilters] = useState(false);
+
+  useEffect(() => {
+    let filter = false;
```
cesarvarela, line 26: No need to change this but `some` can save you some lines in the future, and the `if` is redundant.
```
const filter = REFINEMENT_LISTS.some(r => query[r.attribute]);
setExpandFilters(filter);
```

site/gatsby-site/src/components/discover/Controls.js
```diff
@@ -0,0 +1,67 @@
...
+          <ExpandFilters
+            id="expand-filters"
+            data-cy="expand-filters"
+            onClick={() => setExpandFilters(!expandFilters)}
+          >
+            {expandFilters ? '⏷' : '⏵'} Filter Search
+          </ExpandFilters>
+        </Col>
+      </Row>
+      <Row className="mb-3 hiddenMobile">{expandFilters ? <Filters /> : <div></div>}</Row>
```
cesarvarela, line 62: No need to use an empty div here: `{expandFilters && <Filters />}`,

cesarvarela (approved):
Added two nitpicks, good to merge anyway.

smcgregor:
I whitelisted the domain so the deploy preview should work, and I am not seeing the arrow for the filters. Any ideas?

Screen Shot 2022-06-14 at 9 38 12 PM

lmcnulty:
The arrow was displayed with the unicode character, so it's probably a font issue. I've switched to using an svg icon – let me know if that fixes it.

PR #664, opened 2022-06-01
Add type="date" to date fields in SubmissionForm.js

lmcnulty:
Solves re-opened #627


cesarvarela:
These forms are different on purpose! Even though they are almost the same now, in the future the will drift apart.

For example the `incident_date` field is not in the `reports` form anymore (tests should have failed tho 🤔)

The idea is that `submissions`, `reports` and `incidents` have each their own form.

Issue #659, opened 2022-06-01
Typing into the image URL form element generates a large number of Cloudinary queries


smcgregor:
The Cloudinary query should likely not trigger on typing events.

![Screen Shot 2022-05-31 at 11 10 20 PM](https://user-images.githubusercontent.com/64780/171339134-e6d2d41a-afcb-4993-b973-634e0db7259d.png)

smcgregor:
Also, the image is not updating when the field is programmatically filled by by the `Fetch info` button.

smcgregor:
And after submitting the form, the state of all form elements updates to blank but the image preview remains the same.

PR #670, opened 2022-06-02
Stop extra requests on typing image urls, show errors

lmcnulty:
Solves #659

- The image preview is now updated only after the image_url has not changed for two seconds, preventing excessive network requests
  - ~~Since there's now delay, the test really does need a `wait()`.~~
    - Instead of wait(), now gives cy.get() a timeout of 30 seconds.
- The image preview now renders based on the `values` prop rather than `onChange`, which makes it update on programmatic changes.
- If the entered URL is not valid or an image, an error is shown.
  - Since Cloudinary shows a fallback image without an error, before passing the url to Cloudinary, I attempt to load it directly in an offscreen  element -- if it succeeds, then it gets passed to Cloudinary. Otherwise, it shows an error to the user. It seems like there should be a better way to determine whether the image shown is the fallback, but I couldn't find one.

### DIFF
--- a/site/gatsby-site/src/utils/cloudinary.js
+++ b/site/gatsby-site/src/utils/cloudinary.js
@@ -7,6 +7,8 @@ import { auto as qAuto } from '@cloudinary/base/qualifiers/quality';
 import styled from 'styled-components';
 import config from '../../config';
 import TextInputGroup from 'components/forms/TextInputGroup';
+import { Spinner } from 'react-bootstrap';
+import { isWebUri } from 'valid-url';
 
 const getCloudinaryPublicID = (url) => {
   // https://cloudinary.com/documentation/fetch_remote_images#auto_upload_remote_files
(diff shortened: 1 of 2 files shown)


site/gatsby-site/src/utils/cloudinary.js
```diff
@@ -58,51 +67,81 @@ const PreviewImageInputGroup = ({
...
   };
 
+  // When the form value changes, wait two seconds,
+  // and if it hasn't changed again by then, update the cloudinaryID.
+  // This prevents repeated requests for partially-typed URLs.
+  if (values.image_url != imageUrl.current) {
+    imageUrl.current = values.image_url;
+    setUpdatingImage(true);
+    clearTimeout(timeoutID.current);
+    timeoutID.current = setTimeout(updateCloudinaryID, 2000);
```
smcgregor, line 108: In the future, don't do this manually. Use lodash: https://lodash.com/docs/4.17.15#debounce

Issue #645, opened 2022-05-25
A11y skip-to-content link


lmcnulty:
It's an accessibility [best practice](https://www.w3.org/WAI/WCAG21/Techniques/general/G1) for the first focusable element on a page to be a hidden link that skips past the site navigation and to the main page content. This is important for users of screen-readers that announce elements in the order they appear in the DOM. It's also helpful for keyboard navigation. E.g.

https://user-images.githubusercontent.com/25443411/170301144-34b4fde7-8743-4329-bfc8-279d9d0cb80e.mp4

smcgregor:
Interesting, sounds good to me!

PR #671, opened 2022-06-02
Add a11y skip-to-content link

lmcnulty:
Solves #645

I had to add `id="content"` to the main element in `Layout.js` and `LayoutHideSidebar.js`, so any future layouts would need it as well. It would be nice to somehow enforce that, but I'm not sure of the best way. I tried writing a test that would iterate through every page and test that a `#content` element exists, but it was slow and failed on `/apps/classifications` even though it does have a `#content`, probably because it takes some time to load.

**Edit**: Moved link to navbar, still hidden by default but visible on focus – potentially looks more natural not pushing the rest of the content down.

https://user-images.githubusercontent.com/25443411/171862700-ad61fb5e-0921-438d-b46a-ed07c322c0ec.mp4

Works ok on narrow viewports:

https://user-images.githubusercontent.com/25443411/171862781-956a5920-0fa2-4078-94ed-84311e1d4927.mp4

### DIFF
--- a/site/gatsby-site/src/components/Layout.js
+++ b/site/gatsby-site/src/components/Layout.js
@@ -77,7 +77,7 @@ const Layout = ({ children, collapse, className, location }) => (
           dangerouslySetInnerHTML={{ __html: config.sidebar.title }}
         />
       )}
-      <Content className="mb-5">
+      <Content id="content" className="mb-5">
         <MaxWidth className={className}>{children}</MaxWidth>
       </Content>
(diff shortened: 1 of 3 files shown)


smcgregor:
I am guessing that the link needs to be "displayed" for this to work? The styling is a curious collection of choices, but I can see how this would work.

The build preview didn't trigger since it was originally targeting master. Can you push a new commit to this so the build system will pick it up?

Issue #654, opened 2022-05-31
New report submissions are not shown in queue


LKchemposer:
When I submitted a new report, it didn't show up in the [queue](https://incidentdatabase.ai/apps/submitted?lang=en).

It seems that the new reports are in the DB, but the app backlog for report review is filled up.

smcgregor:
Confirmed. I believe the problem is the graphql endpoint limiting the query to 100 reports. @lmcnulty can you take a look at this Tuesday morning? I think it should be as simple as increasing the `limit`

lmcnulty:
More than 100 reports will appear in the submission queue if in `site/gatsby-site/src/graphql/submissions.js` you add a limit greater than 100 to the `FIND_SUBMISSIONS` query:

``` diff
export const FIND_SUBMISSIONS = gql`
  query FindSubmissions {
-    submissions {
+    submissions(limit: 200) {
```
This looks kind of ugly to me, though. Firstly, I'm not sure why the limit is 100 by default; it seems like this should be a matter of configuring the defaults rather than changing the specific query. I grepped the repo for "100",  but I didn't find anything that seemed related – I'm not sure if there's a better approach there. Secondly, however the limit came to be, it seems like eventually there should be a better solution than just increasing the limit past however many submissions there are. For example, pagination.

From a UX perspective, it seems like the most likely use-cases for the submission queue are (1) editors who want to review reports so they can promote them to issues or add them to existing ones and (2) submitters who want to see that the submission they added is actually there and matches what they expect.

For (1), it seems like it would be most useful to show the oldest $n$ submissions that the editor can act on. For (2), it seems like it would most useful to show the newest $n$ submissions. Perhaps it would be helpful to show those two lists at the top, and then show the full submission queue only after some interaction?

I'm not sure how much time should be put into this right now. Maybe I should just make a pull request adding a larger limit and save anything else for later?

smcgregor:
This is a strange default, but I think we should solve this in two passes,

1) Set the limit explicitly to something like `999`  
2) Figure out a bit more about why this is happening and correct the default

This will unblock @LKchemposer and I think we can come back to (2) in a few weeks.

cesarvarela:
Comes from here: https://www.mongodb.com/docs/atlas/app-services/graphql/types-and-resolvers/#find-multiple-documents

LKchemposer:
[...]

PR #675, opened 2022-06-02
Add `limit: 200` to FIND_SUBMISSIONS query

lmcnulty:
Fixes the problem of missing entries from the submission queue named in #654. For now... Eventually we will probably want to add pagination and a better table view.


smcgregor:
FYI @cesarvarela , merging.

PR #685, opened 2022-06-06
Set required fields

### LINKED ISSUE #611: Set required fields in JSON validation schema
Setting required fields will help with data integrity and avoid build-breaking issues like [here](https://app.netlify.com/sites/staging-aiid/deploys/6282b5d0cdec9a13969e89cb) where an incident has a null date.

Docs:

https://www.mongodb.com/docs/realm/schemas/enforce-a-schema/#validate-null-types

cesarvarela:
closes #611 

Wondering if `title` should be required for `incidents` there are a few that don't have one.


cesarvarela:
The failing test won't pass until the migrations in the PR are run.

smcgregor:
@lmcnulty you and @cesarvarela are solving different problems here, but I think it makes sense for you to build on top of this change set. Can you review and comment if you are good with merging this?

site/gatsby-site/src/components/forms/IncidentReportForm.js
```diff
@@ -108,6 +108,16 @@ const IncidentReportForm = () => {
 
   const [parsingNews, setParsingNews] = useState(false);
 
+  useEffect(() => {
+    try {
+      const url = new URL(values?.url);
+
+      setFieldValue('source_domain', url.hostname);
+    } catch (e) {
```
lmcnulty, line 116: I might be mistaken about this, but I believe that `useEffect` runs after every re-render, and that `useFormikContext()` causes a re-render on each change to the form values: i.e. as the user types a value. Catching an exception is fairly expensive, so that could add input latency. I think it might be better to `import { isWebUri } from 'valid-url';` and check that the url is valid before trying to construct a `URL` object from it.

cesarvarela: From my tests I didn't notice input lag, I would not add another package just for this because it also has its tradeoffs (bigger bundle size), if you do experience lag let's change it.

lmcnulty: I tried testing for input latency, and the difference was small enough that I'm not confident that it exists. `valid-url` is already among our dependencies, but I don't know if it ends up in the bundle and I suppose it doesn't make a difference.

lmcnulty (approved):

site/gatsby-site/src/components/forms/IncidentReportForm.js
cesarvarela: Oh, that package should be explicitly required in `package.json` then. We shouldn't depend on peer deps.

PR #687, opened 2022-06-07
Prevent 'From Date' later than 'To Date' in discover filter

lmcnulty:
Fixes item in #630. In the discover date range filters, the min and max date input values are restricted so that the "from date" cannot be later than the "to date." Also adds missing space between the two fields.


lmcnulty:
The first time the build ran, the "Should display correct BibTex Citation" test failed.  I pressed "clear cache and re-run", since this PR couldn't possibly have caused that, and it succeeded the second time. I can't seem to see the original log now, but I believe the error had something to do with cy.get() not finding the BibTex modal. That's perhaps something to keep an eye one.

cesarvarela:
That test is really flaky, and I have no idea why; feel free to improve on it!

smcgregor:
Does this work if the refinement is set via URL parameters? I am worried that it may not update the ranges appropriately to make the full range selectable. If you don't think that is a problem here, I'll merge.

lmcnulty:
When the refinement was set by url parameters, it made it impossible to select a date. I've changed it so that in the unlikely event that the date range in url parameters is invalid, no restraint is placed on the dates that can be selected.

smcgregor:
Will merge when the build passes.

smcgregor:
Merging since the only failing test is the flaky one.

PR #705, opened 2022-06-14
Setup cypress dashboard

cesarvarela:
let's see if this throws more info on why the test is failing


smcgregor:
@cesarvarela should this still be open/draft?

cesarvarela:
I'm still questioning the usefulness of this. I will reopen in any case after #751

PR #708, opened 2022-06-15
Add semantically similar reports to submission page

lmcnulty:
- Doesn't process texts under 256 non-space characters
- Strips markdown from input
- Suggests reports for all similar incidents, not just the first one
- Prevents GraphQL queries for incidents matching dates before 1970
- Presents the option to set the incident ID to those of the reports that are by the same author or semantically similar
- Sorts semantically similar incidents reports by similarity
- Has basic test coverage

---

Possible issues:

- `stripMarkdown()` removes code blocks, which includes any indented text. It's easy to copy and paste text and have it intended without intending it to be a code block. For instance, when I past HTML lists into the text field, they appear indented and get treated as code. It may be better if we could find a way to treat code as regular text.

### DIFF
--- a/site/gatsby-site/src/components/RelatedIncidents.js
+++ b/site/gatsby-site/src/components/RelatedIncidents.js
@@ -146,6 +146,8 @@ const searchColumns = {
     header: () => <>Most Semantically Similar Incident Reports (Experimental)</>,
     query: relatedIncidentsQuery,
     getReports: async (result) =>
+      // We need a list of reports, not a list of incidents,
+      // so we reduce to get reports with associated incidents
       result.data.incidents.reduce(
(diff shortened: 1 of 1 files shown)


lmcnulty:
I've worked on handling cases with a cold start of the lambda function. The issues I identified were:

- The request would time out. To address this I:
  - Added a single retry on failure
  - Passed `{ timeout : 3000 }` to the axios request
    - I don't think this makes a difference, because I'm pretty sure the timeout happens on AWS's side. However, it can't hurt.

- The loading indicator appeared only while waiting for the GraphQL response, not the NLP API response. For this I had to special case `loading.byText`.

lmcnulty:
Closing in favor of #736.

PR #709, opened 2022-06-15
Feature i18n reports

cesarvarela:
Change-set

- Replaces the old Algolia index generation code path with a new one. Now indexes' name incorporate the language name  `instant_search-${language}` (old index `instant_search` index can be deleted)
- Adds Gatsby i18n packages `gatsby-theme-i18n` & `gatsby-theme-i18n-react-i18next`
- Creates a specific citation page per language `/:language/cite/:incident_id` (default language `en` isn't prefixed)

Required env variables (build will fail if any of these is missing)

- `MONGODB_TRANSLATIONS_CONNECTION_STRING`: MongoDB string that should have write access to `translations` and read access to `aiidprod` databases
- `GOOGLE_TRANSLATE_API_KEY`:  This needs to be set to something
- `GATSBY_AVAILABLE_LANGUAGES`: A comma-separated list of available languages: `en, es`
- `ALGOLIA_ADMIN_KEY`: Self-explanatory (already set)
- `GATSBY_ALGOLIA_APP_ID`: Ditto

To view the language switcher in the discover app, a local storage value needs to be set. Run the following on the browser console:  `localStorage.setItem('i18n', true)`.

After everything is ready:

-`TRANSLATE_DRY_RUN`: Needs to be set to `false.` to have the translation process hit the real API instead of the mock.


site/gatsby-site/src/utils/discoverIndexGenerator.js
```diff
@@ -1,167 +0,0 @@
...
-      }
-
-      allMongodbAiidprodClassifications(
-        filter: { namespace: { eq: "CSET" }, classifications: { Publish: { eq: true } } }
-      ) {
-        nodes {
-          incident_id
-          namespace
-          classifications {
-            Harm_Distribution_Basis
```
cesarvarela, line 49: @smcgregor some keys are missing and not being pushed to Algolia, like `Annotator`, `Reviewer`, `Quality Control`, etc. should all of them be pushed or should I keep a whitelist?

smcgregor: Those fields are internal process fields that should only be displayed to authenticated users. The purpose of defining the fields of the taxonomies in the DB is to support this sort of white/blacklisting.

smcgregor:
* Code looks good (one comment inline for the future)
* Also, why a whole separate database for `translations`?
* I need to play with the UI, but it won't build until I put in all the environment variables and make this run -- can you link me to a deploy preview on your fork?

cesarvarela:
> * Also, why a whole separate database for `translations`?

Not really; It does help to keep the dumps small, though. Or should these also be part of the backup?

My initial thinking was that these would only contain computed values based on other properties, but if we allow editors to make changes, it might make sense to move them to the main database.

(Eventually, these should also have their own schema, so they are queriable through graphQl)

> * I need to play with the UI, but it won't build until I put in all the environment variables and make this run -- can you link me to a deploy preview on your fork?

This one took some time because I found a couple of issues, but here it is:
https://staging--cesar-aiid-staging.netlify.app

smcgregor:
Thanks, is there an address/UI button I should click on the deploy preview? I keep hitting different languages directly in the query string and it is redirecting me back to English.

cesarvarela:
Oh yes, you have to run `localStorage.setItem('i18n', true)` in the console.

The language is now part of the path.

smcgregor:
Got it. I am seeing "Needs publish date" in the cards.

Screen Shot 2022-06-27 at 3 27 46 PM

smcgregor:
@lmcnulty if you have time to take a look at this first thing, please do so. Unless you find something that should not be merged, we will open issues around whatever you surface so we can get this into staging and address it there.

# Merge Time

@cesarvarela I think we should roll this out in these stages:

0. On our 1:1 meeting, let's talk about the separate `translations` database
1. Merge this to staging without turning on translation on staging
2. Turn on translation to Spanish in staging
3. Trigger the build
4. Validate the translations in staging
5. Deploy to production
6. Turn on Spanish and French in production
7. Validate French
8. Turn off the local storage check in staging (default to showing the language switcher
9. Deploy to production
10. Finish the other UI updates
11. Write the announcement blog post and post it in all machine-translatable languages
12. Evaluate where things stand and potentially expand to the ~20 highest performing languages in Google Translate

Issue #710, opened 2022-06-15
Deployment Blocked


smcgregor:
## Blocking Deploy ##

1. When the submit form has contents in the editor notes field, the submit button stops working. It looks to me like the field in the UI is not connected to the corresponding field in the database.

## New Issues (Not Blocking) ##

1. When entering contents into the non-image field, the image is repeatedly getting re-rendered/requested.

(there may be additional edits to this issue as I complete my functional testing on staging)

lmcnulty:
I think the error occurring when promoting a submission is due to missing fields in the [newReport](https://github.com/responsible-ai-collaborative/aiid/blob/staging/site/realm/functions/promoteSubmissionToReport.js#L33) object in promoteSubmissionToReport.js. I tried adding dummy values for the missing fields, and I'm able to promote submissions again.

smcgregor:
Can you post what you changed? I looked at the history of the file, and while it looks like it is missing fields, it looks like they have been missing for a while. And yet this has been working...

lmcnulty:
I changed it to this:
```javascript
  const newReport = {
    title: submission.title,
    report_number: lastReport.report_number + 1,
    ref_number: BSON.Int32(parentIncidents[0].reports.length), // this won't make sense with many to many relationships
    date_downloaded: '2000-01-01',
    date_modified:'2000-01-01',
    date_published:'2000-01-01',
    date_submitted:'2000-01-01',
    epoch_date_downloaded: 10000,
    epoch_date_modified:10000,
    epoch_date_published:10000,
    epoch_date_submitted:10000,
    image_url: 'http://www.google.com/favicon.ico',
    cloudinary_id: '',
    authors: ['nob'],
    submitters: ['nob'],
    text: 'beep',
    url: 'http://www.google.com/favicon.ico',
    source_domain: 'google.com'
  };
```

cesarvarela:
Oh I should have seen that one happening with the required fields change, that's exactly right @lmcnulty.

Feel free to create a pr with those changes, using empty values should work I think.

These issues will be prevented by having tests running on their own env eventually.

PR #713, opened 2022-06-15
Add required fields to newReport

lmcnulty:
This fixes the submission promotion issue in #710. The added fields have empty values, but they seem to get replaced with the correct ones later on.


cesarvarela:
Please replace `['']` with `[]` otherwise `''` will be considered an item.

cesarvarela:
👌

Issue #718, opened 2022-06-17
<main> overflows viewport horizontally


lmcnulty:
At certain screen size / zoom levels, the `<main>` element overflows the viewport horizontally, resulting in cut-off text and horizontal scrolling.

https://user-images.githubusercontent.com/25443411/174312196-4d865ed0-7428-4084-83d9-3fa2f2984f3c.mp4

PR #719, opened 2022-06-17
Prevent <main> from overflowing viewport horizontally

lmcnulty:
This solves #718 by setting an explicit width for the `<main>` element. Honestly, I'm not sure why it isn't already constrained to fit its parent flexbox. Setting `flex-shrink` doesn't help. There may be a more principled solution, but this works as a quick fix.

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

## Other sizes

https://user-images.githubusercontent.com/25443411/174323589-b2d75ee1-2f64-45fc-afe4-e9641fae8f1c.mp4

### DIFF
--- a/site/gatsby-site/src/components/Layout.js
+++ b/site/gatsby-site/src/components/Layout.js
@@ -38,6 +38,14 @@ const Content = styled('main')`
     background: ${({ theme }) => theme.colors.background};
   }
 
+  @media only screen and (min-width: 767px) {
+    /* Subtract the sidebar and margin width */
+    width: calc(100% - 298px - 66px);
+    > * {
+      max-width: 100%;
+    }
+  }
+
   @media only screen and (max-width: 1123px) {
     padding-left: 0;
     margin: 0 10px;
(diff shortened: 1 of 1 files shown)


smcgregor:
@lmcnulty you are proposes this as a quick fix with a follow up to make it better/more robust?

lmcnulty:
@smcgregor It would prefer to find a solution that doesn't require setting an explicit width, since it would have to be changed if we end up removing the sidebar, or adding another one, or changing its size. However, it might not be the most effective use of time – I guess I could use your opinion on whether it's worth going deeper into this.

smcgregor:
If/when we re-template, this problem would likely go away (and introduce some new ones, of course), so I am OK going with a quicker fix here. However, CSS is among my weaker web dev points and I am not sure what edge cases may exist here. @cesarvarela can you comment on this one?

cesarvarela:
Not a fan of using `calc` and such precise pixel values  but this seems to work well and as you say this will be replaced with the re-skin so :shipit:

PR #736, opened 2022-06-27
Add semantically similar reports to submission page

lmcnulty:
Supersedes #708

https://user-images.githubusercontent.com/25443411/176209746-9f4c1e36-0958-40ad-9b97-be75cc051abe.mp4

- Adds `SemanticallyRelatedIncidents` as a child component of `RelatedIncidents`, allowing it to maintain its own state. This simplifies the logic compared to the prior attempt.
- Allows the user to set the submission `incident_id` from the related incidents
- Shows reports from all related incidents, not just the first one
  - This results in longer lists, so now each list is constrained to 80% of the viewport height and has its own scrollbar
- Strips markdown from plain text before hitting API
- Runs NLP only on texts longer than 256 non-space characters
- Retries API requests twice in case of failure
- Catches exceptions from API and database request failures
- Displays appropriate error messages to the user
- Includes basic test coverage
- Following from [feedback on the prior PR](https://github.com/responsible-ai-collaborative/aiid/pull/708#discussion_r906228431), Minimum length test now checks for error message rather than non-existence of element:
  ```javascript
  cy.get('[data-cy=related-byText]').contains('Reports must have at least').should('exist');
  ```

### DIFF
--- a/site/gatsby-site/src/components/RelatedIncidents.js
+++ b/site/gatsby-site/src/components/RelatedIncidents.js
@@ -75,8 +75,11 @@ const searchColumns = {
     ),
     query: relatedReportsQuery,
     getReports: async (result, client) => reportsWithIncidentIds(result.data.reports, client),
-    isSet: (incident) =>
-      incident.date_published && isValid(parse(incident.date_published, 'yyyy-MM-dd', new Date())),
+    isSet: (incident) => {
+      const parsedDate = parse(incident.date_published, 'yyyy-MM-dd', new Date());
+
+      return isValid(parsedDate) && getUnixTime(parsedDate) > 0;
+    },
     getQueryVariables: (incident) => {
       const datePublished = parse(incident.date_published, 'yyyy-MM-dd', new Date());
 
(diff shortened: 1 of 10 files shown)


smcgregor:
@cesarvarela I requested your review on this. I did a more thorough review on an earlier version so this reflects many of the changes I already suggested. The functionality looks good and I took a brief tour of this updated code and functionality and didn't see major issues. I plan on stress testing it once you are 👍

site/gatsby-site/src/components/RelatedIncidents.js
```diff
@@ -33,6 +41,31 @@ const relatedReportsQuery = gql`
   }
 `;
 
+const reportsWithIncidentIds = async (reports, client) => {
```
cesarvarela, line 44: In the future, we should set up a schema relationship to avoid having to do this: https://github.com/responsible-ai-collaborative/aiid/blob/master/site/realm/data_sources/mongodb-atlas/aiidprod/incidents/relationships.json

cesarvarela (approved):
Added a couple of comments

smcgregor:
First try:
`{"errorMessage":"2022-06-29T23:18:21.243Z c89b6c18-936a-4dd8-89c8-6f2838f0d7c8 Task timed out after 10.01 seconds"}`

But the second try (after adding a space to the same text) worked.

I am merging this now, but we should work to improve the cold start behavior in the future.

PR #722, opened 2022-06-21
Add nlp-generated similar incidents incidents to citation page

lmcnulty:
- Tests pass in local environment after schema change + migration

PR #737, opened 2022-06-27
Add nlp-generated similar incidents incidents to citation page

lmcnulty:
Supersedes #722 

- `nlp_similar_incidents` is added to the schema for incidents and submissions
- A migration is included to initialize it as an empty array for all existing items.
- Semantically similar incidents are stored on submission.
- When `SemanticallyRelatedIncidents` are displated in the submission queue, the initial values are retrieved from the database instead of from the NLP api.
- Similar incidents are displayed at the bottom of the citation page.
- Tests fail in CI because of schema changes but pass in dev environment after migrations.

Following from [feedback on the previous PR]()
- `nlp_similar_incidents` is not added to the reports collection.
- The similar incident cards are their own component.

### DIFF
--- a/site/gatsby-site/src/components/cite/SimilarIncidents.js
+++ b/site/gatsby-site/src/components/cite/SimilarIncidents.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 import styled from 'styled-components';
 import { Card, Button } from 'react-bootstrap';
 import { formatISO, format, parse } from 'date-fns';
@@ -7,6 +7,8 @@ import { faFlag, faQuestionCircle } from '@fortawesome/free-solid-svg-icons';
 import { Image } from '../../utils/cloudinary';
 import { fill } from '@cloudinary/base/actions/resize';
 import md5 from 'md5';
+
+const blogPostUrl = null;
 
 const Subtitle = styled.p`
   font-size: 120%;
@@ -72,13 +74,18 @@ const FlexSeparator = styled.div`
 const FlagButton = styled(Button)`
   padding: 0px !important;
   color: var(--bs-gray-600) !important;
-  :hover {
+  &:hover {
+    color: var(--bs-gray-500) !important;
+  }
+  &.flagged {
     color: var(--bs-red) !important;
   }
 `;
 
 const SimilarIncidentCard = ({ incident }) => {
   const parsedDate = parse(incident.date, 'yyyy-MM-dd', new Date());
+
+  const [flagged, setFlagged] = useState(false);
 
   return (
     <Card>
@@ -101,8 +108,12 @@ const SimilarIncidentCard = ({ incident }) => {
             </span>
           </div>
           <FlexSeparator />
-          <FlagButton variant="link">
-            <FontAwesomeIcon icon={faFlag} className="fa-flag" />
+          <FlagButton
+            variant="link"
+            className={flagged ? ' flagged' : ''}
+            onClick={() => setFlagged(!flagged)}
+          >
+            <FontAwesomeIcon icon={faFlag} />
           </FlagButton>
         </CardFooter>
       </Card.Body>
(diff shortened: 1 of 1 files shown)


smcgregor:
@lmcnulty Can you work out a way to show deploy previews on PRs like these? Options:

* Post a link to your already existing deploy preview
* Create your own project here that points to your development DB. https://app.netlify.com/teams/responsible-ai-collaborative/overview
* Create your own Netlify account and do the same as above

I'll take a look at the code now...

site/gatsby-site/migrations/2022.06.23T20.09.47.initialize-empty-nlp-similar-incidents.js
```diff
@@ -0,0 +1,39 @@
+const config = require('../config');
+
+/** @type {import('umzug').MigrationFn<any>} */
+exports.up = async ({ context: { client } }) => {
+  const submissionsCollection = client
+    .db(config.realm.production_db.db_name)
+    .collection('submissions');
+
+  await submissionsCollection.updateMany(
```
smcgregor, line 9: Sometimes `updateMany` gives me problems. Have you experienced any issues? Otherwise, this looks good.

site/gatsby-site/src/components/SemanticallyRelatedIncidents.js
```diff
@@ -0,0 +1,164 @@
...
+        report_number
+        title
+        url
+      }
+    }
+  }
+`;
+
+const semanticallyRelated = async (text, max_tries) => {
+  const url = `/api/semanticallyRelated?text=${text}`;
```
smcgregor, line 22: I think the retry logic should likely be in the API and hidden from the UI.

site/gatsby-site/src/components/cite/SimilarIncidents.js
```diff
@@ -0,0 +1,194 @@
...
+import styled from 'styled-components';
+import { Card, Button } from 'react-bootstrap';
+import { formatISO, format, parse } from 'date-fns';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faFlag, faQuestionCircle } from '@fortawesome/free-solid-svg-icons';
+import { Image } from '../../utils/cloudinary';
+import { fill } from '@cloudinary/base/actions/resize';
+import md5 from 'md5';
+
+const similarIncidentsQuery = gql`
```
smcgregor, line 12: I believe this will dynamically hit the database every time a user visits a citation page, but everything else on the citation page is rendered at build time. I think this DB logic should likely be moved to the pageCreators: https://github.com/responsible-ai-collaborative/aiid/blob/master/site/gatsby-site/page-creators/createCitiationPages.js

smcgregor:
I left a few comments for discussion or updating. Once we reach convergence there, we will kick this over to @cesarvarela as the final reviewer.

site/gatsby-site/src/components/SemanticallyRelatedIncidents.js
lmcnulty: Got it. I made the change in #736 and merged it here.

site/gatsby-site/migrations/2022.06.23T20.09.47.initialize-empty-nlp-similar-incidents.js
lmcnulty: I haven't had any problems with it here, but if anyone does, I'll switch to using batched `updateOne`s as used in some of the other migrations.

lmcnulty:
> Can you work out a way to show deploy previews on PRs like these?

I've created my own Netlify project for this that creates deploy previews for the [pull requests](https://github.com/lmcnulty/aiid/pull/2) to my staging branch.

smcgregor:
* The cards are quite large. I think it may be better to make them take up half the column rather than the whole column (when viewed on larger screens).
* The flagging feature is currently non-operative?
* The "more info" element should like to the blog post, even if it does not exist yet
* I believe many of the editor elements for assigning/removing similarity will come in with other PRs, so I will hold commentary there
* It looks like there are a few features that are going to become bound together here and they should all land together, should we complete review here and then roll it into the next PR for review?

PR #727, opened 2022-06-24
Nlp similar incidents migration

lmcnulty:

PR #738, opened 2022-06-27
NLP similar incident values migration

lmcnulty:
Supersedes #727. Depends on previous PRs, but the parts original to this one are in [the migration file](https://github.com/responsible-ai-collaborative/aiid/pull/738/files#diff-876234383d101d7fca8d135b16694ecbabc500c8153a9132f94a4e407c3929f4).

Following from feedback:
- Locks to state.csv download to revision rather than branch
- Fixes incorrect usage of "similarity" as name for coordinates
- Uses arbitrary set of hardcoded IDs for test runs rather than a random selection at runtime
- Uses POST instead of get, removes truncation of coordinates
  - Oddly, this seems to work correctly when supplying an embedding, but not when supplying text.


smcgregor:
Looks good to me, pending the merge on the upstream changes.

smcgregor:
* We should get one final update to the state.csv file before merging this.
* We will merge once @cesarvarela is 👍

PR #752, opened 2022-06-29
Add language selector to submit form

cesarvarela:


site/gatsby-site/src/components/ui/PopOvers.js
```diff
@@ -89,3 +89,5 @@ export const tags = Popover(
   'Tag the type of incident report',
   `The purpose of report tags is to support sorting reports on the citation page for the incident. When in doubt, you can leave this field blank and a database editor will select appropriate tags at ingestion time.`
 );
+
+export const language = Popover('Language of the incident report', `????`);
```
cesarvarela, line 93: @smcgregor help here, please.

smcgregor: `This is the written language of the text found in the report, not the language of the people or countries involved in the report. The reports will be translated to supported languages when we ingest the text.`

cesarvarela:
Checked again and this should be ok to merge!

smcgregor:
@lmcnulty first up on review, then I'll do final and merge.

site/gatsby-site/src/components/i18n/languages.json
```diff
@@ -0,0 +1,446 @@
...
+    "code": "zu",
+    "name": "Zulu"
+  },
+  {
+    "code": "he",
+    "name": "Hebrew"
+  },
+  {
+    "code": "zh",
+    "name": "Chinese (Simplified)"
```
lmcnulty, line 444: This results in "Chinese (Simplified)" being rendered in the list twice, which looks like a mistake. It seems like it would make more sense to set the name to just "Chinese" and position it directly after "Chinese (Simplified)" and "Chinese (Traditional)." Likewise, it seems like we should just use "he" for Hebrew, as "iw" seems to be [outdated](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes#he).

Incidentally, I look forward to the day when we get submissions in Latin and Esperanto.

site/gatsby-site/src/components/submissions/SubmissionForm.js
```diff
@@ -273,7 +274,24 @@ const SubmissionForm = () => {
         </Form.Group>
 
         <Form.Group className="mt-3">
-          <Label popover={POP_OVERS['tags']} label={'Tags'} />
+          <Label popover={POP_OVERS['language']} label={'Language'} />
+          <Form.Select
+            name="language"
+            placeholder="Report Language"
+            className="mt-3"
```
lmcnulty, line 281: Using `mt-3` here makes the space between the label and the input larger than it is for the other form entries.

image

lmcnulty (changes requested):

cesarvarela:
Updated

lmcnulty (approved):

PR #765, opened 2022-07-04
Translate citation pages

cesarvarela:


site/gatsby-site/i18n/locales/en/actions.json
```diff
@@ -0,0 +1,3 @@
...
...ntents of the report are not parsed properly</li><li>The authors of the report are not associated with the report</li><li>The report is associated with the wrong incident</li><li>The text contents of the report are not parsed properly</li></ul><p>Flagged content will still be displayed within the database, but a database editor willperiodically review the incident reports that have been flagged. Please note that thecontent contained within incident reports (e.g., the commentary within a news article) doesnot need to be correct or consistent across articles. If an article is wrong, misleading, orfraudulent, the best response is to submit additional incident reports that correct therecord. The incident database is meant to capture the complete state of knowledge anddiscourse for incidents, not as an arbiter of what happened in individual incidents. Infuture versions of the database it will additionally be possible to apply tags to incidentreports to classNameify their content.</p><p>Please do NOT flag content if,</p><ul><li>You disagree with the report</li><li>The linked report has disappeared from the web</li><li>The report should not be considered an &quot;incident&quot;</li></ul>"
```
lmcnulty, line 2: Missing spaces: “willperiodically”, "thecontent", “orfraudulent”, ...

It seems like storing blocks of HTML this long in JSON makes them harder to read and edit. I'm guessing this is required by the translation plugin? If it's just going to be a couple of keys like this, that's probably fine, but if there were many, it might be worth it to write a preprocessing script to construct the JSON from html or markdown files.

cesarvarela: Thanks.

Yeah, it didn't feel very good doing it like this, but it isn't worth doing anything else more complicated **yet**

lmcnulty (approved):

PR #769, opened 2022-07-06
Landing Page i18n support

### LINKED ISSUE #766: Multi lingual support - Landing page

cesarvarela:
closes #766


site/gatsby-site/src/components/landing/Featured.js
```diff
@@ -22,45 +23,49 @@ export default function Featured() {
   return (
     <Card className="text-center">
       <Card.Body>
-        <Card.Title as="h2">The Database in Print</Card.Title>
+        <Card.Title as="h2">
+          <Trans ns="landing">The Database in Print</Trans>
```
lmcnulty, line 27: This is not really related to this PR, but I have been wondering whether "in Print" is really the appropriate term to use here. It seems like we're linking to online coverage, so something like "in the Press" would be more accurate.

site/gatsby-site/src/components/forms/QuickAddForm.js
```diff
@@ -55,7 +58,7 @@ const QuickAddForm = ({ className = '' }) => {
           message: (
             <>
               {'Report successfully added to review queue. It will appear on the  '}
```
lmcnulty, line 60: Is there a reason why the toast messages aren't wrapped in `t()` ?

lmcnulty (approved):
It looks like the changes here are a subset of the ones in #780, which I did first because it was on top in my inbox. I've copied the two comment I made there.

PR #777, opened 2022-07-08
Discover incident variety

### LINKED ISSUE #748: Discover page - option to display results aggregated by incident id
- User story: "Sometimes, I don't want to see many reports pertaining to one incident, but rather, I want to see what unique incidents are associated with my search query."

Sample results displayed for the search query "YouTube," where 9/12 results pertain to one incident:

image

lmcnulty:
Adds a switch to return only the closest-matching report for each incident in the search results. Resolves  #748, not including the suggestion of curating a set of results to display in the absence of a search query, which can be in its own PR.

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

On mobile, the filters modal is repurposed as a general "Search Options" modal containing the switch and the previously unavailable-on-mobile DisplayModeSwitcher.

I used the translation functions where appropriate, but since I don't know Spanish, the new values in the translation file are currently left untranslated:

```json
  "Options": "Options",
  "Search Options": "Search Options",
  "1st report only": "1st report only"
```


site/gatsby-site/i18n/locales/es/translation.json
```diff
@@ -19,5 +19,8 @@
   "Show Details on Incident #{{id}}": "Ver detalles del Incidente #{{id}}",
   "Filter by Incident ID #{{id}}": "Filtrar por ID de Incidente #{{id}}",
   "Flag Report": "Reportar",
-  "Filters": "Filtros"
+  "Filters": "Filtros",
+  "Options": "Options",
```
cesarvarela, line 23:
```
"Options": "Opciones",
  "Search Options": "Opciones de Búsqueda",
  "1st report only": "Sólo el primer reporte"
```

site/gatsby-site/src/components/discover/Controls.js
```diff
@@ -26,7 +23,25 @@ const ExpandArrow = styled(FontAwesomeIcon)`
   vertical-align: -0.2em !important;
 `;
 
-const Controls = ({ query }) => {
+const Hbox = styled(Col)`
```
cesarvarela, line 26: Once we move to Tailwind we should avoid using Styled Components.

site/gatsby-site/src/components/discover/Controls.js
```diff
@@ -40,7 +55,19 @@ const Controls = ({ query }) => {
...
+          <Form.Check
+            type="switch"
+            id="hide-duplicates"
+            checked={hideDuplicates}
+            onClick={(event) => {
+              setHideDuplicates(event.target.checked);
+            }}
+          />
+          <Form.Label for="hide-duplicates">
+            {t('1st report only').split(' ').join('\u00a0')}
```
cesarvarela, line 68: What's the split for? Not sure it will place nice with translations 🤔

lmcnulty: `\u00a0` is a non-breaking space. The label text sometimes gets split into two lines when it's just regular spaces, which looks bad.

I've been assuming that the t() function just returns the translation as a string, in which case it seems like it would work for any language. Even for those that don't use spaces between words, it will just return the same string. Are there any edge-cases I'm missing?

cesarvarela: Sorry forgot  about this pr, but you can achieve something similar with one of the CSS rules `break-word` or `word-wrap` or something similar (I always have to google the exact names 😆), and it will be less obscure.

PR #792, opened 2022-07-12
Make images load fallback on failed request, add extra escape

### LINKED ISSUE #723: Images 404ing
Two images on incident 66 are 404ing.

Screen Shot 2022-06-21 at 10 13 45 PM

Screen Shot 2022-06-21 at 10 12 48 PM

lmcnulty:
Resolves #723

Cloudinary currently loads a fallback if the `publicId` of an `Image` doesn't have a referent, but it gives a 400 error if the `publicId` is invalid (e.g. if it contains characters that aren't permitted). For some of our images that were 400ing, the response header said that the ID was invalid. The `CloudinaryImage` constructor is supposed to escape urls, but it behaves weirdly, because some of the images only load when you url-encode the `%` character. But there are other causes – many of our 400ing images no longer load from the original source. So this PR makes the Image component check if the image loaded, and if not, it loads the fallback instead.


clari182 (approved):

site/gatsby-site/src/utils/cloudinary.js
```diff
@@ -34,7 +40,32 @@ const Image = ({ publicID, className, alt, transformation = null, plugins = [laz
...
-  return <AdvancedImage alt={alt} className={className} cldImg={image} plugins={plugins} />;
+  useEffect(() => {
+    const useFallbackIfLoadFailed = () => {
+      const img = imageElement.current?.imageRef.current;
+
+      if (!(img?.naturalHeight > 0)) {
+        if (img?.complete) {
+          setCloudinaryID('fallback.jpg');
+        } else {
+          setTimeout(useFallbackIfLoadFailed, 1000);
```
cesarvarela, line 51: I think this needs a cleanup function:
 - component mounts
 - time out is started
 - user navigates away from the page
 - component is unmounted
 - timeout will still execute 
 - memory leak

cesarvarela (changes requested):

site/gatsby-site/src/utils/cloudinary.js
lmcnulty: Added

cesarvarela:
image
 There is something going on here

cesarvarela:
When editing an image I'm also getting the fallback image, on current staging branch:

image

The image is there though:

image

clari182 (approved):

cesarvarela:
I want to merge this but the deploy preview still has this issue:

image

lmcnulty:
Ah, I still haven't figured this one out yet. I will get back to it soon.

lmcnulty:
@cesarvarela The problem is fixed in [the current deploy preview](https://deploy-preview-792--staging-aiid.netlify.app/apps/discover?display=details&lang=en&page=1). It turns out that `img.complete` is helpfully set to `true` when `src` is undefined. When I was running it locally it would work correctly when hot-loading code but not on an initial page load, which I guess is why I missed it originally.

cesarvarela (approved):

PR #807, opened 2022-07-14
Translate submit form

### LINKED ISSUE #806: Multi lingual support to Submission form

cesarvarela:
closes #806


site/gatsby-site/i18n/locales/es/validation.json
```diff
@@ -0,0 +1,10 @@
+{
+  "*URL required": "*URL requerida",
+  "*Title is required": "*Título requerido",
+  "*Author is required. Anonymous or the publication can be entered.": "*Autor requerido. Anónimo o la publicación puede ser introducido",
+  "*Date published is required": "*Fecha de publicación requerida",
+  "*Date downloaded required": "*Fecha de descarga requerida",
+  "*Must enter URL in http://www.example.com/images/preview.png format": "*Debe introducir la URL en el formato http://www.ejemplo.com/imágenes/preview.png",
```
pdcp1, line 7: @cesarvarela What this `*` means in this context? Are they part of the markdown of the page? If they are, should we remove the markdown from the translations to keep the page styles in another place?

site/gatsby-site/i18n/locales/es/submit.json
```diff
@@ -0,0 +1,28 @@
+{
+  "New Incident Report": "Nuevo Reporte de Incidente",
+  "submitFormDescription": "El siguiente formulario creará un nuevo informe de incidentes para <2>revisión</2> e inclusión en la base de datos de incidentes de AI. Verifique cuidadosamente sus entradas para detectar problemas de contenido (por ejemplo, copiar y pegar accidentalmente anuncios). Para obtener detalles sobre el proceso de ingestión de la base de datos, consulte las <5>páginas de investigación</5> o <8>contáctenos si tiene preguntas.</8>",
```
pdcp1, line 3: @cesarvarela I think we should document how the `<Trans>` component work on the README file https://github.com/cesarvarela/aiid/tree/feature-i18n-submit#ui-translations
Maybe a link to the component documentation is enough, especially for these `<2>`, `<5>` tags and how they are processed in the translation logic.

pdcp1: I can infer that each number tag like `<2>` matches with a child HTML element of the element to translate, but it's better to have it explained.

pdcp1 (changes requested):

site/gatsby-site/i18n/locales/es/validation.json
cesarvarela: No special meaning it is just the * of "hey this is required"

site/gatsby-site/i18n/locales/es/submit.json
cesarvarela: Agree let's track that in another issue https://github.com/responsible-ai-collaborative/aiid/issues/831

PR #820, opened 2022-07-18
Make `WebArchiveLink`s split buttons with original and archive urls

lmcnulty:
Implements [suggestion](https://github.com/responsible-ai-collaborative/aiid/issues/61#issuecomment-1185726082) in #61, adding dropdown to choose between original or archive.org url.


cesarvarela:
@lmcnulty can you add a test for this please?

PR #833, opened 2022-07-20
Move language switcher to header

cesarvarela:
To make the language switcher appear you have to run `localStorage.setItem('i18n', '1')`


pdcp1 (approved):
It looks good 👍

PR #848, opened 2022-07-25
Render jsx in popovers, click to toggle visibility

### LINKED ISSUE #843: Help Instructions Not Rendered
The submission page's submit instructions are not encoding the HTML properly. See image below.

Screen Shot 2022-07-24 at 8 00 48 PM

lmcnulty:
Resolves #843.

There was previously no way to actually use the link when rendered, because the popover would disappear as soon as you moved your mouse toward it. This PR makes it so that clicking the (?) icon will cause the popover to remain visible until it's clicked again, allowing interaction with elements inside the popover.

### DIFF
--- a/site/gatsby-site/src/components/forms/Label.js
+++ b/site/gatsby-site/src/components/forms/Label.js
@@ -1,21 +1,29 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { OverlayTrigger, Form } from 'react-bootstrap';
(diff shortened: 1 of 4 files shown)


cesarvarela:
I find the click on the question mark thing obscure. I had to read the PR description to realize how it works. I would add a hover animation (change color?) and a toggle state (keep the hover state color?), but it can be done in another PR.

Issue #847, opened 2022-07-25
Deployment Issues


smcgregor:
This is an issue associated with the current staging version immediately prior to deployment. I'll highlight issues according to whether they are blocking (i.e., "don't deploy until these are fixed"), non-blocking (i.e., "fix these after deployment"), and Blocking Feature Announcement (i.e., things that need to be done before we can make the translation feature show).

# Blocking Deployment

* Khoa ingested two Spanish reports (one on incident 188 and another on 186). For 188, "dry run" was not set to false, so it did the filler translation. I then tried to fix it in the UI. When you edit a Spanish-language report on the English citation page, you edit the source Spanish report and not the English translation. This means I cannot correct a bad translation if I want to.
* We have nothing on the card indicating it was translated from Spanish. We need a banner across the cards indicating it was machine translated, likely with a link to switch to the Spanish version of the site.

# Non-blocking

* The width of the page on initial render jumps between widths. I believe it is rendering a default, then updating to whatever the screen size is. This is particularly clear when it comes to the search box text, which changes between "Search over 1600 reports of AI harms" and "Search 1600+ reports".
* Language selection could have flags in addition to the text for selecting the language 🏴󠁧󠁢󠁥󠁮󠁧󠁿🇪🇸. This may be super annoying to implement given how we turn on/off languages, in which case please feel free to ignore this one.
* Is `Descubrir` and `Enviar` the right translations of the discover and submit applications? @cesarvarela knows better than I do, but I would have thought the imperative form would be good here (e.g., `descubre`)
* @lmcnulty I get "Could not compute semantic similarity" very often, which I think is a quality of service problem pre-dating this deployment. The API either needs to be made more reliable from the user perspective, or the user-messaging needs to be more friendly.

# Blocking Feature Announcement

* Markdown links often kick the user back to English. For example, https://staging-aiid.netlify.app/es/taxonomies
* This should be translated to Spanish: https://staging-aiid.netlify.app/es/blog/raic-holds-first-all-hands-meeting
* I need to write the blog post talking about the feature for the announcement
[...]

PR #852, opened 2022-07-25
Switch quicksearch placeholder text determination from js to css

lmcnulty:
This fixes the issue from #847 where

>The width of the page on initial render jumps between widths. I believe it is rendering a default, then updating to whatever the screen size is. This is particularly clear when it comes to the search box text, which changes between "Search over 1600 reports of AI harms" and "Search 1600+ reports".

The text is now set with css, meaning that the browser can render the initial html appropriately for the screen size without having to wait for React to re-render the component. Unfortunately, css doesn't provide a way to set placeholder text, so it we have to simulate it with an `:after` styled to look like a placeholder.

### DIFF
--- a/site/gatsby-site/src/components/forms/SearchInput.js
+++ b/site/gatsby-site/src/components/forms/SearchInput.js
@@ -26,11 +26,6 @@ const SearchFormControl = styled(FormControl)`
   border-bottom-right-radius: 0.3rem !important;
 `;
 
-const SearchInputGroup = styled(InputGroup)`
-  max-width: 640px;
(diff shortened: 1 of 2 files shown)


cesarvarela:
Not very a11y friendly but no easy alternative either.

lmcnulty:
@cesarvarela For a11y, maybe we could use an [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) that would just provide the full "Search over 1600 reports of AI harms" to screen readers?

PR #860, opened 2022-07-26
Deploy to Production

cesarvarela:
Deployed here: https://deploy-preview-860--staging-aiid.netlify.app

Netlify build status is not updating here, but the build passed: https://app.netlify.com/sites/staging-aiid/deploys/62e19fbf38c692285af6926e

edit:

`GATSBY_EXCLUDE_DATASTORE_FROM_BUNDLE=1` needs to be set on Netlify's env


smcgregor:
## Non-Blocking Issues

* When you hover the "1st report only" text, the cursor doesn't indicate it is clickable, but if you do click it, it will toggle. The cursor should be a pointer.
* The initial results when "1st report only" is toggled on are much better than when it is toggled off. I think we want it to be off by default, but the results when nothing has been searched should be that higher-diversity result set.
* The text below does not correspond to the form,
Screen Shot 2022-07-27 at 2 43 38 PM
* Issue with the encoding of the link,
Screen Shot 2022-07-27 at 2 44 57 PM

(still working through it here)

smcgregor:
@lmcnulty I think the flagging of unrelated incidents needs some work as well. It isn't clear what users need to do to flag something, and when it is flagged, the editors won't know what the flagging corresponds to.

lmcnulty:
The issue with the unrendered HTML in the popover text is handled by #848.

lmcnulty:
> When you hover the "1st report only" text, the cursor doesn't indicate it is clickable, but if you do click it, it will toggle. The cursor should be a pointer.

#875

lmcnulty:
> The initial results when "1st report only" is toggled on are much better than when it is toggled off. I think we want it to be off by default, but the results when nothing has been searched should be that higher-diversity result set.

#877

smcgregor:
None of these are blocking, but they need new issues to track them or they need changes.

* Document `GATSBY_EXCLUDE_DATASTORE_FROM_BUNDLE` in the README
* Change "1st report only" in the discover app to be more informative to users. Some options: "Distinct incidents only", "Show single report per incident", etc. When this is toggled, it also effectively switches to "incident mode" from "report mode" and the results counter still says "259 reports found". It would be better for this to read "259 matching incidents".
* When you have selected a page via the pagination UI on the discover app, then subsequently apply a filter, the app will give you no results because you are still looking at a later page. Pagination should be cleared when you change the filters or search terms.

PR #783, opened 2022-07-11
Add z-index to main and right sidebar

lmcnulty:
Resolves #750

<table>
  <tr><th>Before</th><th>After</th>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

PR #869, opened 2022-07-27
Raise z-index of header to prevent covered language switcher menu

lmcnulty:
#783 increased the z-index of the main element, so it would cover the menu of the language switcher. This increases the navigation z-index to prevent it.


cesarvarela:
I did it first woohooo https://github.com/responsible-ai-collaborative/aiid/pull/863/commits/a8a00d736cad11f07956140494fe3face1586820

PR #877, opened 2022-07-28
Discover featured items

lmcnulty:
Adds field in `config` where we can manually specify reports to show when there's no search in the discover app. Apart from those, on the initial view it sorts alphabetically by the text body rather than the title, since those are more varied.

image


cesarvarela:
Instead of creating new indexes, why not:

- add a featured flag to index entries and set it to true if the entry is on the featured list
- if `searchState.query == '' && Object.keys(searchState.refinementList).length == 0` enable a `featured=true` virtual filter

(may not be possible 🤔 )

cesarvarela:
I like that; it is similar to sorting by "recommended" that's done on other sites. We probably want to always sort by that. 

Then the different indexes are not needed, right?

lmcnulty:
I tried always sorting by featured, and it caused featured items to appear as the top result for searches even when they weren't a good match for the query. Algolia's Premium Plan has a [relevant sorting](https://www.algolia.com/doc/guides/managing-results/refine-results/sorting/in-depth/relevant-sort/) mechanism to handle this, but I don't have access to that feature in my dev environment. It would require a virtual replica index.

site/gatsby-site/src/utils/AlgoliaUpdater.js
```diff
@@ -204,16 +207,30 @@ class AlgoliaUpdater {
...
-    });
+    index
+      .setSettings({
+        ...algoliaSettings,
+        attributeForDistinct: 'incident_id',
+        indexLanguages: [language],
+        queryLanguages: [language],
+        replicas: [featuredReplicaIndexName],
+      })
+      .catch((e) => console.error(e))
```
cesarvarela, line 224: please remove this one.

site/gatsby-site/src/utils/AlgoliaUpdater.js
```diff
@@ -204,16 +207,30 @@ class AlgoliaUpdater {
...
+      .then(() => {
+        const featuredReplicaIndex = this.algoliaClient.initIndex(featuredReplicaIndexName);
+
+        featuredReplicaIndex
+          .setSettings({
+            ranking: ['desc(featured)', 'desc(text)'],
+          })
+          .catch((e) => console.error(e));
```
cesarvarela, line 232: And this one. 

It is preferable that the build crashes and nothing gets deployed than deploying the app with some indexes working and some broken.

cesarvarela (changes requested):

site/gatsby-site/src/utils/AlgoliaUpdater.js
lmcnulty: Ok.

cesarvarela (approved):

site/gatsby-site/src/utils/AlgoliaUpdater.js
```diff
@@ -204,16 +207,27 @@ class AlgoliaUpdater {
...
 
     await index.saveObjects(entries);
 
-    await index.setSettings({
-      ...algoliaSettings,
-      attributeForDistinct: 'incident_id',
-      indexLanguages: [language],
-      queryLanguages: [language],
-    });
+    index
```
cesarvarela, line 216: There is a missing `await` here, execution will continue and not wait for this line to finish.

To avoid this type of error it is better to not mix `async/await` syntax with `Promises` syntax.

cesarvarela (changes requested):

site/gatsby-site/src/utils/AlgoliaUpdater.js
```diff
@@ -204,16 +207,27 @@ class AlgoliaUpdater {
...
+        ...algoliaSettings,
+        attributeForDistinct: 'incident_id',
+        indexLanguages: [language],
+        queryLanguages: [language],
+        replicas: [featuredReplicaIndexName],
+      })
+      .then(() => {
+        const featuredReplicaIndex = this.algoliaClient.initIndex(featuredReplicaIndexName);
+
+        featuredReplicaIndex.setSettings({
```
cesarvarela, line 227: This line too is missing an `await `or  returning the promise

cesarvarela:
@lmcnulty don't forget about this one, this is the last PR before deploying to prod 🚀

site/gatsby-site/src/utils/AlgoliaUpdater.js
lmcnulty: Added. Does it look good now?

cesarvarela: Almost! Line 227 is missing an await  or a return

cesarvarela: woot not sure if you didn't push or  github is showing me an old diff

lmcnulty: It's showing me an old diff unless I go directly [the commit](https://github.com/responsible-ai-collaborative/aiid/pull/877/commits/caa08c69bf3439cfcb74d3de81dcecb68e4db000).

cesarvarela: that was it thanks ⚡

cesarvarela:
The new indexes won't be created until we merge to staging, but do these tests need an update?

1) Should load at least 30 hits
2) Filters by incident Id using card button
3) Opens an archive link

lmcnulty:
I found that only "filters by incident Id using card button" was failing locally since my Algolia index had changed so that the expected report wasn't showing up. I added the ability to set the "featured" value numerically in the config, and made report 23 show up first so the test will pass.

The others two I think are failing because the new index doesn't exist yet. When running with an up-to-date index, [the discover tests all pass](https://app.netlify.com/sites/lmcnulty-staging/deploys/62eab44a12eacd00090d7756).

For some reason the submit test was failing because the top result from the NLP api was not what it was before. I'm not sure why that could be, but I've set it to use the specific one. That probably belongs in another PR, though, so I'm going to cherry-pick the commit and add it in another PR.

cesarvarela (approved):

cesarvarela:
I'm not too fond of this solution because it will push us above the 10k records limit of the free plan quicker, but we haven't found a better solution yet, so :shipit:

Issue #876, opened 2022-07-28
Deleting Reports from the UI Breaks the Build


smcgregor:
I deleted a report on an incident via the UI and it broke the build. I believe it is a problem in the foreign key outside the reports collection not having anything to point to.

1. [x] Remove the foreign key
2. [ ] Update the `delete` action to do this automatically

The offending production build: https://app.netlify.com/sites/aiid/deploys/62e2277926f49700087fa2fb

smcgregor:
I manually removed report 1116 from incident 60 on the incidents collection using the MongoDB interface. This should fix the build, but the GraphQL endpoint for deleting reports still needs an update for this.

cesarvarela:
Interesting, I expected that because we are using the build-in delete a and the built-in relationships, it would "cascade," but it does not.

The fix should be to manually call the `unlinkReportFromIncident` mutation after deleting.

lmcnulty:
I was coming up with a solution (#878) that just uses `updateIncident` rather than `unlinkReportFromIncident`, but maybe the latter would be cleaner.

cesarvarela:
Oh, I just looked, and it is not implemented yet, so we'll have to stick to your solution.

Issue #586, opened 2022-05-04
Report caught errors to Rollbar


cesarvarela:
Updating the toast component, so it reports to Rollbar every time it is shown may be a start.

pdcp1:
Closing this since it was already merges into Staging

PR #878, opened 2022-07-28
Delete report from incident when report is deleted

lmcnulty:
Addresses #876. 

[Tests pass](https://app.netlify.com/sites/lmcnulty-staging/deploys/62e408ba69875210abe6fc71#L725) on my copy of the database after running migration to remove orphan reports, although the deploy is still failing with:

> The function "__DSG" is larger than the 50MB limit. Please consider reducing it.
> The function "__SSR" is larger than the 50MB limit. Please consider reducing it.


site/gatsby-site/src/pages/cite/edit.js
```diff
@@ -101,11 +102,27 @@ function EditCitePage(props) {
...
+          },
+        },
+      });
+
       addToast({
         message: `Incident report ${reportNumber} deleted successfully.`,
         severity: SEVERITY.success,
       });
     } catch (e) {
+      console.error(e);
```
cesarvarela, line 125: leftover

site/gatsby-site/src/pages/cite/edit.js
```diff
@@ -84,6 +84,7 @@ function EditCitePage(props) {
...
```
cesarvarela, line 87: letfover

cesarvarela (changes requested):
It needs to add a mock to the delete report test because, as it is, this is unlinking reports in the live DB 💥

lmcnulty:
I've added a `conditionalIntercept`. Does that look right to you, @cesarvarela ?

site/gatsby-site/src/pages/cite/edit.js
lmcnulty: Perhaps this is a silly question, but is it bad to log an error when an actions fails? It seems like it could help some people report bugs.

cesarvarela: The error will get kind of lost there, we are using (or we should be using) Rollbar for this.

 I created this issue a long time ago, and I think this is where errors should be logged: https://github.com/responsible-ai-collaborative/aiid/issues/586

(Uncaught errors are already being captured and logged by Rollbar)

lmcnulty: Got it!

site/gatsby-site/cypress/e2e/citeEdit.cy.js
```diff
@@ -31,15 +31,14 @@ describe('Edit report', () => {
...
-        req.body.operationName == 'FindParentIncident' && req.body.variables.report_number == 10,
+      (req) => req.body.operationName == 'FindIncident',
       'findIncident',
       incident
     );
 
     cy.visit(url);
 
-    cy.wait(['@findReportWithTranslations', '@findIncident']);
+    cy.wait(['@findReportWithTranslations', '@findIncident'], { timeout: 30000 });
```
cesarvarela, line 41: Not need for a custom timeout here, it is already 30000

site/gatsby-site/cypress/e2e/citeEdit.cy.js
```diff
@@ -201,21 +200,29 @@ describe('Edit report', () => {
...
 
+    cy.conditionalIntercept(
+      '**/graphql',
+      (req) => req.body.operationName == 'UpdateIncident',
+      'updateIncident'
+    );
+
+    cy.visit(url);
+
+    cy.wait(['@findReportWithTranslations', '@findIncident'], { timeout: 30000 });
```
cesarvarela, line 218: ditto

site/gatsby-site/cypress/e2e/citeEdit.cy.js
```diff
@@ -347,7 +354,7 @@ describe('Edit report', () => {
 
     cy.visit(`/cite/edit?report_number=23`);
 
-    cy.wait('@findReportWithTranslations');
+    cy.wait('@findReportWithTranslations', { timeout: 30000 });
```
cesarvarela, line 357: ditto

cesarvarela (changes requested):
I want to merge this one, but I think the interception is not working. I ran the `integrity` test before and after, but the test didn't pass the second time, and the report was orphaned.

cesarvarela:
@lmcnulty found the reason why this wasn't working, the `cy.incercept` method lets the request go through if the response is not mocked:

https://github.com/responsible-ai-collaborative/aiid/pull/891/files

cesarvarela:
I've reset the staging DB to prod, so this build should start going green again once the similar incidents migration completes

site/gatsby-site/cypress/e2e/citeEdit.cy.js
lmcnulty: Hmm... when I ran it locally it would not pass without that. I think it was defaulting to 8000 for me. 🤷‍♀️

lmcnulty:
Wow, it feels like it's been a long time since I've seen a [successful build](https://app.netlify.com/sites/staging-aiid/deploys/62e9909c07a189045beacd00).

site/gatsby-site/cypress/e2e/citeEdit.cy.js
cesarvarela: [It should be pulling it from here:https://github.com/responsible-ai-collaborative/aiid/blob/65ff4d5ba9574adf7f18b057d70886b8b9e1cb45/site/gatsby-site/cypress.config.js#L7 which is set to `15000` 🤔

cesarvarela (approved):

cesarvarela:
Yeah, we plan on running a separate test environment per PR instead of all PRs sharing the same env. That will hopefully improve this

Issue #868, opened 2022-07-27
Suggestion - bring back quick incident title view when incident id is inputted


LKchemposer:
Could I have the feature where, once an incident id is assigned the title of the incident is shown right below, back?

It was a great quick sanity check to confirm whether the to-be-added report belongs to the incident from just its incident title. As of now, the report titles are shown in the "Incidents reports matched by ID: X" section, but that requires scrolling down a bit.

lmcnulty:
How about we move each of the "Incident reports matched by..." right under the corresponding input? That seems like it would solve this issue and generally make the feature more useful.

lmcnulty:
I've opened [a pull request](https://github.com/responsible-ai-collaborative/aiid/pull/894) for this that makes the list of related reports show up right under the incident ID input.

![image](https://user-images.githubusercontent.com/25443411/182690859-50dace84-8892-4df5-9f68-bb27220b2c84.png)

LKchemposer:
Would it be possible to also show the `title` of the incident (and at max, 5 reports of that incident)? This feature is mainly used as a sanity check to confirm that the to-be-ingested report fits the incident id, so it doesn't need to take up a lot of space.

lmcnulty:
This is solved my #894 now in staging, so I'm closing this for now.

PR #894, opened 2022-08-03
Show related incidents below input

lmcnulty:
Addresses #868. Makes each of the the related incident views appear under the corresponding form field. Also removes the option to set similarity based on the URL or ID, as those will identify the same incident, not similar ones, which could be confusing.

image


smcgregor:
I think it would be better to display the incident title and potentially description here, rather than the complete list of reports. Some reports do not have informative titles, whereas we decide the incident titles.

lmcnulty:
The `RelatedIncidentsArea` can now display reports or incidents. Incidents are displayed under the incident id field and the text field:

image

---

image

Whereas for the fields that are attributes of the report (author, date published, and url) we still display the reports instea:

image

cesarvarela (changes requested):
Looks good but it is missing translations

site/gatsby-site/src/components/RelatedIncidents.js
```diff
@@ -55,11 +56,11 @@ const reportsWithIncidentIds = async (reports, client) => {
   }));
 };
 
-const searchColumns = {
+const allSearchColumns = {
   byDatePublished: {
     header: (incident) => (
       <>
-        Incidents reports matched by published date: <b>{incident.date_published}</b>
+        <Trans>Incidents reports matched by published date:</Trans> <b>{incident.date_published}</b>
```
cesarvarela, line 63: Informes de incidentes por fecha de publicación:

(this should be only one trans component that includes the `b` tags)

site/gatsby-site/src/components/RelatedIncidents.js
```diff
@@ -84,20 +85,24 @@ const searchColumns = {
   byIncidentId: {
     header: (incident) => (
       <>
-        Incidents reports matched by ID: <b>{incident.incident_id}</b>
+        <Trans>Incident matched by ID:</Trans> <b>{incident.incident_id}</b>
```
cesarvarela, line 88: Incidentes por ID:

(this one needs to be in one Trans component too.)

site/gatsby-site/src/components/RelatedIncidents.js
```diff
@@ -84,20 +85,24 @@ const searchColumns = {
...
+    editSimilar: false,
+    editId: false,
+    showIncidents: true,
   },
 
   byAuthors: {
     header: (incident) => (
       <>
-        Incidents reports matched by authors: <b>{incident.authors}</b>
+        <Trans>Incidents reports matched by authors:</Trans> <b>{incident.authors}</b>
```
cesarvarela, line 105: Informes por autor:

(ditto)

site/gatsby-site/src/components/RelatedIncidents.js
```diff
@@ -111,21 +116,35 @@ const searchColumns = {
   byURL: {
     header: (incident) => (
       <>
-        Incidents reports matched by URL: <b>{incident.url}</b>
+        <Trans>Incidents reports matched by URL:</Trans> <b>{incident.url}</b>
```
cesarvarela, line 119: Informes por URL

(ditto)

cesarvarela (changes requested):
Next time please create the i18n keys in the appropriate json and we'll comment with the translations there, so it is faster than me having to search for each <Trans> component 🍷

lmcnulty:
Sorry about the missing keys – I would have added them but thought we had decided not to do that on account of a future system to detect missing translations automatically.

cesarvarela:
Yea, I think I miscommunicated that, once French is added we'll have to look into the automated thing, but for now just create the empty JSONs and just ping either Clara, Pablo or me, and we'll give the translated text.

cesarvarela (approved):

PR #915, opened 2022-08-08
Increase number of retries for semanticallyRelated

lmcnulty:
The semanticallyRelated api usually fails on a cold start. I added retries to try to address this, but it may not have been aggressive enough. This PR increases the retries and wait time. It will be taken out of draft when I'm satisfied that the numbers appropriate.


cesarvarela:
Closing as stale

Issue #917, opened 2022-08-10
Translation Deployment Issues


smcgregor:
WIP, I am still reviewing the functionality.

These are the issues associated with the translation feature.

# Blocking publication deployment

None

# High priority changes

* [x] The badge indicating something is translated by AI should link to the blog post
* [x] This menu is not localized. I think the English should also read: "View the original report at its source" and "View the report at the internet archive"
<img width="803" alt="Screen Shot 2022-08-10 at 12 09 54 PM" src="https://user-images.githubusercontent.com/64780/183998516-c6ad0243-976e-4436-bb21-572c5a401608.png">

* [x] There are some bad links in this list: https://staging-aiid.netlify.app/es/about_apps/
* [x] We should blur the offensive image of the blog post unless someone clicks it
* [x] Update the announcement blog post date to the 11th
* [ ] We want to be indexed in search engines in Spanish properly. I don't know what is involved here, but we need to look into it.
* [x] `Did our AI mess up? Flag  the unrelated incidents` Does not provide an obvious location to flag the non-related incidents.
* [x] Link in the question mark is not localized
<img width="444" alt="Screen Shot 2022-08-10 at 1 35 56 PM" src="https://user-images.githubusercontent.com/64780/184015143-fae9e029-cdad-4532-bce0-938e0cc93b5a.png">

* [x] Discover application has unused language selection parameter. For example, https://incidentdatabase.ai/es/apps/discover/?display=details&lang=en&page=1
* [ ] The instructions that pop up when you hover on the question mark of the submit form at currently always in English
* [ ] Toasts, like the one that pops up with submitting a report, have some non-localized links.

smcgregor:
Hello @cesarvarela @lmcnulty @clari182 @pdcp1 , plugging back in for a short bit to get the translation feature out. I am reviewing the functionality now and creating a backlog associated with it. So far none of the issues are blocking rollout, but they will need to be addressed in short order. Please divide and conquer!

I'll edit the issue as I encounter hot points.

cesarvarela:
@smcgregor you mean everywhere the badge is present it should link to the blog post? or only in the citation page?

If there is nothing particularly complex I can tackle them all

smcgregor:
So far, only tweaks. I am going to get the deployment out in a moment so we can do the last changes in staging.

LKchemposer:
small UI issue, where the `X` and the magnifying glass icon of the search bar are on top of the dropdown menu for language change.

[...]

PR #918, opened 2022-08-10
Fix i18n small issues 3

cesarvarela:
- [x] The badge indicating something is translated by AI should link to the blog post
- [x] This menu is not localized. I think the English should also read: "View the original report at its source" and "View the report at the internet archive"
- [x] There are some bad links in this list: https://staging-aiid.netlify.app/es/about_apps/
- [x] We should blur the offensive image of the blog post unless someone clicks it
- [x] Update the announcement blog post date to the 11th
- [x] Did our AI mess up? Flag the unrelated incidents Does not provide an obvious location to flag the non-related incidents.
- [x] Link in the question mark is not localized

PR #920, opened 2022-08-10
Add localized link to (?) icon in similar incidents

lmcnulty:
Fixes issue in https://github.com/responsible-ai-collaborative/aiid/issues/917

### DIFF
--- a/site/gatsby-site/src/components/cite/SimilarIncidents.js
+++ b/site/gatsby-site/src/components/cite/SimilarIncidents.js
@@ -11,8 +11,7 @@ import { UPDATE_INCIDENT } from '../../graphql/incidents';
 import md5 from 'md5';
 import { useUserContext } from 'contexts/userContext';
 import useToastContext, { SEVERITY } from '../../hooks/useToast';
-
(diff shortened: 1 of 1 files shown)


cesarvarela:
woot I'm working on these! The only one I'm missing is to show the translated titles on the similar incidents cards:

image

Not sure if this is possible tho, are those the titles of the incident or some of the reports linked to the incident?

lmcnulty:
Not sure if you're still looking for this, but those are the titles of the incidents. They should all have titles by now, but if one of them doesn't it falls back to the title of the first report.

cesarvarela:
@lmcnulty then we should show the incident title as it was before when the selected language it is not English, because we don't have translations for it yet, something like here:

https://github.com/responsible-ai-collaborative/aiid/blob/master/site/gatsby-site/src/templates/cite.js#L136

Can you work on that?

cesarvarela:
Closing because this is being dealt with in #918  (the comment above still applies though)

lmcnulty:
@cesarvarela Ok, I'm on it.

cesarvarela:
thanks 🙏

PR #922, opened 2022-08-10
Blur offensive image

lmcnulty:
Makes the offensive image blurred and shown on click, as mentioned in #917. Also sets the blog post date to the 11th.

"Sensitive: click to view" still needs to be translated.


cesarvarela:
A couple of things should be improved:

- add the component as a short code, so there is no need to import it. Example: https://github.com/responsible-ai-collaborative/aiid/blob/staging/site/gatsby-site/src/components/ui/MdxComponents.js#L25
- instead of loading the image statically it should use Gatsby's image pipeline (it should use`gatsby-plugin-image`)

No need to do this right now because we are already using the `Details` component for this, and it is good enough to go live with it, but this is definitively nice to have.

lmcnulty:
@cesarvarela I've made those change and added alt-text to the images on the English blog post while was there.

site/gatsby-site/blog/multilingual-incident-reporting/index.mdx
```diff
@@ -8,18 +8,20 @@ author: 'Sean McGregor and Cesar Varela'
 slug: '/blog/multilingual-incident-reporting'
 ---
 
+import SensitiveImage from '../../src/components/images/sensitive';
```
cesarvarela, line 11: This shouldn't be necessary

cesarvarela (commented):
Removing the extra `import` and good to go 👍

PR #977, opened 2022-08-18
Update dependencies

cesarvarela:
This should allow us to do a npm install without the legacy peers flag

Once merged we should Delete `NPM_VERSION` &  the `NPM_FLAGS` flags from Netlify


site/gatsby-site/package.json
```diff
@@ -127,6 +127,7 @@
     "lint-staged": "^10.5.3",
     "netlify-plugin-cypress": "^2.2.0",
     "postcss": "^8.4.14",
+    "postcss-reporter": "^7.0.5",
```
pdcp1, line 130: @cesarvarela Is this `postcss-reporter` package a replacement for an outdated package? Or is something new that we are including in this PR?

cesarvarela: I was trying to debug the build failing and forget about it, it might be necessary, but it is for another PR.  Deleted.

pdcp1 (approved):
Looks good 👍

Issue #954, opened 2022-08-17
Navigation icons


lmcnulty:
We should use icons to guide navigation through the site. For instance, in the sidebar, each item could begin with icon, such as a magnifying glass for Discover.

Issue #851, opened 2022-07-25
Separate Documentation Elements of Site


smcgregor:
Currently the site is designed around documentation. While we need to maintain proper documentation of the applications and processes, we should separate the design considerations of docs and the presentation of the data in the database. The first step on this is a Figma (or similar) design mockup of the site navigational structure and how docs+database apps fit together.

lmcnulty:
I think we can put most of the documentation under the "About" section of our navigation hierarchy, then split the apps between (1) those that help people find and examine the incidents in the database and (2) those that are helpful to those adding incidents to the database or otherwise participating.

I imagine the hierarchy of the navigation being structured something like this:

- Welcome to the AIID
- About
  - People (Board of directors, collaborators, transcluded leaderboard)
  - Incident Report acceptance criteria
  - Roadmap (database roadmap → roadmap)
    - In Development
    - Future
    - Delivered
    - Initial collection methodology
  - Researcher Guide – I think this should be limited to things that unlikely to be relevant to non-researchers, like citation information. Its current child pages I think can go elsewhere in the hierarchy.
  - Taxonomies
- Blog (Aiid Blog → Blog)
- Explore Collection
  - Discover Incidents
  - TSNE Visualization – This is interesting enough that it could get its own page.
  - Wordcounts
  - Taxonomy Graphs – We currently have pie charts for items in the CSET database, but https://incidentdatabase.ai/taxonomy/cset, but to get to them you have to go to "taxonomies" in the sidebar, then pick "CSET", then scroll past a couple paragraphs of text. I think it would be nice to have the graphs centered on their own page and link to the details on the taxonomy from there.
- Participate
  - Submit Incident Reports
  - Leaderboard
  - Submission Queue
  - Flagged Incidents
  - Classifications View

smcgregor:
* `TSNE Visualization – This is interesting enough that it could get its own page.` Agree. It could be a "summary" (this is what we have been calling pages that are reports on the data in the database) or potentially an "app" (these generally have more dynamic interaction with the database such as read and write capabilities).
[...]

PR #979, opened 2022-08-18
Restructuring

lmcnulty:
Addresses #851 and #954.

- Moves documentation links to the footer, now in a columnar layout

  image

- Adds icons to the sidebar items

  

  - The blog stays in the sidebar. Even though it's documentation of sorts, it's intended for the general public and we have space for it currently.

- Rearranges the taxonomy page template to put the first graph above the fold.
  - It would be nice to eventually incorporate the list of categories we use as links to the discover app into the graph itself.
  
  image

- ~~"Taxonomies" link in sidebar goes directly to CSET~~
  - ~~Header contains a link to `/taxonomies` for switching and learning about how we use taxonomies.~~
  - ~~I think going directly to a display of an implemented taxonomy with graphs, etc. will be more engaging than hiding it behind the taxonomies list. People are more likely to care about what taxonomies are available once they've seen one.~~
    - ~~Perhaps we could add a dropdown menu to switch between taxonomies?~~
    - In revision, I've instead added a carousel containing some of the more interesting charts under the CSET bullet on the /taxonomies page.
      
      https://user-images.githubusercontent.com/25443411/186501120-87cbc50b-c8c1-4dd7-8ea0-e3d6f4f32c5b.mp4


smcgregor:
* Design looks good in the screen captures you provide.
* Please link the taxonomies nav to the taxonomies page as it currently exists. CSET is working on a V2 that would deprecate the V1 and @npit is working on another taxonomy as well. Linking directly to a single taxonomy compromises this element of flexibility we are looking to emphasize. I think a better solution to the problem you are highlighting is to have cards/short panels that are graphical and can display on the current taxonomy listing. Maybe presenting the most informative pie chart in a card presentation?

cesarvarela:

cesarvarela:
@lmcnulty please fix conflicts, and now that #1001 is merged please remove all `tw` prefixes.

lmcnulty:
@cesarvarela Done.

lmcnulty:
@cesarvarela Fixed.



I had to set the sidebar ul padding to 0 with inline styles, because bootstrap css was overriding the tailwind classes.

cesarvarela:
`Should submit a new report not linked to any incident once all fields are filled properly`

It is failing and it is usually not flaky I believe. Can you make sure it is not failing because something else  broke?

cesarvarela:
The test should use `data-cy=toast` anyway, I'll merge and if it is broken fix it in another PR

PR #983, opened 2022-08-19
Submission length

### LINKED ISSUE #846: User is not notified that the report text is too long
All fields were filled in, but the report was not submitted upon hitting the "Submit" button. Refresh did not resolve the issue, unfortunately. Also, not sure if this issue can be replicated 😞 

https://user-images.githubusercontent.com/39468345/180720321-63733483-800e-42c2-8178-09c7672f8f84.mov

lmcnulty:
Resolves #846. Makes the field for the report text show errors the same way as the other fields.

image


cesarvarela:
A couple questions:

- Why are not the error messages being set through yup?
- Why not use the feedback class from react-bootstrap?

lmcnulty:
There was no good reason, I just didn't that those options existed. Changed.

cesarvarela:

PR #992, opened 2022-08-22
Use the new Atlas Service App graphQL URL format

pdcp1:
Atlas App Service has changed the GraphQL URL format for new apps from:

`https://realm.mongodb.com/api/client/v2.0/app/<REALM_APP_ID>/graphql`

to:
`https://<REALM_APP_REGION>.aws.realm.mongodb.com/api/client/v2.0/app/aiidstitch2-<REALM_APP_ID>/graphql`

So we have to change the code to use a new ENV variable `GATSBY_REALM_APP_GRAPHQL_URL` with the full URL value instead of just replacing the Realm App Id.

To get the `GATSBY_REALM_APP_GRAPHQL_URL` value, go to Atlas Service App page, then select "GraphQL" on the left side of the page, and then copy the GraphQL Endpoint URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ Create a `.env` file under `site/gatsby-site` with the following contents:
 
 ```
 GATSBY_REALM_APP_ID=aiidstitch2-vsdrv
+GATSBY_REALM_APP_GRAPHQL_URL=[to be updated in "Deploy the Realm App" section]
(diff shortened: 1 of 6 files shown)


pdcp1:
Closing this PR since we should use `Global` Atlas App Service instead of `Local`.
No code change required.

Issue #969, opened 2022-08-18
Landing page should have row of incidents


lmcnulty:
Essentially replacing the "Latest incident report" with a row of items from Discover. Could be a carousel.

Issue #896, opened 2022-08-04
Change Relevant UI from Reports to Incident Titles


smcgregor:
Elements like those below are currently scoped to incident reports rather than incidents. Now that we have guaranteed incident titles and descriptions, we should change everything that is only showing incident report details for the purpose of presenting an incident to now use the incident titles, and potentially descriptions.

<img width="692" alt="Screen Shot 2022-08-04 at 8 49 19 AM" src="https://user-images.githubusercontent.com/64780/182890861-bf689519-d93f-490b-8d86-8999b2374e4d.png">

lmcnulty:
I believe this is resolved now by #896 and #894.

PR #993, opened 2022-08-22
Use incident titles instead of report titles

lmcnulty:
Addresses #896. The specific instance named in the issue is solved by #894.

- Uses incident titles in the RandomIncidentsCarousel
  
  

- Adds incident titles to /summaries/incidents

  

I haven't found any other remaining places where we use the report titles when incident titles exist. There's the "latest report" item on the landing page, but that should be replaced in dealing with #969.


cesarvarela:

PR #1024, opened 2022-08-26
NLP Incident Detection

lmcnulty:
[Deploy Preview w/ Forked DB](https://deploy-preview-10--lmcnulty-staging.netlify.app/apps/newsSearch/) -- The necessary candidates collection is missing from staging, but it's now on production. I believe we're hoping to get this in by Tuesday.

---

Part of **Epic 3.3**: Incident collection and NLP incident detection tooling

https://user-images.githubusercontent.com/25443411/189740757-dff9720f-a2e9-46d5-b0c1-634d90add263.mp4

<details>
<summary>Full-page screenshot</summary>

</details>

This adds a page at `/apps/newsSearch` ~~which hits the new API endpoint `/api/parseFeeds` to get a list of links from RSS feeds of technology-related news sites. It then uses the existing `/api/parseNews` endpoint to extract the article text.~~ It now uses the data produced by the job in [nlp-monitoring](https://github.com/responsible-ai-collaborative/nlp-monitoring), which searches for AI-related keywords and, if found, hits the NLP api to get embeddings. It sorts the results according to their cosine similarity with the mean embedding of all reports in the database.

I initially tried just relying on the NLP, with no keyword search, but I found that it was mostly detecting whether something was an "incident," not whether it was "ai." The top marches tended to be security issues with no AI involved. E.g.

- [Plex imposes password reset after hackers steal data for >15 million users](https://arstechnica.com/?p=1875864) similarity: 0.9979934723877592
- [Twitter whistleblower raises security concerns](https://www.bbc.co.uk/news/technology-62633191?at_medium=RSS&at_campaign=KARANGA) similarity: 0.9979198660556315
- [LastPass Says No Passwords Stolen in Data Breach](https://www.cnet.com/tech/services-and-software/lastpass-says-no-passwords-stolen-in-data-breach/#ftag=CAD590a51e) similarity: 0.9978390209434157

Also, NLP is expensive and keyword searching is cheap.

---

Issues:

- ~~Currently this is not very optimized. All processing begins at page load, there's no caching, and network requests are all made one at a time.~~

- ~~After a certain number of articles are processed, my local dev server always crashes with `FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory`.~~

- Relying on a keyword search seems to work alright, but it's uncool.

- Instead of comparing to the average of every embedding, it might be better to compare to averages of reports with particular taxonomic labels. So, check if it's similar to the self-driving car incidents OR to the discrimination incidents OR ...

- It might also help to compare to sets of reports that aren't AI incidents but could be mistaken for them. E.g. collect some non-AI security reports, average their embeddings together, then ignore results that match the security incidents much more closely than the AI incidents.

<details>
<summary>Old Screenshot</summary>

</details>


lmcnulty:
**Note**: The backend for this feature will end up in the [nlp-monitoring](https://github.com/responsible-ai-collaborative/nlp-monitoring) repo.

cesarvarela (commented):
@lmcnulty Looks like this https://github.com/responsible-ai-collaborative/nlp-monitoring/actions/runs/3202676302/jobs/5231899034 is not running for staging?

lmcnulty:
@cesarvarela That's correct. We want to avoid hitting the NLP api too much, so I think if we need the candidates data on staging we should copy it from production. I could add that in the workflow, but someone would need to add the staging connection string as a repository secret in nlp-monitoring.

PR #1025, opened 2022-08-26
Reduce TSNE nIter

lmcnulty:
Reducing the TSNE nIter will put less strain on the build system, which might reduce crashiness. I think 1,000 works fine, I just set it to 10,000 for good measure.


cesarvarela:
Wasn't aware of this, looks like it is not necessary for $now

FYI looks like I fixed the crashing here https://github.com/responsible-ai-collaborative/aiid/pull/1022

Can you check why the TSNE viz isn't working please? I don't know exactly what too look for.

It is the only red mark left https://app.netlify.com/sites/staging-aiid/deploys/630d1cf391d70b000a6f311c

Issue #563, opened 2022-04-27
Add Navigation to Forms


smcgregor:
In particular, when you edit an incident you are told it saves, but you need to use the back button to get back to the incident citation page.

smcgregor:
After you save the form, it gives you a toast with a link. There should also be a link somewhere on the form for going back to the citation page for the incident since you should not need to save the changes to get back to where you came from.

Also, with this issue the "report not found" message that flashes before the report loads should be changed to a spinner.

PR #1027, opened 2022-08-29
Add links to toasts, use tailwind colors

lmcnulty:
To handle #563, this adds links in the toasts on form submission, providing a way back to the modified incidents.

It also sets the background colors to ones used in tailwind.


site/gatsby-site/src/hooks/useToast.js
```diff
@@ -13,25 +13,25 @@ export const SEVERITY = {
     key: 'success',
     icon: faCheckCircle,
     faClass: 'far fa-check-circle fa-lg',
-    color: '#43a047',
```
cesarvarela, line 16: Can this be set somewhere else using classes? It would make theming easier, and make it easier to have an eventual dark mode.

clari182: If the colors already exist in tailwind, they could be applied as classes after faClass like:
`faClass: 'far fa-check-circle fa-lg',
bgColor:'bg-green-700'
`

And then used as:
`<ToastBodyContent className={severity.bgColor}>`

lmcnulty: @clari182 When I try this, I just get a white background. When I inspect the element, it shows class="… bg-green-700 …", but the browser doesn't apply any rules based on it. Do you have an idea why that might be? Is that because it's now inside a ".bootstrap"?

clari182: @lmcnulty Do you have the latest staging changes?
This could be happening because I added some changes in the tailwind config that are on staging now.
- the prefix `tw-` has been removed
- the property "important" has been added:   `important: true` as first prop in `tailwind.config.js` inside module.exports
Something else that could be happening is if you're using the class dynamically. Tailwind doesn't allow classes to be used in that way like `bg-${color}` for example.
Let me know if this helps.

lmcnulty: @clari182 I thought those changes were already merged, but merging staging again seems to have fixed it.

site/gatsby-site/src/pages/cite/edit.js
```diff
@@ -216,19 +220,29 @@ function EditCitePage(props) {
...
+            Incident report {reportNumber} updated successfully.{' '}
+            <LocalizedLink to={'/cite/' + values.incident_id}>
+              View Incident {values.incident_id}.
+            </LocalizedLink>
+          </>
+        ),
         severity: SEVERITY.success,
       });
     } catch (e) {
+      console.error(e);
```
cesarvarela, line 245: please remove this

site/gatsby-site/src/pages/incidents/edit.js
```diff
@@ -45,7 +46,12 @@ function EditCitePage(props) {
       });
 
       addToast({
-        message: `Incident ${incidentId} updated successfully.`,
+        message: (
```
cesarvarela, line 49: Looks like this is missing translations?

site/gatsby-site/src/pages/cite/edit.js
```diff
@@ -179,13 +181,15 @@ function EditCitePage(props) {
...
         const embedding = incidentEmbedding(exSiblings);
 
-        await client.mutate({
-          mutation: UPDATE_INCIDENT,
-          variables: {
-            query: { incident_id: parentIncident.incident.incident_id },
-            set: { embedding },
-          },
-        });
+        if (embedding) {
```
cesarvarela, line 184: Looks like this PR does something else besides the title, whats the `embedding` check for?

cesarvarela (changes requested):

site/gatsby-site/src/pages/cite/edit.js
lmcnulty: I think when I was testing that the toasts appeared correctly, I noticed that the form couldn't be submitted when the incident being edited didn't have an embedding. Since the NLP api sometimes fails to return a result, we can't count on there being one.

site/gatsby-site/src/pages/incidents/edit.js
lmcnulty: I've added the translation calls and English on both sides. As far as I can tell, `<Trans>` doesn't work inside the toasts for some reason, so in a couple places I had to use two `t()`. I included complete sentences in each one, so I think that should work for any language.

cesarvarela: Oh yes, it works but needs to be provided with a `i18n` instance. My guess is that it can't reach react's context, but there is a workaround here: https://github.com/responsible-ai-collaborative/aiid/blob/30f6d9a23246d14d7dc89a9ad42eebe3eb0cc7d6/site/gatsby-site/src/pages/apps/submitted.js#L36 please do it that way if possible.

lmcnulty: @cesarvarela Fixed. Thanks!

cesarvarela (approved):

cesarvarela:
@lmcnulty it's conflicts time!

lmcnulty:
@cesarvarela Conflicts are fixed.

PR #1039, opened 2022-08-31
Improve blog cards styling

clari182:
[Deploy preview](https://deploy-preview-1039--staging-aiid.netlify.app/)

Uses tailwind styled based cards for the blog with a new "Read more" button


site/gatsby-site/src/components/styles/Docs.js
```diff
@@ -38,7 +38,6 @@ export const Edit = styled.div`
 `;
 
 export const StyledMainWrapper = styled.div`
-  max-width: 800px;
```
lmcnulty, line 41: This makes plain textual pages, like `/about`, [too wide to read comfortably](https://en.wikipedia.org/wiki/Line_length). I think it would be better to unset `max-width` for the blog post listing specifically.

image

lmcnulty: Otherwise, it looks good!

clari182: @lmcnulty fixed! Ready for another round

site/gatsby-site/src/templates/post.js
```diff
@@ -51,7 +51,7 @@ export default function Post(props) {
           page="post"
         ></SocialShareButtons>
       </div>
-      <StyledMainWrapper>
+      <StyledMainWrapper className="max-w-full">
```
lmcnulty, line 54: I think full-width is good for the blog index, but for the posts the readability issue of long lines still applies. I would just revert the change on this line, and then it will be good to go.

clari182: @lmcnulty done!

lmcnulty (approved):

PR #1020, opened 2022-08-25
Deploy to Production

cesarvarela:
- tailwind & bootstrap scoping (means we can use tailwind freely, and bootstrap can be limited only to forms)
- email-password/facebook signups
- menu restructuring
- TSNE visualization

PR #1042, opened 2022-09-01
Add light-outline variant to language switcher

### LINKED ISSUE #938: Language selector has too much visual weight
Using the bootstrap `btn-primary` class on the language switcher gives it visual weight equal to the most important actions available on each page, like "search" on the landing page. It's perhaps ok to highlight the feature prominently during its rollout, but as we move on it should probably receive less weight. I think it would look best if the button had no background – the down arrow is sufficient to indicate that it opens a menu. We could also add [the language icon](http://www.languageicon.org/) in white to make it easy to find without throwing off the visual balance of the site.

image

lmcnulty:
Resolves #938, mentioned in #1020.


cesarvarela (changes requested):
Please fix conflicts

lmcnulty:
@cesarvarela Fixed.

cesarvarela (approved):

PR #1047, opened 2022-09-01
Improve landing styles

clari182:
https://deploy-preview-1047--staging-aiid.netlify.app/

Applied Flowbite components to landing


clari182:
@smcgregor , with @lmcnulty  we've been working on applying Flowbite to the landing page.
It'd be great to have your thoughts on it, so you can let us know what you like/dislike.
Here's the PR preview so you can take a look.
https://deploy-preview-1047--staging-aiid.netlify.app/

smcgregor:
@clari182 this is looking good. I have a few comments on rough points (quick list below).

* The blank space caused by the cards being uneven in height is an eyesore
* Using badges to indicate the content type is a bit rough, but I think we need to develop some notion of composable blocks where the headers clearly indicate what the block is and the block contains the dynamic content.
* Some of the relative sizing of headers and contents should be made a bit more distinctive
* Maybe the leaderboards should not have the table outline? It looks very "busy".
* The image from the latest incident report doesn't have any sort of border separating it from the rest of the content

Note: this is a real quick take and I am hoping you will take it as indication on where to put more attention, and not necessarily take my input as a hard line requirement at this point. I'll do a more in-depth review when you mark it as ready for as much.

clari182:
@smcgregor I made some updates based on your comments. Let me know what you think.
I think we can still make more improvements but this is a good first step towards an improved landing page.

cesarvarela (changes requested):
Missing translation here:

image


ditto:

image

ditto (this isn't working on production either)
: 
image

smcgregor:
@clari182 I see Cesar beet me to it! One quick take: I think you need to set a max width. I have a very wide monitor and the page looks strange if you make the window too wide.

cesarvarela:
@clari182 please merge staging, there is no reason for the tests to fail

clari182:
@cesarvarela staging merged

clari182:
@smcgregor I set a max width for wider screens. Let me know if it looks better like this, and if we're g2g for merge. Thanks!

smcgregor:
@clari182 Thanks! That certainly fixed the width issue. I am going to approve this on my end. My only comment at the moment is that there is a little bit of a disconnect between elements that are in a card on the landing page, and elements that have a header outside any card. For instance, the "Latest Incident Report" header is outside a card while most of the others are inside a card. This seems a bit discordant to me, but I would guess that putting that component in a card may not look very good so I don't have much constructive to say here. In short, I suggest looking at putting everything in a card and if it doesn't look better, please ignore me...

smcgregor (approved):

clari182:
@smcgregor I agree. But I haven't found the right solution for that yet. But I'll create an issue to keep thinking about it.

cesarvarela (changes requested):
Safari:

The image goes haywire, not sure wth happens here, but I would first fix the wrong height:

image

image

firefox: 

image

image

Chrome: 

image

image


Everything else looks good but his kind of jumps at you

clari182:
@cesarvarela ready for review again. Here's the latest build: https://app.netlify.com/sites/staging-aiid/deploys/632a477390f5db378b9100e5

cesarvarela:
@clari182 please fix conflicts and it should also fix tests

cesarvarela (approved):

PR #1048, opened 2022-09-01
Make 1st report only switch persist refresh

lmcnulty:
Addresses item in https://github.com/responsible-ai-collaborative/aiid/pull/1020#issuecomment-1233459530.

> When 1st report only is turned on and the page is reloaded (or the URL is shared), the setting is not sticky.


clari182 (approved):

clari182:
Works as expected, approved!

PR #1049, opened 2022-09-01
Feature/french localization

clari182:
- French translations in place
- To test this locally you'll need to add fr to your `GATSBY_AVAILABLE_LANGUAGES` var


lmcnulty:
Once this is ready, would you like any review on the language? I have a fairly advanced level of written French – I tested into a [C1-level](https://en.wikipedia.org/wiki/Common_European_Framework_of_Reference_for_Languages#Common_reference_levels) French class while studying at Université Panthéon-Sorbonne – but if you're at a native or near-native level I probably won't be of any help.

clari182:
Thanks @lmcnulty but I've got it I think. I'm a french native speaker :)

cesarvarela (commented):
Is it possible to merge this without enabling the language? (To prevent this branch from conflicting ad infinitum)

Issue #1063, opened 2022-09-07
Issues in Staging


smcgregor:
Starting the list for the eventual deploy to production preview, since I am playing with staging at the moment.

- [x] The TSNE visualization is broken:

<img width="1070" alt="Screen Shot 2022-09-06 at 5 34 13 PM" src="https://user-images.githubusercontent.com/64780/188763370-70c55607-15e0-486c-b07a-5d17ac810286.png">

smcgregor:
I think this covers everything. This should be ready to deploy, but there are a few items that will need addressing in short order.

- [ ] The accessibility of the incidents application is much higher in this layout, so the annoying design elements are of higher importance. The spinner below does not look good and should be updated. The table that is displayed after the data loads should also be updated to match the table display @cesarvarela is working with on his entity branch, but that will be an issue when the entity pages land.

![Screen Shot 2022-09-12 at 3 16 22 PM](https://user-images.githubusercontent.com/64780/189768213-9141d79f-bfd6-422c-8ca6-9f1e5dbb9407.png)

- [ ] Nothing on this page is translated, and the links all direct to the english language pages. https://staging-aiid.netlify.app/es/apps/incidents/
- [ ] The search box on the landing page should be changed to 1700 from 1600.
- [ ] Clicking on modals appears to be triggering a re-render on the underlying page: https://staging-aiid.netlify.app/apps/discover?display=details&hideDuplicates=1&page=1 Maybe 
- [ ] The flag report button on the Discover application modal is not rendering properly,

![Screen Shot 2022-09-12 at 3 27 57 PM](https://user-images.githubusercontent.com/64780/189769534-1d6f64b6-9981-4218-ab89-a95a0881ff01.png)

- [ ] The footer on mobile is not margined properly (I believe @lmcnulty has a branch for this one)
- [ ] This page has a 404 and it does not give a spinner while the table is loading. https://staging-aiid.netlify.app/apps/classifications/
- [ ] IMPORTANT but not blocking: the submission form has taken a big step backward in usability. I don't want to list all the issues with it here, but this needs attention. Whoever takes on this issue should likely better segment submitting new incidents+reports and new reports better. Many of the fields before the user gets to the incident number field are irrelevant if the user is submitting to an already existing incident. Some of the fields also need their instruction popover text. Aside from these, there are a large number of hot points in terms of dynamic content that jumps things around and produces a rather scary experience.
- [ ] Double stated title

[...]

Issue #940, opened 2022-08-15
Arrows in "Random Reports" are not visible against light backgrounds


lmcnulty:
https://user-images.githubusercontent.com/25443411/184678721-cabed58c-befd-4ac8-aab1-a846ff5520ca.mp4

smcgregor:
Low priority on this one since this is the default way the component works out of the box and we have hotter points to work at the moment.

PR #1067, opened 2022-09-07
Switch taxonomy carousel to tailwind, remove hack

lmcnulty:
Switching to flowbite removes the ugly hack I used to force re-renders in the bootstrap carousel.

Also enables dark-mode on a class-level, since this component is only visible with dark controls. ~~We'll want to do the same thing for the random reports carousel because of #940~~ Actually dark mode and light mode both look equally bad there because there's always some places where it doesn't contrast with the background.

Now also makes the carousel query statically, per https://github.com/responsible-ai-collaborative/aiid/issues/1063#issuecomment-1260141144. This fixes the crash that was occurring before.


cesarvarela (changes requested):
Looks like the preview is crashing:

https://deploy-preview-1067--staging-aiid.netlify.app/taxonomies


image


image

cesarvarela (approved):

Issue #957, opened 2022-08-17
"Discover • Submit" affordances


lmcnulty:
These buttons don't have anything to indicate that they're interactive. Since they're under the logo, they can be mistaken for a tagline.

![image](https://user-images.githubusercontent.com/25443411/185219005-ff069e1a-5cd2-4fe9-be21-360c9e571a99.png)

Issue #945, opened 2022-08-15
Page headings


lmcnulty:
The top-level headings that are on most pages of the site need work. They take up a lot of vertical space and aren't aligned with other element of the page. It's also weird that the lower-level headings are larger, [as was pointed out in the think-aloud](https://github.com/responsible-ai-collaborative/aiid/issues/902).

![h1-space svg](https://user-images.githubusercontent.com/25443411/184687510-8ccd2e44-205d-4cba-b6fb-1163d059aa00.png)

The new incident titles can also get very long, which becomes a problem with the large bold text.

![image](https://user-images.githubusercontent.com/25443411/184686033-40842e70-10bd-4e2d-9c55-d486cdb406ae.png)

PR #1071, opened 2022-09-07
Quick access buttons, headings, and alignment

lmcnulty:
image

- Makes the "Discover", "Submit" buttons look like buttons, per #957.
- Aligns page heading with the buttons and makes them not smaller than `<h2>`s, per #945.
- Updates markdown pages to always use `<h2>` for the top section-level headings
- Aligns header bar elements to the sidebar


cesarvarela:
@lmcnulty can you make a new PR with only the fix for the double headings please?

lmcnulty:
@cesarvarela Here it is: #1084

cesarvarela:
@lmcnulty please fix conflicts

site/gatsby-site/src/global.css
```diff
@@ -502,12 +502,17 @@ blockquote {
 }
 
 .titleWrapper {
-  display: flex;
   flex-direction: column;
   align-items: flex-start;
-  padding-bottom: 20px;
   border-bottom: 1px solid rgb(230, 236, 241);
-  margin-bottom: 32px;
+  min-height: 5rem;
```
cesarvarela, line 508: Not necessary to be done on this PR, but we should be removing `styled-components` instances wherever possible.

site/gatsby-site/src/tailwind.css
```diff
@@ -4,20 +4,28 @@
...
 
-    h1 {
-        font-size: calc(1.375rem + 1.5vw);
-    }
-
     .h1, .h2, .h3, .h4, .h5, .h6, h1, h2, h3, h4, h5, h6 {
         @apply font-medium mb-2 mt-0 leading-tight;
     }
 
+    .h1, h1 {
```
cesarvarela, line 11: I'm not sure about customizing heading styles. Since we are adopting Flowbite, I expect them to look like what they do in Flowbite's docs.

Maybe we are missing their typography plugin? https://github.com/themesberg/flowbite-typography

@clari182, what do you think?

cesarvarela (commented):
@smcgregor @LKchemposer, this is a very visible change. What do you guys think about it?

site/gatsby-site/src/tailwind.css
clari182: @cesarvarela yes, the idea is to eventually remove all custom styles and use Flowbites. Maybe in some cases, there are still some concurrency issues with Tailwind/Bootstrap so this was necessary. But for good practice, it should be removed and fixed in another way.

smcgregor:
In playing with the deploy preview (no code review): 👍

site/gatsby-site/src/tailwind.css
cesarvarela: That sounds good, @clari182. Please write that in an issue or something; we can tackle this in a future "tech-debt-paying" week.

PR #1073, opened 2022-09-09
Candidates collection

lmcnulty:
Adds a collection of "candidate" submission added by the [nlp-monitoring](https://github.com/responsible-ai-collaborative/nlp-monitoring/) system. Goes along with #1024.


lmcnulty:
Merged into #1024.

Issue #1075, opened 2022-09-09
Staging issue: Can't create new reports


pdcp1:
Function `promoteSubmissionToReport.js` is not working on Staging.

```
Uncaught (in promise) Error: (QueryExceededMemoryLimitNoDiskUseAllowed) Executor error during find command :: caused by :: 
Sort exceeded memory limit of 33554432 bytes, but did not opt in to external sorting.
```

Error cases:
On page `/apps/submitted`
- Click on `Add New Report`: A new report is submitted to an existing incident
- Click on `Add New Incident`: A new report is submitted to a new incident

pdcp1:
Closing this since #1076 was merged to Staging

PR #1076, opened 2022-09-09
Fix promote submission to report

pdcp1:
This PR:

- Adds `projection` to sort queries on `promoteSubmissionToReport` function to improve performance
- Adds DESC index on `reports.report_number` and `incidents.incident_id`

Closes https://github.com/responsible-ai-collaborative/aiid/issues/1075

### DIFF
--- a/site/gatsby-site/migrations/2022.09.09T20.08.13.create-reports-and-incidents-index.js
+++ b/site/gatsby-site/migrations/2022.09.09T20.08.13.create-reports-and-incidents-index.js
@@ -1,0 +1,32 @@
+const config = require('../config');
+
+/**
+ *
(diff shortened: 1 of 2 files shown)


cesarvarela (changes requested):
After looking here:
https://github.com/responsible-ai-collaborative/aiid/blob/master/site/realm/graphql/custom_resolvers/mutation_promoteSubmissionToReport.json#L21

, we shouldn't be using a projection to return a partial incident because the graphQL schema expects a full Incident.

I would say let's undo the updates to the function and only push the migrations, considering that it already fixes the issue.

cesarvarela (approved):

pdcp1:
@cesarvarela done! I just keep the migrations

PR #1080, opened 2022-09-12
Feature/submit flowbite

### LINKED ISSUE #1175: Put the `alleged deployer` field before/above `alleged developer` field
Also, a minor ask (optional): if you could put the `alleged deployer` field before/above `alleged developer` field, that would help me so much. I've recently run into issues where I put the wrong entities in, because I was in a flow state and I got used to the workflow of deployer-before-developer of the incident table. (a classic case of UX design influences behaviors)

_Originally posted by @LKchemposer in https://github.com/responsible-ai-collaborative/aiid/issues/1112#issuecomment-1254735968_

clari182:
[Deploy preview](https://deploy-preview-1080--staging-aiid.netlify.app/)

Also:
- Fixes #1175


clari182:
@smcgregor here's the PR for the new wizard/form for submissions. 
https://deploy-preview-1080--staging-aiid.netlify.app/
I know Luna is also working on the form improving some other things (PR #1124), so I'll mark this as ready for review once it's merged. 

Here is the list of things I still have to fix on this PR. Feel free to add what you think should be changed/improved.

- [ ] Fix max width of the form container. On bigger screens, it's too wide
- [ ] Fix Tags like input styling to match the normal inputs

smcgregor:
@lmcnulty can you take first UX review and I will follow on?

lmcnulty:
- The "previous" button should get less emphasis than the "next" button. I would set it to `color="light"`. They should also probably be the same width.

  image


- Having this in blue makes it look like a link: 
  

- This seems like it should be a multi-line input:
   

- The normal link color doesn't contrast enough against the dark gray of the popovers: 
  https://webaim.org/resources/contrastchecker/?fcolor=0F66E7&bcolor=22262a

  image

- This is a matter of opinion, but I don't really like the Flowbite "success" buttons. To me, they seem too dark to indicate "success" – they look more like primary buttons of an app with a green color scheme. I think this could look good as a `<Button gradientDuoTone="greenToBlue">` instead.

- There are missing/uneven margins in a few places:
   

  

  e.g. 
  

---

I have some tentative opinions about the arrangement of the form – this demands further thought, though. I kind of think the report text input should be on the first page. That will give the NLP api time to respond while the user is filling out the other fields. I also wonder if we could dedicate a page to establishing whether the report is for a new incident, because the distinction is kind of confusing for people new to the project. Perhaps instead of showing the related incidents under the corresponding input, we could put them all on, say, the second page, along with the incident ID input.

clari182:
@lmcnulty @smcgregor I made some updates based on Luna's comment. The only thing I haven't tackled is this:

> I have some tentative opinions about the arrangement of the form – this demands further thought, though. I kind of think the report text input should be on the first page. That will give the NLP api time to respond while the user is filling out the other fields. I also wonder if we could dedicate a page to establishing whether the report is for a new incident, because the distinction is kind of confusing for people new to the project. Perhaps instead of showing the related incidents under the corresponding input, we could put them all on, say, the second page, along with the incident ID input.

Let me know what changes can be made to the form regarding steps/order of fields.

cesarvarela:
This should be disabled if the field is empty or not a valid URL:

image


also show a invalid URL message:

image

(unless it was intentionally removed 🤔 )

clari182:
@cesarvarela fixed!

smcgregor:
This looks great! Suggestions:

1. In any multi-step process, there are significant risks of losing users at the transition between each of the steps. One way to mitigate this is to take partial submissions and rely on editors to complete partially filled submissions. For this, we could put two buttons on each of the stages: "Submit" and "Add More Info", and submitting in the first step only assigns the fields on the first form.
2. The deploy preview is not whitelisted on the server. Are you hooked into the staging backend? I can temporarily add the deploy preview so the dynamic elements should work.
3. The semantically similar (and other potential duplicates) panels break up the visual flow and make the form difficult to read. Yes, this is a huge problem with the current forms as well! Maybe try playing with badges in place of a whole panel? Maybe a second column on the form that slides into view? Mocks may be useful for this.
4. The image address and image preview could likely be on a single row, at least on desktop. The image is taking up a huge amount of space.
5. Needing to decide whether you select an incident ID or select an incident date is awkward. I don't like asking users to do things like leave something blank (yes, this is current practice, but let's improve it). Maybe a radio button asking whether it is an existing incident should conditionally display the date selector or incident selector?
6. `Description` needs help text.
7. Move `tags` just above the `editor notes` input. It is less important here.

I could not run the code start to finish due to the origin issue, but even with that, this is a big improvement over the current form.

clari182:
@smcgregor I have some follow-up comments:

> 1. In any multi-step process, there are significant risks of losing users at the transition between each of the steps. One way to mitigate this is to take partial submissions and rely on editors to complete partially filled submissions. For this, we could put two buttons on each of the stages: "Submit" and "Add More Info", and submitting in the first step only assigns the fields on the first form.

Yes I also think this is important and maybe all mandatory fields should be on the first step. Otherwise, the user might want to submit but gets stuck without knowing exactly why. I'm working on this now.

> 2.  The deploy preview is not whitelisted on the server. Are you hooked into the staging backend? I can temporarily add the deploy preview so the dynamic elements should work.

I don't have access to whitelist the preview URL, but I think it's best to add it while testing. I think once staging is merged the whitelist list is reset so the preview URL will be deleted.

> 3. The semantically similar (and other potential duplicates) panels break up the visual flow and make the form difficult to read. Yes, this is a huge problem with the current forms as well! Maybe try playing with badges in place of a whole panel? Maybe a second column on the form that slides into view? Mocks may be useful for this.

Agree, I haven't figure out how to improve this yet so I'm working on this now. I think that when no related reports are found we should not display the panels, and when some are found display them with selectable badges.

> * The image address and image preview could likely be on a single row, at least on desktop. The image is taking up a 
huge amount of space.

I reduced the image size a bit so it doesn't occupy as much space. I'll try other options to see how I can improve this further.

> * Needing to decide whether you select an incident ID or select an incident date is awkward. I don't like asking users to do things like leave something blank (yes, this is current practice, but let's improve it). Maybe a radio button asking whether it is an existing incident should conditionally display the date selector or incident selector?

Radio button sounds good

> * `Description` needs help text.

What would be the help text? I don't think there ever was one.

> * Move `tags` just above the `editor notes` input. It is less important here.

Done

smcgregor:
How about, for the popover description of the description:
```
Description

A short, factual, journalistically neutral, and complete description of the incident mentioning the incident/issue, location, and harm.
```

cesarvarela:
Some backgrounds have a grayish color and some are transparent, doesn't seem intentional.

image

cesarvarela:
While working on my pr, I've clicked many times on the submit button it let me submit the same thing multiple times:

image

It should be disabled while submitting, and maybe have a spiner?

clari182:
This is ready for review.
The only thing remaining is to improve the related incidents table but I'll create an issue and do this after this is merged.

site/gatsby-site/cypress/e2e/submit.cy.js
```diff
@@ -48,6 +58,7 @@ describe('The Submit form', () => {
     cy.get('button[type="submit"]').click();
 
     cy.wait('@insertSubmission').then((xhr) => {
+      console.log(xhr.request.body.variables.submission);
```
cesarvarela, line 61: please remove this

site/gatsby-site/cypress/e2e/submit.cy.js
```diff
@@ -33,13 +33,23 @@ describe('The Submit form', () => {
...
+    cy.get('[data-cy="to-step-2"]').click();
+
+    cy.wait(0);
+
+    cy.get('input[name="submitters"]').type('Something');
+
     cy.get('[name="language"]').select('Spanish');
 
-    cy.get('[data-cy="extra-fields"]').click();
+    cy.wait(200);
```
cesarvarela, line 48: This is 1/5th of a second, does this change anything?

site/gatsby-site/src/components/Layout.js
```diff
@@ -20,7 +20,7 @@ const Layout = ({ children, collapse, className, location }) => (
         />
       )}
       <div id="content" className="tw-content pb-5">
-        <div className={`${className ? className : ''} 50rem:max-w-full 50rem:relative`}>
+        <div className={`${className ? className : ''} max-w-full md:max-w-5xl 50rem:relative`}>
```
cesarvarela, line 23: This will affect almost every page on the site 🤔

site/gatsby-site/src/components/RelatedIncidents.js
```diff
@@ -118,7 +118,7 @@ const allSearchColumns = {
       return (
         <>
           <Trans authors={authors}>
-            Incidents reports matched by authors: <b>{{ authors }}</b>
+            Incidents reports matched by authors: <b className="break-all">{{ authors }}</b>
```
cesarvarela, line 121: This change forces breaks a
t
e
v
e
r
y
l
e
t
t
e
r

doesn't seem like we want this 🤔

site/gatsby-site/src/components/forms/Label.js
```diff
@@ -1,54 +1,62 @@
...
-                <Trans
-                  ns="popovers"
-                  i18nKey={`${popover}.text`}
-                  components={{ linkto: <Link /> }}
-                />
-              </Popover.Body>
-            </PopoverWrapper>
-          }
-          {...(show ? { show } : {})}
+      <OverlayTrigger
```
cesarvarela, line 41: This is because flow bite is not can't guess the correct tooltip position?

site/gatsby-site/src/components/forms/SubmissionWizard/StepOne.js
```diff
@@ -0,0 +1,337 @@
...
+
+  const handleSubmit = (values, last = false) => {
+    props.next(values, last);
+  };
+
+  return (
+    <StepContainer name={props.name}>
+      <Formik
+        initialValues={props.data}
+        onSubmit={() => {}}
```
cesarvarela, line 66: This doesn't seem right

site/gatsby-site/src/components/forms/SubmitForm.js
```diff
@@ -126,7 +132,7 @@ const SubmitForm = () => {
 
       await insertSubmission({ variables: { submission } });
 
-      resetForm();
+      // resetForm();
```
cesarvarela, line 135: 👀

site/gatsby-site/src/components/incidents/IncidentIdField.js
```diff
@@ -37,10 +37,15 @@ export default function IncidentIdField({
   }, []);
 
   return (
-    <div className="bootstrap">
-      <Form.Group className={className + ' bootstrap'}>
+    <div className="">
```
cesarvarela, line 40: Maybe delete the attribute entirely

cesarvarela (commented):
functional review:

- the loading overlay allows me to edit form fields. Not sure if this is intentional or not
- after submitting one report, there is no way to submit another without refreshing the page
- something is going on with validation/field updates:
  -  paste a URL and click fetch
  - delete contents of `title` field
  - enter an incident date and click submit
  - a submission with the previous title is submitted even if the text field is empty. It issues the old value as the title, which makes me think that something is going on between the `data` variable and the `values` variable from Formik.

site/gatsby-site/src/components/forms/SubmissionWizard/StepOne.js
clari182: Formik's onSubmit is required, but I can't use it in this case because I can't pass custom data. And in order to know if I'm submitting for a Next Step or for the Final Submit I need to pass a custom param.
Since Formik requires an onSubmit function I need to pass an empty function.

site/gatsby-site/src/components/forms/Label.js
clari182: No, it's because Flowbite React's tooltip doesn't work well:
- Once the tooltip is open there is no way to hover it, which is needed here.
So I decided to keep using Bootstrap OverlayTrigger as it was.

clari182:
> * the loading overlay allows me to edit form fields. Not sure if this is intentional or not
Fixed

> * after submitting one report, there is no way to submit another without refreshing the page
Fixed 

TODO:
> * something is going on with validation/field updates:
>   
>   * paste a URL and click fetch
>   * delete contents of `title` field
>   * enter an incident date and click submit
>   * a submission with the previous title is submitted even if the text field is empty. It issues the old value as the title, which makes me think that something is going on between the `data` variable and the `values` variable from Formik.
> 
> edit:
> 
> This feature greatly changes the submit flow, but the PR lacks new tests.
> 
> I think it should add new ones, like, testing submitting on step 1, step 2, step 3, etc.

cesarvarela:
@clari182 please fix conflicts

cesarvarela:
@smcgregor any comments before merging?

cesarvarela (commented):
@clari182 don't forget about conflicts

clari182:
@cesarvarela conflicts fixed!
@smcgregor I think this might deserve another review since it has the latest form changes now.

smcgregor:
I think the deploy preview domain may need to be whitelisted before this will work?

cesarvarela:
@smcgregor whitelisted

clari182:
@LKchemposer this is the PR that re-designs the submit form. It's now divided into steps. Please feel free to take a look and let me know your thoughts on design/functionality. Thanks!

smcgregor:
Overall, much improved. My review is below. I am 👍 for merging this to staging.

## Changes needed

The following changes can be made after this is merged to staging/production.

1. I believe Luna has some code targeting the old form where the incident can be given a title separately from the title of the report.
2. This is sort of a problem on production, but it seems to have gotten worse. Look at the left/right side of the in focus element.
Screen Shot 2022-11-02 at 2 16 37 PM
3. Hitting the fetch button triggers all validations to run immediately, rather than waiting for the fetch to complete.
4. The bottom of the scroll areas doesn't have anything to indicate where the scroll area ends. It just stops rendering the contents. These helpful panels should likely be boxed somehow.
Screen Shot 2022-11-02 at 2 19 14 PM
5. `Incidents reports matched by URL:` -> `Checking for duplicate report addresses...`. Also, I don't think the URL needs to be restated in the header of that check area. The value being queried doesn't need to be in the header for any of the duplication checks if the duplication check is right next to the form element.
6. The duplication check boxes are the only remaining hot point that jumps out to me in terms of design. I think one way to resolve this may be to eliminate the headers and attach the body of the duplication check boxes directly to the form element.
7. I don't think there should be both a toast and the green notification on the submit page -- one or the other. Also keep in mind that some people will be submitting the form multiple times in succession, so I don't think we should hide the form after submission.
8. Date downloaded should default to the current day.
9. I am not a fan of the gradient submit button. If that is consistent with our current design target, then let's go for it. Otherwise I prefer a less busy button pallet.
10. Nobody has used the CSV submit functionality in ages. I think we should remove this functionality since I doubt it works at this point.
Screen Shot 2022-11-02 at 2 36 55 PM

clari182:
@cesarvarela I can tackle 2., 8., 9. and 10. before merging.
The rest I think it's best to treat separately.

clari182:
@cesarvarela @smcgregor updates based on your comments:
> The following changes can be made after this is merged to staging/production.
> 
> 1. I believe Luna has some code targeting the old form where the incident can be given a title separately from the title of the report.

Added this in issue #1346

> 2. This is sort of a problem on production, but it seems to have gotten worse. Look at the left/right side of the in focus element.
>    Screen Shot 2022-11-02 at 2 16 37 PM

Fixed

> 3. Hitting the fetch button triggers all validations to run immediately, rather than waiting for the fetch to complete.

Added this in issue #1346

> 4. The bottom of the scroll areas doesn't have anything to indicate where the scroll area ends. It just stops rendering the contents. These helpful panels should likely be boxed somehow.
>    Screen Shot 2022-11-02 at 2 19 14 PM

Will be fixed here https://github.com/responsible-ai-collaborative/aiid/pull/1324

> 5. `Incidents reports matched by URL:` -> `Checking for duplicate report addresses...`. Also, I don't think the URL needs to be restated in the header of that check area. The value being queried doesn't need to be in the header for any of the duplication checks if the duplication check is right next to the form element.

Added this in issue #1346

> 6. The duplication check boxes are the only remaining hot point that jumps out to me in terms of design. I think one way to resolve this may be to eliminate the headers and attach the body of the duplication check boxes directly to the form element.

Added this in issue #1346

> 7. I don't think there should be both a toast and the green notification on the submit page -- one or the other. Also keep in mind that some people will be submitting the form multiple times in succession, so I don't think we should hide the form after submission.

Added this in issue #1346

> 8. Date downloaded should default to the current day.

Fixed

> 9. I am not a fan of the gradient submit button. If that is consistent with our current design target, then let's go for it. Otherwise I prefer a less busy button pallet.

Fixed

> 10. Nobody has used the CSV submit functionality in ages. I think we should remove this functionality since I doubt it works at this point.
>     Screen Shot 2022-11-02 at 2 36 55 PM

Fixed

PR #1082, opened 2022-09-12
Fix positioning of translation badge

cesarvarela:
Was broken on staging


clari182:
@cesarvarela code looks good however the TranslationBadge test is failing. I think it's because Bootstrap uses the `badge` class inside their badges while flowbite doesn't. They do use a `data-testid='flowbite-badge'` so maybe it can be used in the test.

Issue #956, opened 2022-08-17
Make a generic components for incident and report cards


lmcnulty:
We currently have a bunch of different places where we display incidents or reports in cards. These all have different styles:

<img src="https://user-images.githubusercontent.com/25443411/185217536-a7e8d71a-a0a4-4f74-a9ab-318ef37aacfc.png" height="300"/> <img src="https://user-images.githubusercontent.com/25443411/185217725-04e7d0b7-2308-42cd-8881-920d934c5a0a.png" height="300" /> <img src="https://user-images.githubusercontent.com/25443411/185218008-74155edd-3747-40d8-86e2-9065d40d434f.png" height="300"/>

![image](https://user-images.githubusercontent.com/25443411/185217599-accf910f-f441-4ac1-bc78-219a7f6382f0.png)

![image](https://user-images.githubusercontent.com/25443411/185218278-3bddd6d8-4492-47d1-8f24-f2ac7657298f.png)

We should convert these into a single generic component that can render appropriately in response to parameters. This should go along with moving to Tailwind.

smcgregor:
@clari182 I think you may be working this one in another way?

clari182:
@smcgregor Yes. Now we are using Flowbite, so any time we need a card we use their card's component. There are exceptions though, that can be improved, like the sections' cards on the citation page.
I can take this issue and work on improvements where possible!

Issue #582, opened 2022-04-29
Improve discover app cards


smcgregor:
> As a user, I'd like to be able to visit the "Show Details" page for an incident if I click anywhere on the incident card that is not a specific detail button (Authors, Submitters, etc.) Also, the list view mode cards do not have a link anywhere to visit the "Show Details" page.

From @wil-gerard

smcgregor:
I don't like making everything clickable on the discover app cards, but the other suggestion sounds good to me.

PR #1095, opened 2022-09-14
Add generic card component for incidents and reports 

lmcnulty:
Per #956.

image

image

image

image

image

image

## Breakage Attempts / Testing

- Citation Page
  - IncidentCard
    - Using the action buttons:
      - Edit the report
      - Flag the report
      - Open the submitters modal
      - Open the authors modal
      - Open the source and web archive link
    - From the header
      - Open the source article by clicking the title
      - Open the source article by clicking the image
      - Open the source article and web archive from the subtitle.
      - Add anchor to url to scroll to card
      - Resize window (slowly) from width 300 to 2000 

  - SimilarIncidents
    - Flag the report
    - Resize window (slowly) from width 300 to 2000 
    - Go to citation page by clicking title
    - Go to citation page by clicking image
    
- Tsne Page
  - Checked console
    - **Found** attempt to update unmounted component. Fixed.
  - Hover mouse over a bunch of points rapidly
    without waiting for the card contents to load.
  - Open citation page from image and title
  - Pan and zoom while the card is active
  - Pan with cursor over card
  - Click on the card while it's loading.
  - Resize window (slowly) from width 300 to 2000 

- Landing Page -> LatestIncidents
  - Go to citation page from
    - "Latest Incident Report" header
    - Title
    - Image
    - "Read More" button
  - Follow link in description
  - Resize window (slowly) from width 300 to 2000 

- Discover
  - Checked console
    - **Found** missing key for children. Fixed.
  - Search an incident
  - Filter a search
  - For each display mode...
    - Resize window (slowly) from width 300 to 2000 
      - **Found**: Action buttons overflowing 
        on medium screen size. Fixed.
    - Open link

    - With action buttons
        - Flag the report
        - Open the submitters modal
        - Open the authors modal
        - Open the source and web archive link
        - Filter by incident id
  - Zoom in and out randomly

- newsSearch
  - Resize window (slowly) from width 300 to 2000 
    - **Found**: Buttons wrap without vertical padding. Fixed.
  - Open archive and source links
  - Dismiss a candidate
  - Restore a candidate
  - Hover match tag for cosine similarity.


lmcnulty:
The action buttons seem to have disappeared on discover. This should also include a solution to #582.

lmcnulty:
For #582, it feels intuitively like I should press the action button containing the incident ID to see incident details, but that filters by ID instead. That's a useful feature, so I think a good solution is to keep that, but add a link to the incident page when a filter by ID is in place.

image

cesarvarela (commented):
I didn't look much into this PR yet, but this looks weird:
image

Also, please try setting this flag: https://github.com/gatsbyjs/gatsby/discussions/28138, and check that this PR doesn't introduce hydration errors (if you haven't already)

site/gatsby-site/src/components/cite/TsneVisualization.js
```diff
@@ -588,3 +565,13 @@ var Visualization = styled.div`
     z-index: 2;
   }
 `;
+
+import React, { useState, useEffect } from 'react';
```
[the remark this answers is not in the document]
lmcnulty: It's unorthodox, but I think it's the correct place for them. When you open a file, which do you think is more useful as the first thing you see: the signature of the default function, or `import React from 'react';` ? In low-level languages, you need to define everything before it's used for the benefit of the compiler, but in a high-level, dynamic language like JavaScript, you can order declarations for the benefit of the reader. 

I understand if this is too weird to be included, I just wanted to give it a try.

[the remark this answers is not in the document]
cesarvarela: I agree with the readability part, but to do this properly, we should do it on every file and update the linter settings, etc. 

Also, even if it works, technically, it is not valid javascript (on vanilla js it will fail)

I don't know if you use vscode, but I've found this as a consolation prize:

image

site/gatsby-site/i18n/locales/es/translation.json
```diff
@@ -154,5 +154,6 @@
...
   "About Taxonomies": "Acerca de las taxonomías",
   "Taxonomies are contributed to the AI Incident Database by persons and organizations working to structure the data and provide views into the data. Each taxonomy must be of sufficient quality and completeness to be included in the AI Incident Database, but the taxonomies are the responsibility of the persons and organizations contributing them.": "Las taxonomías son aportadas a la base de datos de incidentes de IA por personas y organizaciones que trabajan para estructurar los datos y proporcionar vistas de los datos. Cada taxonomía debe tener la calidad y la exhaustividad suficientes para incluirse en la base de datos de incidentes de IA, pero las taxonomías son responsabilidad de las personas y organizaciones que las aportan.",
-  "and": "y"
+  "and": "y",
+  "Viewing reports for incident #{{id}}. <3> View details on incident #{{id}}</3>.": "^Viewing reports for incident #{{id}}. <3>View details on incident #{{id}}</3>."
```
cesarvarela, line 158: Viendo reportes para el incidente #{{id}}. <3>Ver detalles sobre el incidente #{{id}}</3>.

site/gatsby-site/src/components/IncidentReportCard.js
```diff
@@ -0,0 +1,297 @@
+export default function IncidentReportCard(props) {
```
cesarvarela, line 1: What's the reason for using the same component to render an incident and a report?

My idea always has been to have `/components/reports/ReportCard` and `/components/incidents/IncidentCard`, (considering that rendering of each might diverge in the future)

site/gatsby-site/src/components/cite/TsneVisualization.js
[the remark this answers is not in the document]
lmcnulty: > I agree with the readability part, but to do this properly, we should do it on every file and update the linter settings, etc.

Yeah, you're right. It's probably too much effort to be worth it. I've put them back on top.

> Also, even if it works, technically, it is not valid javascript (on vanilla js it will fail)

Which part isn't valid? When I load this in the browser it displays "Hello, World!" even with the import on the bottom:

```html
<!doctype html>
<title>Title</title>
<script type="module" src="helloWorld.js"></script>
<script type="module" src="main.js"></script>
```

```js
// helloWorld.js
export default function helloWorld() {
  document.body.innerHTML = 'Hello, World!';
}
```
```js
// main.js
helloWorld();
import helloWorld from "./helloWorld.js";
```

> I don't know if you use vscode, but I've found this as a consolation prize:

I don't, but thanks for the idea! I came up with a vimrc one-liner to replicate it:

```vim
autocmd Filetype javascript :let @/ = "import" | silent! :execute "normal vN zf"
```

site/gatsby-site/src/components/IncidentReportCard.js
lmcnulty: There's more divergence in how we render reports in one place and the next than in how we render reports vs incidents. For instance, the list-mode discover cards representing reports to me look more like the horizontal incident card on the landing page than the the report cards on the citation page. For that reason, this component is very flexible in its rendering, but while providing a unified visual design. It can have a vertical or horizontal layout. It can have children as headers or footers. It can render different subsets of the data. And it can render data that represents a report or an incident – or something else. This PR uses it for the candidates, and it could probably be used for submissions too if we someday wanted that. Most of its present and future capabilities are independent of what data it's rendering. Let's say that we make it so that hovering over the date shows a tooltip with a more precise date. Well, that's something we'd probably want for both incidents and reports. Most foreseeable changes are like that. If something is applicable only to one or the other, like showing the taxonomy classifications, it seems likely to be better implemented as a child component than as an addition to the card component.

lmcnulty:
> I didn't look much into this PR yet, but this looks weird: image

Ok, I made the IncidentCards full-width again, as it currently. I think that looks bad because it makes the images super-large and the text too wide to read, so we will need to change the design of the page a bit to fix that, but it's out of scope for this PR.

> Also, please try setting this flag: [gatsbyjs/gatsby#28138](https://github.com/gatsbyjs/gatsby/discussions/28138), and check that this PR doesn't introduce hydration errors (if you haven't already)

I set that and it was failing to render server-side, giving me:

```
Failed to Server Render (SSR)
Error message:
React.Children.only expected to receive a single React element child.
```

I added this to IncidentReportCard, and it stopped giving me that error, so I think it's oke:

```js
  if (children) {
    if (!Array.isArray(children)) {
      children = React.Children.toArray([children]);
    } else {
      children = React.Children.toArray(children);
    }
  }
```

site/gatsby-site/src/components/cite/TsneVisualization.js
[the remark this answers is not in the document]
cesarvarela: Oh, so it hoists imports too... feels so wrong 😆

site/gatsby-site/src/components/IncidentReportCard.js
cesarvarela: Sorry I meditated a little bit on this. I get what you say; the line between incidents and reports is a bit blurry, but too many conditionals in the render function are a code smell; we are forcing an abstraction that isn't there. If you had to unit test this component, you'd end up with tons of tests, which makes me think this is not the best approach. 

For practical purposes, as we develop new features, it is better to have some separation (even at the cost of duplication/inconsistencies) when you have many people working on the same project. I get flashbacks from past projects with things breaking here and there because the same component is used everywhere :shudders:

lmcnulty: I've thought about this too, and you're right. I think the thing to do is separate it into different components based on the display logic rather than incidents vs reports vs candidates vs issues vs... Additionally, separate out child components for things like the subtitle that can be shared between the different display types.

PR #1099, opened 2022-09-15
Feature/flowbite readme

### LINKED ISSUE #1066: Update Readme for dev on how to use Flowbite
We should have the Readme have a Flowbite section to serve as a guide for developers that wish to contribute.

clari182:
- Resolves #1066


clari182:
@cesarvarela eventually we should get rid of that folder since we're going with the flowbite route. So I don't think the readme is necessary for the elements folder. In case we keep some components in there we should have a Readme for each one of them. I'll create an issue to clean that folder and add a Readme for the components we decide to keep.

PR #1115, opened 2022-09-19
Add is_incident_report flag

cesarvarela:
Adds a new field `is_incident_report`  to reports. which should be `true` for reports linked to an incident and `false`  otherwise.


site/gatsby-site/migrations/2022.09.19T20.06.43.is-incident.js
```diff
@@ -0,0 +1,14 @@
...
+ *
+ * @param {{context: {client: import('mongodb').MongoClient}}} context
+ */
+
+exports.up = async ({ context: { client } }) => {
+  await client.connect();
+
+  const reportsCollection = client.db(config.realm.production_db.db_name).collection('reports');
+
+  await reportsCollection.updateMany({}, { $set: { is_incident_report: true } });
```
pdcp1, line 13: We assume all existing reports are linked to an incident. Is that correct?

site/gatsby-site/migrations/2022.09.19T20.06.43.is-incident.js
```diff
@@ -0,0 +1,14 @@
+const config = require('../config');
+
+/**
+ *
+ * @param {{context: {client: import('mongodb').MongoClient}}} context
+ */
+
+exports.up = async ({ context: { client } }) => {
```
pdcp1, line 8: Does it make sense to have an `export.down` in this case?

pdcp1 (changes requested):

pdcp1:
@cesarvarela the integrity test is failing, please review.

image

site/gatsby-site/migrations/2022.09.19T20.06.43.is-incident.js
cesarvarela: yes

cesarvarela:
@pdcp1 yes the test will fail until the migration is ran, which only happens once merged to staging

site/gatsby-site/migrations/2022.09.19T20.06.43.is-incident.js
cesarvarela: I think it does not because it is not used by anything anyway

pdcp1: OK, that makes sense to me too.

pdcp1 (approved):

Issue #1112, opened 2022-09-19
New submission form issues


LKchemposer:
1. "Report description" should be "incident description". The distinction here is relevant because we need submitters to give info about the incident central to the report, not describe the report itself.
2. Description and alleged-parties fields should be hidden or automatically populated when an `incident id` is assigned, preferably hidden. Otherwise, would they create conflict with the existing filled-in fields if they get ingested in the database?
3. Description limit needs to be longer. It is currently set at 200 characters, many of existing descriptions are longer than that. Please set it at 500 at least.
4. Please give some cues about which fields are mandatory and which are not

cesarvarela:
Have you tried this one: https://staging-aiid.netlify.app/apps/submit ? Would like to hear your thoughts.

LKchemposer:
Love the staging submit form! A few things:
1. "Report description" should be "incident description," and higher character limit for it, per above
3. If we haven't rolled out tags (officially), not sure if we should hide it
4. The "press Enter to begin new item" mechanic for the alleged parties is not intuitive for me, although it is the same mechanic in the "incident table." Is it possible to make `,` as the button to separate item instead? This is also for the `authors` field since in many cases, authors are copy-pasted in with `,` and it would take a bit of effort to press Enter for each author to register them in.
5. I really like the dropdown with the "Tell us more" text. "Tell us more about the incident ... " can be good for clarity, but may not be necessary.

LKchemposer:
Also, a minor ask (optional): if you could put the `alleged deployer` field before/above `alleged developer` field, that would help me so much. I've recently run into issues where I put the wrong entities in, because I was in a flow state and I got used to the workflow of deployer-before-developer of the incident table.

Issue #943, opened 2022-08-15
Text input corners are rounded on left and angular on right


lmcnulty:
This was fixed on the landing page some time ago, but it still occurs on the submission form and possibly other places. This should probably be fixed as we migrate to tailwind.

<img src="https://user-images.githubusercontent.com/25443411/184684408-1d7003af-cb96-4c90-bf01-cfffbb89605e.png"/>

PR #1124, opened 2022-09-20
Submit form: Support commas in tag input, mark required fields

lmcnulty:
Implements all of the changes named in #1112.

- In tag inputs, a comma now begins a new tag, and a pasted comma-separated list get split into its components.
- Adds an asterisk before required fields.
- Hides the entity fields when .
- Sets the max description length to 500.

### DIFF
--- a/site/gatsby-site/src/components/forms/TextInputGroup.js
+++ b/site/gatsby-site/src/components/forms/TextInputGroup.js
@@ -36,8 +36,8 @@ const TextInputGroup = ({
             onChange={handleChange}
             onBlur={handleBlur}
             value={values[name] || ''}
-            className={touched[name] && errors[name] ? 'has-error' : null}
             isInvalid={errors[name] && touched[name]}
+            className={addOnComponent ? 'rounded-r-none' : 'rounded-md'}
             {...props}
           />
           {addOnComponent}
--- a/site/gatsby-site/src/components/forms/TextInputGroup.js
+++ b/site/gatsby-site/src/components/forms/TextInputGroup.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import { Form, InputGroup } from 'react-bootstrap';
 import Label from './Label';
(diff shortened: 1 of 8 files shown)


site/gatsby-site/src/components/forms/TextInputGroup.js
```diff
@@ -14,31 +14,40 @@ const TextInputGroup = ({
...
-      </InputGroup>
-    </Form.Group>
-  </div>
-);
+}) => {
+  const [optional, setOptional] = useState(true);
+
+  useEffect(async () => {
+    setOptional(await schema.fields[name].isValid(undefined));
+  }, []);
```
cesarvarela, line 25: Not sure I follow this 🤔

cesarvarela (commented):
One question before merging

site/gatsby-site/src/components/forms/TextInputGroup.js
lmcnulty: `schema` contains the full `yup` schema. We descend into the field matching the current input group, then we test if the field is optional according to the schema by seeing if `isValid` returns `true` given `undefined`. For some reason, `isValid` is asynchronous, so we need to wrap it in a `useEffect`.

cesarvarela: Oh my! There must be a better way to do this, but let's go with it for now. Forms need a rework anyway, including moving them to flow bite/tailwind.

cesarvarela:
@lmcnulty since you are at it:

`author` and `date_downloaded` show an asterisk but no invalid feedback message:

image

*This is by pressing `Submit`)

lmcnulty:
@cesarvarela I fixed the validation for tag inputs. The "Date Downloaded" field doesn't show an error because it defaults to the current day as of the last PR.



While I was there I also fixed the corner weirdness from #943.
<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

cesarvarela:
@lmcnulty please merge staging, tests should pass now

Issue #1098, opened 2022-09-15
Fix Facebook Login Setup Information


smcgregor:
`On Facebook Authentication settings, set the "Client ID" with the Facebook App Id. To get the Facebook App ID you should go to the [Facebook Developer Portal](https://developers.facebook.com/apps/), and check your app.`

The wording on this confused me for a while on which settings were being set where. I _think_ "Facebook Authentication settings" is referring to the secret in MongoDB, but I am still not sure.

Please re-work the documentation to be clearer and give more context. For instance, do I need to create a full Facebook organization and change the app mode from `Development` to `Live`?

smcgregor:
Also, please @cesarvarela and @pdcp1 make sure that all these manual deployment changes are highlighted in the deployment pull request. I did not expect to be working this tonight.

pdcp1:
@smcgregor You're right, in the documentation, I assume that the Facebook App is already created. Next time I will add more context and detailed instructions in the deployment PR.

To complete the FB integration we have to create another FB App. I created the FB Staging App using my personal Facebook account. If you give me your Facebook email user I can add you as an Administrator.

If you already have a Business account associated with AIID on Facebook, you should create an app under that Business.
If not, you can create an App using your personal Facebook user account.

In both cases, to create a new FB App you should go to https://developers.facebook.com/apps/create, click "Create App" > Select "Consumer" type

To help you in this process, please add me (pablocostapossamai@gmail.com) as an Administrator or Developer in the new FB App under the Roles FB app section (https://developers.facebook.com/apps/[appID]/roles/roles), so I can work on the setup on my side.

pdcp1:
@smcgregor It says that if we are not friends on Facebook, you have to use the username to add a user as Administrator. Please try with `pablo.costa`

smcgregor:
Done!

pdcp1:
@smcgregor Great! I made the changes on the Facebook App, now we have to add the FB App secret value as a Realm App secret.

1. First, copy the **FB app secret** value from https://developers.facebook.com/apps/1313947142744767/settings/basic/
2. Go to Production Atlas App Service > Values
3. "Create new value" with name `facebookAppSecret`, type `Secret` and value = the **FB app secret** copied on step 1

<img width="1156" alt="image" src="https://user-images.githubusercontent.com/6564809/190713843-cc5689f1-1ce4-48ec-8cf3-e9680bd29f16.png">

4. Click "Save Draft"
[...]

PR #1127, opened 2022-09-20
Update FB App ID on `providers.json`

pdcp1:
Related with #1098 

https://github.com/responsible-ai-collaborative/aiid/issues/1098#issuecomment-1252798523


cesarvarela:
So this will break FB login on staging but allow us to deploy to production and be happy ever after?

pdcp1:
> So this will break FB login on staging but allow us to deploy to production and be happy ever after?

@cesarvarela Yes, exactly

PR #1146, opened 2022-09-23
Restore tsne to sidebar, show all incidents

lmcnulty:
The embedding update workflow now runs on production, so the TSNE visualization can be made visible. This also fixes an issue where only the first 100 incidents were shown:


cesarvarela:
Does this run on staging too?

lmcnulty:
@cesarvarela Yes, it should. It's actually already rendering on staging: https://staging-aiid.netlify.app/summaries/spatial.

cesarvarela:
@lmcnulty  Yes, but I mean the automatic action

lmcnulty:
Oh, sorry. No, the workflow doesn't run on staging.

PR #1158, opened 2022-09-26
Update landing page carousel style

lmcnulty:
The current random reports carousel looks a bit bare on staging. This brings back the central label with a frosted glass background, plus it makes the query select from the first 300 rather than 50 incidents.

https://user-images.githubusercontent.com/25443411/192364620-9de0bf98-6afd-4a36-aece-8d0505387247.mp4

### DIFF
--- a/site/gatsby-site/src/components/landing/RandomIncidentsCarousel.js
+++ b/site/gatsby-site/src/components/landing/RandomIncidentsCarousel.js
@@ -1,80 +1,98 @@
-import React from 'react';
-import { StaticQuery, graphql, Link } from 'gatsby';
+import React, { useState } from 'react';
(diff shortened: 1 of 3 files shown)


cesarvarela:
I like the design improvements, but it is probably better to keep it static like it was before; I'm guessing that's why it was set to only the first 50 🤔

smcgregor:
I don't remember setting this to only query the first 50, so I would guess that restriction can be lifted.

This should not generate any dynamic queries with page load since it would significantly increase page render time. It can be random with every build cycle and keep our render snappy. This is more a concern for the landing page than any other part of the app, particularly since it affects both search ranking and bounce rate.

lmcnulty:
Fixed.

Issue #1132, opened 2022-09-21
Entity Logos


lmcnulty:
It would like nice if we could show the logos of the companies shown here:

![image](https://user-images.githubusercontent.com/25443411/191599692-8c50588a-d123-4e05-ba71-8846ab9d57a6.png)

Doing it properly would require a fair amount of work, since we'd need to develop a way to associate entities with logos, but if we just wanted to make the landing page look good in the short term we could special-case the top few to show hardcoded image paths.

LKchemposer:
I really like showing the logo idea, I agree that it'll look much nicer – i.e., more visual elements.

Could we pipe the logos from Wikipedia directly? Alternatively, how cumbersome would it be to keep a folder – online or otherwise – of the logos of the entities? In cases where the entities are not organizations, we should maybe have a placeholder.

smcgregor:
I like how iNaturalist pulls in content from Wikipedia for its species
pages. Ideally we could do something very similar for our entities pages in
the future and have a lot more details on the entities. That will take work
beyond this particular issue though.

PR #1159, opened 2022-09-26
Add hard-coded entity logos to landing page

lmcnulty:
The first step for #1132.

image


cesarvarela:
I like making the more button secondary, and adding the missing shadow 💯 

Before merging, will this break if Amazon enters the top 3?

lmcnulty:
@cesarvarela Thanks! I've added the logos for Amazon and YouTube.

PR #1173, opened 2022-09-28
Re-enable French

cesarvarela:
Sadly this build only passes if Cypress is not run, which signals once again a memory problem.

It can be seen working here: https://staging--cesarvarela-staging.netlify.app/


smcgregor:
Can you clock around the deploy preview under different languages? It seems to have slowed down considerably.

cesarvarela:
Oh yeah that'because I accidentally reverted this commit in this PR:

image

Of course, I'll drop it before merging.

It seems that the commit does what it says, though :D

PR #1181, opened 2022-09-28
Login infinite loading

clari182:
- Fixes infinite loading when going to the login page


cesarvarela:
I'm not 100% sure why this is necessary, but it seems to fix the issue and doesn't seem to break anything.

PR #1196, opened 2022-10-03
Disable button and add spinner on report edit

### LINKED ISSUE #1179: The edit page for reports gives no user feedback (e.g., spinner) while an update to the report is pending.
_Originally posted by @smcgregor in https://github.com/responsible-ai-collaborative/aiid/issues/1063#issuecomment-1260147053_

lmcnulty:
Resolves #1179.

image


cesarvarela:
@lmcnulty, I think it should be consistent with the other buttons that have the spinner inside (fetch report, promote report) or do you think we should change all submit buttons to work like this?

lmcnulty:
@cesarvarela Good point. I switched to a spinner inside the button, and I did so for all the form submission buttons where they were missing.

PR #821, opened 2022-07-18
Run tests on GitHub

pdcp1:
This PR adds a new Github Action called `Cypress e2e tests`

### ⚠️ Important note before merging
As soon this PR is merged to staging, a new Github action will be created and a set of Github secrets (ENV variables) should be created on the repo.
Please check the `README` file to get the Github secret variables.

Closes #667

PR #1200, opened 2022-10-03
Run tests on GitHub

cesarvarela:
Builds on #821, adds support for environments


cesarvarela:
For the time being, we'll use CircleCi to run tests. Maybe if more powerful boxes become available to non-enterprise users we could switch to Github.

Issue #1188, opened 2022-09-30
Spatial View – page description


LKchemposer:
Description should use as much plain and simple language as possible. The current description is precise, which is great!, but is less accessible to users who are not familiar with "natural language processing," "semantically similar," and what it means for text to be similar without some knowledge of NLP basics.

Suggestion for text:

The map of incidents (circles) below shows how similar in content the incidents are to each other. For example, the cluster in the top right corner are all incidents related to autonomous vehicles. We determine this similarity using a natural language processing system, which you can read more about here!

<img width="1124" alt="image" src="https://user-images.githubusercontent.com/39468345/193206017-9cf59457-5633-45f4-bc21-e1949823da1c.png">

Issue #1187, opened 2022-09-30
Spatial View – off-centered node labels


LKchemposer:
The node labels are off-centered and, for 3-digit numbers, really stand out (circled)

<img width="1124" alt="Screen Shot 2022-09-30 at 1 32 51 PM" src="https://user-images.githubusercontent.com/39468345/193205764-2c7e2638-d470-4666-aee8-fc54e14b6bf7.png">

Issue #1186, opened 2022-09-30
Spatial View – legends missing


LKchemposer:
User story: "I don't know what the color means. It'd help greatly with usability and readability if there are legends for the colors in the corner of the window. I only figured out what the colors correspond to only after hovering over the nodes, but I also have to do that every time I changed the option from the dropdown."

<img width="1124" alt="image" src="https://user-images.githubusercontent.com/39468345/193204768-7ea1fdd7-e588-49eb-95b8-ebbab16e15b5.png">

PR #1206, opened 2022-10-04
Query tsne statically, add legend

### LINKED ISSUE #1204: Spatial Visualization Loading
The spatial visualization currently queries all incident data on page load. This may be the correct behavior for the classifications, since not all of them are shown initially, but it would make more sense for the position of the data points to be queried at build-time.

lmcnulty:
Resolves #1204, #1188, #1187, and #1186. Additionally:

- makes it so that you can move the mouse to the incident card that pops up on hover without it disappearing
- makes it so that on mobile, tapping a plotpoint once shows the card instead of navigating to the incident page
- allows you to focus a given category by hovering it in the legend

image


site/gatsby-site/i18n/locales/es/translation.json
```diff
@@ -103,6 +103,7 @@
...
+  "The spatial view above shows each incident in the database as a plot point containing its incident ID number. Incidents are positioned so that those with similar report texts fall closer together. For example, incidents concerning autonomous vehicles form a tight cluster. We determine incident similarity using a natural language processing system, which you can read more about in <1>our blog post on its rollout</1>.":"^The spatial view above shows each incident in the database as a plot point containing its incident ID number. Incidents are positioned so that those with similar report texts fall closer together. For example, incidents concerning autonomous vehicles form a tight cluster. We determine incident similarity using a natural language processing system, which you can read more about in <1>our blog post on its rollout</1>.",
```
cesarvarela, line 106: Instead of having such a large key please use something like `tsneDescription`, to prevent translation files from getting too big.

cesarvarela:
```
La vista espacial anterior muestra cada incidente en la base de datos como un punto que contiene su número de incidente. Los incidentes se colocan de modo que aquellos con textos similares estén más cerca unos de otros. Por ejemplo, los incidentes relacionados con vehículos autónomos forman un grupo compacto. Determinamos la similitud de los incidentes mediante un sistema de procesamiento de lenguaje natural, sobre el que puede obtener más información en <1>nuestra publicación de blog sobre su implementación</1>.
```

cesarvarela (changes requested):

cesarvarela (approved):
Functional review:
- I expected the legend items to be toggleable, like in the taxonomy pie charts, which you can turn on/off
- using the trackpad to resize is not very fluid, d3 has a couple of helpers that make this smoother (you can test this in the location visualization), but this is something to fix in the future, if ever.

Let me know what you think about the toggleable legend, and I'll merge it afterward.

lmcnulty:
@cesarvarela The legend now allows toggling categories. Zooming is smoother with some points hidden, but it is worth thinking about how we can improve it in the general case.

https://user-images.githubusercontent.com/25443411/195873343-be2714ac-be63-42f8-8c51-b4a11008d94f.mp4

PR #1219, opened 2022-10-06
Run cypress post-build

cesarvarela:


cesarvarela:
Closing as netlify can't handle tests after a build is deployed; it either crashes or reaches Netlify's 35min max build time

PR #1222, opened 2022-10-07
Remove description truncation to 200 chars, fix validation

### LINKED ISSUE #1201: Editor Submission Form – `description` gets cut off
For some reason, the description text was cut off from how it was written in the Submitter Form. Is there a legacy character limit for this description?

image

lmcnulty:
Resolves #1201.

### DIFF
--- a/site/gatsby-site/src/components/forms/SubmitForm.js
+++ b/site/gatsby-site/src/components/forms/SubmitForm.js
@@ -104,14 +104,12 @@ const SubmitForm = () => {
     try {
       const date_submitted = format(new Date(), 'yyyy-MM-dd');
 
-      const description = values.description ? values.description : values.text;
-
       const submission = {
         ...values,
         incident_id: values.incident_id == '' ? 0 : values.incident_id,
         date_submitted,
         date_modified: date_submitted,
-        description: description.substring(0, 200),
+        description: values.description || values.text,
         authors: isString(values.authors) ? values.authors.split(',') : values.authors,
         submitters: values.submitters
           ? !isArray(values.submitters)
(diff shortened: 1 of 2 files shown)


site/gatsby-site/src/components/forms/SubmitForm.js
```diff
@@ -104,14 +104,12 @@ const SubmitForm = () => {
...
 
-      const description = values.description ? values.description : values.text;
-
       const submission = {
         ...values,
         incident_id: values.incident_id == '' ? 0 : values.incident_id,
         date_submitted,
         date_modified: date_submitted,
-        description: description.substring(0, 200),
+        description: values.description || values.text,
```
cesarvarela, line 112: I think this should be just `values.description`. Otherwise, when promoting to an incident, the `description` field will be huge.

site/gatsby-site/src/components/submissions/SubmissionForm.js
```diff
@@ -48,13 +48,10 @@ export const schema = yup.object().shape({
...
     .max(500, "*Titles can't be longer than 500 characters")
     .required('*Title is required'),
-  description: yup.string().when(['_id', 'incident_id'], {
-    is: (_id, incident_id) => _id !== undefined && (incident_id == '' || incident_id === undefined),
-    then: yup
-      .string()
-      .min(3, 'Description must have at least 3 characters')
-      .max(500, "Description can't be longer than 500 characters"),
-  }),
+  description: yup
```
cesarvarela, line 51: What's this change doing 🤔 ?

lmcnulty: The description field is no longer required, so we don't need to check that there's no incident id to place a length restriction. Until I made this change, the validation messages were not showing up for the description field – not sure why.

PR #1229, opened 2022-10-07
Fix edit report

### LINKED ISSUE #1228: Edit Report not working
When clicking Edit next to a report it redirects to a blank page

clari182:
- resolves #1228


cesarvarela (commented):
This seems to fix the symptoms but not the disease 🤔

PR #1274, opened 2022-10-17
Remove description and cloudinary_id from submission required

### LINKED ISSUE #1272: Cannot submit report
Pressing the Submit button:

... gives this box with no text, and no confirmation that the report is submitted

The report does not appear in the queue either

lmcnulty:
Resolves #1272. Bypassing staging because this is a major bug in production.

### DIFF
--- a/site/realm/data_sources/mongodb-atlas/aiidprod/submissions/schema.json
+++ b/site/realm/data_sources/mongodb-atlas/aiidprod/submissions/schema.json
@@ -130,13 +130,11 @@
     },
     "required": [
         "authors",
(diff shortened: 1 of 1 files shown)


smcgregor:
@cesarvarela can you review? This looks good to me, but I am surprised the cloudinary_id needs to be removed.

lmcnulty:
@smcgregor Yes, I'm not sure what could have changed with the cloudinary_id that would cause this to be necessary, but when making just the description optional I was getting the same error, but with cloudinary_id instead.

Issue #724, opened 2022-06-22
Demonstrations of AI System Failures


smcgregor:
The current criteria admits both harms and near harms, but incidents like https://incidentdatabase.ai/cite/12 don't fit -- it is a demonstration of a weakness present in a very common modeling approach. It does and will have many incidents in the real world demonstrating that it is "real," but these lab results don't pass the current criteria and the single report attached to this incident would be more appropriately attached to an incident that involves vector embeddings (maybe https://incidentdatabase.ai/cite/37 ). The difficult thing is that most of the time we will not know the exact system type being employed. So these demonstrations of incidents-in-waiting, AKA "recalls" in other industries, may not ever get attached to an incident even when it could be informative...

I think when we are ingesting these incident reports in the future, we should attach it to an incident that has already been reported (if it exists). If it does not exist, then we should create a new incident for it, and in the forthcoming "related incidents" feature @lmcnulty is working on we can cross reference the earlier demonstration of the system failure.

The options for instances when there is not a real world incident to attach to are,

a) Expand the criteria to include these "recall like" incidents (Demonstrated "weakness" maybe?)
b) Create a new first class report type (e.g., something like "demonstration") and keep it separate from the incidents
c) Only add these reports to real world harms or near harms as context
d) Don't add these reports to the DB at all
e) Create a catchall collection of reports that don't fit the criteria, but are still useful to index. This will allow for building the knowledge base while not getting in the way of operating a rigorously collected database

I am leaning towards (a) or (e) here.

CC @LKchemposer

smcgregor:
(Cross posted from Slack)

Hi @channel, something we have been discussing for quite some time in the dataset side of the AIID is the problem of incident definition. In particular, separating incidents where something has happened in the real world, and where someone demonstrates there is a problem with a technology absent of a harmed party. @Daniel Atherton referred to this in an email as an “incident in waiting.” In other settings, it could be referred to as a “recall” or a “vulnerability.” I believe there is consensus among those I have spoken to for including these non-incident reports in some form since they serve our change thesis and userbase. Further, they are often subsequently associated with real world realizations of those previously discovered issues.

<< Request >>

1. Please raise any objections you may have to expanding to a class of reports that are not incidents. Please presume that the additional class of reports will be easy to segment (via the user interface) from the original and ongoing collection of incidents and their reports.

[...]

PR #1285, opened 2022-10-18
Promote submissions to issues

cesarvarela:
From here https://github.com/responsible-ai-collaborative/aiid/issues/724#issuecomment-1228958315

1. [x] Submissions in the submit form should have a toggle that defaults to "incident" and additionally has the option of selecting "issue". If "issue" is selected, the form should not display the incident-associated fields.
2. [x] Make all reports in the report collection orphan-able, meaning that they can be associated with incidents in the incident collection, but that is optional
4. [x] Make all reports have a boolean value, `is_incident_report`, and require it to be set to `true` for all the current reports. It should be set to `false` if the report is not associated with any incident
5. [x] Add some tests to CI that confirms the requirement that `is_incident_report` is `true` whenever it is associated with one or more incidents, and `false` otherwise
7. [x] Add a toggle to the `discover` app that says `Display Issues`. This should default to false at the outset, but there will be a lot more downstream design here.
8. [x] The new type of reports in the discover app should link to a page that shows the complete report in isolation. This should just be the report card that is in the citation page at the moment -- we want to use a many of the components for `incidents` as possible here

This can be seen working  here: https://deploy-preview-34--cesarvarela-staging.netlify.app/


pdcp1:
@cesarvarela I'm testing the preview link https://deploy-preview-34--cesarvarela-staging.netlify.app and I see something weird on the Discover page https://deploy-preview-34--cesarvarela-staging.netlify.app/apps/discover/?display=details&is_incident_report=false&page=1

The info (title, description) displayed on the Discover search page for the existing 4 issues doesn't match with the Report details page. For example the first issue on this screenshot and the Issue details page https://deploy-preview-34--cesarvarela-staging.netlify.app/reports/2169

image

pdcp1:
We should test the sharing options on the Issue details page by adding another case to `socialShareButtons.cy.js`

site/gatsby-site/src/components/discover/REFINEMENT_LISTS.js
```diff
@@ -92,6 +92,15 @@ const REFINEMENT_LISTS = [
     faClasses: 'far fa-flag',
     type: 'refinement',
   },
+  {
+    attribute: 'is_incident_report',
+    placeholder: 'none',
+    label: 'dsad',
```
pdcp1, line 98: Change `dsad` label to something meaningful. It's also visible on mobile:

image

site/gatsby-site/src/components/discover/OptionsModal.js
```diff
@@ -56,6 +57,17 @@ function OptionsModal({ searchState, setSearchState }) {
...
             </Form.Label>
+
+            <ToggleDisplayIssues attribute="is_incident_report" id="display-issues" />
+            <Form.Label htmlFor="display-issues">
+              <Trans>Display Issues</Trans>
+            </Form.Label>
+
+            <ToggleOnlyIssues attribute="is_incident_report" id="only-issues" />
+            <Form.Label htmlFor="only-issues">
+              <Trans>Only Issues</Trans>
```
pdcp1, line 68: Missing translations for these two filters. Also the UI is broken on mobile:

image

site/gatsby-site/src/components/discover/hitTypes/Details.js
```diff
@@ -58,16 +66,14 @@ export default function Details({
...
-                navigate(path);
-              }}
+              onClick={() => navigate(detailsPath)}
             >
               <StyledLabel>
-                <Trans>Show Details on Incident #{{ id: item.incident_id }}</Trans>
+                {item.is_incident_report ? (
+                  <Trans>Show Details on Incident #{{ id: item.incident_id }}</Trans>
+                ) : (
+                  <Trans>Show Details on Issue #{{ id: item.report_number }}</Trans>
```
pdcp1, line 75: Missing translations for `Show Details on Issue ...`

pdcp1 (changes requested):

site/gatsby-site/cypress/e2e/issues.cy.js
```diff
@@ -0,0 +1,72 @@
...
+
+    cy.wait('@promoteSubmission')
+      .its('request.body.variables.input')
+      .then((input) => {
+        expect(input.incident_ids).to.deep.eq([]);
+        expect(input.submission_id).to.eq('5f9c3ebfd4896d392493f03c');
+        expect(input.is_incident_report).to.eq(false);
+      });
+
+    cy.contains('[data-cy="toast"]', 'Successfully promoted submission to Issue 182').should(
```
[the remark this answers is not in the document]
cesarvarela: They are not https://github.com/responsible-ai-collaborative/aiid/issues/724#issuecomment-1228958315

site/realm/functions/promoteSubmissionToReport.js
```diff
@@ -109,9 +118,15 @@ exports = async (input) => {
...
-  return incidents.find({incident_id: {$in: incident_ids }}).toArray();
+  return {
+    incident_ids,
+    report_number,
+  }
 };
+
+if (typeof module === 'object') {
+  module.exports = exports;
+}
```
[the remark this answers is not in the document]
cesarvarela: Yes

site/gatsby-site/src/components/discover/REFINEMENT_LISTS.js
cesarvarela: Yeah `dasd` it is because it shouldn't be visible, it's hidden now 👍

cesarvarela:
@pdcp1 
>  I see something weird on the Discover page 
This is because I was manually updating the Algolia index and the build got stale, it is up to date now and this doesn't seem to be happening anymore

> we should test the sharing options on the Issue details page by adding another case to socialShareButtons.cy.
Yeah, I missed that, but need to wait for an issue to be published first anyway

pdcp1:
@cesarvarela 
> Yeah, I missed that, but need to wait for an issue to be published first anyway

You're right, on the `socialShareButtons.cy.js` file we should add an additional case to test the Issue page sharing options, but it needs an existing Issue on the testing environment to work.
Please add an issue to don't forget to add these tests later when we have Issues created.

pdcp1 (approved):

cesarvarela:
@lmcnulty we need your input here: https://deploy-preview-34--cesarvarela-staging.netlify.app/apps/discover?display=details&is_incident_report=true&page=1

The toggles "Display Issues" and "Issues Only" add confusion to the already confusing  "First Report Only" toggle. Maybe you can suggest us a better way to express this.

"Display Issues" toggles an Algolia facet, making Incident Reports and Issues visible.

"Issues Only" toggles the same Algolia facet in a way that the Discover app displays Issues only (and not Incident Reports)

The Algolia facet is based on the new `is_incident_report` flag added to every Report.

The background on what an Issue is comes from this: https://github.com/responsible-ai-collaborative/aiid/issues/724

lmcnulty:
I would replace both switches with a dropdown with the following options:

- Incidents and Issues (Display Issues: **On**, Issues only: **Off**)
- Incidents only (Display Issues: **Off**, Issues only: **Off**)
- Issues only (Display Issues: **On**, Issues only: **On**)

smcgregor:
@lmcnulty that confuses me a bit... I think the problem is that we are switching between reports, the incidents they are collected into, and the issues that have no collection. How about buttons that read:

* `Display reports<checkmark highlighted> incidents<checkmark faded>` This is a replacement to the "1st report" toggle, and would eventually change to returning the incident titles and descriptions rather than the report titles and text.
* `Search incidents <?>` The question mark would display the incident short definition.
* `Search issues <?>` The question mark would display the issue short definition.

Supported configurations:

* Display reports, search incidents, search issues
* Display reports, search incidents, search issues <no>
* Display reports, search incidents <no>, search issues
* Display reports, search incidents <no>, search issues <no> -- This would return nothing
* Display incidents, search incidents <no>, search issues <no> -- This would return nothing
* Display incidents, search incidents, search issues <no>

Not supported:

* Display incidents, search incidents <no>, search issues <no>
* Display incidents, search incidents <no>, search issues

smcgregor:
@lmcnulty thanks! Updated.

lmcnulty:
> Display reports, search incidents `<no>`, search issues `<no>` -- This would return nothing
> Display incidents, search incidents `<no>`, search issues `<no>` -- This would return nothing

This seems like kind of a ["half the screen is gray"](https://www.joelonsoftware.com/2000/04/12/choices/#:~:text=trouble%20sending%20email.-,half%20the%20screen%20was%20grey,-%2C%20she%20said.) option. When possible, a UI should not expose unwanted states, such as a search for nothing. It seems like we've identified four states that a user might actually want:

```
Display reports,    [x] search incidents, [x] search issues
Display reports,    [x] search incidents, [ ] search issues
Display reports,    [ ] search incidents, [x] search issues
Display Incidents,  [x] search incidents, [ ] search issues
```

I still think a dropdown is the right UI element for this. The option labels would be something like:

- Incidents
- Incident and Issue Reports
- Incident Reports
- Issue Reports

A dropdown allows transitions between any of those four states in one action, 
whereas using three separate toggles will mean that some state transitions will take three actions, and most will take at least two. It also doesn't expose unwanted states.

Additionally, having toggles for both "display incidents" and "search incidents" visible at the same time is going to be extremely confusing. Users aren't thinking about the difference between "search" and "display," and even if they were, that's not the main distinction between those two options. The distinction is "Incidents (as opposed to issues)" vs "Incidents (as opposed to reports)". Rather than asking users to synthesize the abstract categories (e.g. "Incidents-not-Issues + Reports-not-Incidents"), I think it's clearer to present the concrete categories ("Incident Reports").

smcgregor:
👍 Sounds good to me.

cesarvarela:
@smcgregor @lmcnulty dropdown added lmk your thoughts

smcgregor:
@cesarvarela it looks like the build failed so there is no dropdown?

cesarvarela:
Yeah it failed, just fixed it: https://deploy-preview-34--cesarvarela-staging.netlify.app/

cesarvarela:
@smcgregor choose your flavor:

image


image


image

smcgregor:
I like the outline one.

cesarvarela:
@smcgregor oka to merge to staging?

smcgregor:
I clicked through to the report below on this page: https://deploy-preview-34--cesarvarela-staging.netlify.app/apps/discover?display=details&is_incident_report=false&page=1

And it 404s for these:

https://deploy-preview-34--cesarvarela-staging.netlify.app/reports/2171

https://deploy-preview-34--cesarvarela-staging.netlify.app/reports/2177

Disconnect between the build and the Algolia index?

cesarvarela:
Checked everything and added more tests, seems to be because I'm using the same env for other branches yes,.

cesarvarela:
@smcgregor I made the `incident_date` optional, and now I realize that I need to make the `alleged*` fields optional too, sounds good?

image

smcgregor:
Where is this form displayed? Those are all related to incidents rather than reports.

cesarvarela:
@smcgregor https://deploy-preview-34--cesarvarela-staging.netlify.app/apps/submitted

`mail@cesarvarela.com:1234568`

The form doesn't let you update the submission without those fields, so if you want to change the title for a submission you plan to submit as an issue you can't.

The "Add as Incident" button checks that those fields are set.

PR #1288, opened 2022-10-18
Prevent 0 from being counted as missing range value

### LINKED ISSUE #1273: Date Published filter dropdown is empty
In production:

lmcnulty:
Resolves #1273.

The date range filters could use some other improvements, but this causes them to at least appear.


cesarvarela:
Weird, I would like to know what changed 🤔

PR #1296, opened 2022-10-19
Add footer to LayoutHideSidebar, make body fill viewport

### LINKED ISSUE #1255: The discover app doesn't have the footer.
🥏

lmcnulty:
Resolves #1255.

## Breakage Attempts
- Checked that in each of `/apps/discover`, `/apps/reports/`, `/apps/classifications`, `/apps/incidents`, and `/entities`
  - The footer appears
  - It does not overlap with the other page content
  - There is appropriate whitespace between the footer and content
  - While the page content loads, the footer rests at the bottom of the viewport
    - **Found problem**: fixed by setting html, body, __gatsby, and focus wrapper height to 100% and setting content wrapper to flex-grow.
- Checked that changes to html and body styling don't result in:
  - double scrollbars 
  - horizontal scrolling
  - broken scrolling to anchors
  - unexpected wrapping at various screen sizes

### DIFF
--- a/site/gatsby-site/src/components/LayoutHideSidebar.js
+++ b/site/gatsby-site/src/components/LayoutHideSidebar.js
@@ -78,7 +78,7 @@ const LayoutHideSidebar = ({ children, location, menuCollapseCallback, className
   };
 
   return (
-    <>
+    <div className="flex flex-col h-full">
       <Header />
       <Wrapper className={`grow ${className}`}>
         <LeftSideBarWidth className={'hiddenMobile'} collapse={collapse}>
@@ -103,7 +103,7 @@ const LayoutHideSidebar = ({ children, location, menuCollapseCallback, className
         </Content>
       </Wrapper>
       <Footer />
-    </>
+    </div>
   );
 };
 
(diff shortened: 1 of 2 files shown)


site/gatsby-site/src/components/LayoutHideSidebar.js
```diff
@@ -79,7 +80,7 @@ const LayoutHideSidebar = ({ children, location, menuCollapseCallback, className
   return (
     <>
       <Header />
-      <Wrapper className={className}>
+      <Wrapper className={`grow ${className || ''}`}>
```
cesarvarela, line 83: nitpick: it looks like the `||` is not needed because, by default, `className` is empty already

site/gatsby-site/src/global.css
```diff
@@ -48,6 +48,21 @@ body {
...
 
+/********************************************************** 
+ * This allows us to set `flex-grow` on the main content  *
+ * so that the footer rests at the bottom of the viewport *
+ * if the content is not large enough to fill it.         *
+ **********************************************************/
+html, body, #___gatsby {
+  height: 100%;
+}
+#gatsby-focus-wrapper {
```
cesarvarela, line 59: Since looks like you only need `flex` for the `LayoutHidden` component, can you add that there?

Setting `flex` to every other page seems unnecessary, and it is kind of unexpected.

@clari182 is also working on layouts/menus might have something to add here

PR #1297, opened 2022-10-19
Use <a> instead of <Link> in September blog post

### LINKED ISSUE #1295: Monthly Roundup – September blog post issue
Embedded links for incidents in the table (left column) in "New Developments" section do not show up.

image

lmcnulty:
Resolves #1295.

## Breakage Attempts
- Checked that each link renders as a LocalizedLink with the React developer tools.
- Followed each link and confirmed that it led to the expected location.
- Tested that the back and forward buttons worked as expected.

### DIFF
--- a/site/gatsby-site/blog/incident-report-2022-september/index.es.mdx
+++ b/site/gatsby-site/blog/incident-report-2022-september/index.es.mdx
@@ -63,9 +63,9 @@ _Incidentes antiguos que tienen nuevos informes o actualizaciones._
   <tr>
     <td className="align-top border-1 border-gray-200 px-4 py-2">
       <strong>Incidente n.º 293: </strong>
-      <Link to="/cite/293">
+      <a href="/cite/293">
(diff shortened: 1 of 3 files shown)


cesarvarela (commented):
Should this also be done for the other languages?

smcgregor:
If this is about styling, shouldn't `<Link>` style the same as `<a>`?

lmcnulty:
> Should this also be done for the other languages?

Yes, good catch. I'll fix that shortly.

> If this is about styling, shouldn't `<Link>` style the same as `<a>`?

This fixes some of the links not rendering as links at all. I thought that `<Link>` was exported for use in MDX, but actually in MdxComponents.js we have:

```jsx
const Components = {
  …
  a: ({ href, ...props }) => <Link {...props} to={href} />,
  …
};
```

My understanding is that this causes `<a>`s to be rendered as `<Link>`s, and because `<Link>` was not defined in `Components`, it was just rendering as plain text.

PR #1298, opened 2022-10-20
Fix/quick action hover

### LINKED ISSUE #1133: Quick action button hovered color consistency
image

On hover, the quick action buttons should use the same shade of blue found elsewhere on the landing page. Either that the orange used in the sidebar. I'm leaning towards the latter because they're visible anywhere and I think we probably have more places with non-matching blues than oranges.

clari182:
- closes #1133 

I updated the hover of the quick access btns to orange (same as the left menu)


site/gatsby-site/src/components/discover/QuickAccess.js
```diff
@@ -22,7 +22,7 @@ export default function DiscoverQuickAccess({ className }) {
     <Wrapper className={`${className} p-4 md:bg-text-light-gray h-24`}>
       <LocalizedLink
         to="/apps/discover"
-        className="fs-5 tw-btn text-center shadow w-1/2 h-16 bg-white hover:bg-primary-blue hover:text-white"
+        className="fs-5 tw-btn text-center shadow w-1/2 h-16 bg-white hover:bg-light-orange hover:text-white"
```
lmcnulty, line 25: While it fits with the other sidebar items, I think orange looks a bit weird for buttons, especially since we don't have any other orange buttons. But it's a matter of opinion. I would suggest trying it with the blue we get with the Flowbite buttons, and if you like that less, I'm fine with orange too.

lmcnulty (approved):

site/gatsby-site/src/components/discover/QuickAccess.js
cesarvarela: I agree with Luna, tell us what you think @clari182

clari182:
Fixed based on comments @lmcnulty @cesarvarela . Let me know if it looks better now

PR #1313, opened 2022-10-25
Feature/collapsable menu

### LINKED ISSUE #1059: Make main menu collapsible
I think instead of having the weird floating blue menu button, we should make the menu collapsible, for example: 

expanded:
image

collapsed:
image

This will also prevent the menu from totally going away when you click on the Discover button.

clari182:
- closes #1059


cesarvarela (commented):
Doesn't seem like you need a context since you are using the state in only one place or am I missing something 🤔

site/gatsby-site/src/components/LayoutHideSidebar.js
```diff
@@ -1,11 +1,10 @@
-import React, { useState } from 'react';
```
cesarvarela, line 1: Why not delete this file altogether?

clari182:
Thanks @lmcnulty !! I'll work on your comments

clari182:
@lmcnulty ready for another review. I went with the orange arrow at the top that replaces the discover and submit buttons. However I didn't place it between the two when the menu is open because it felt a little strange IMO. Let me know what you think. I also tackled the rest of your comments.

lmcnulty:
> However I didn't place it between the two when the menu is open because it felt a little strange IMO

I get what you mean, but I find the collapse button floating outside of the sidebar a bit weird as well. What if we made a notch for it to go in? I find This would have been hard to explain so I made a [PR](https://github.com/responsible-ai-collaborative/aiid/pull/1366) for it, although it seems not to be building right now.



I also made the sidebar no longer sticky, because the bottom was still inaccessible when the screen wasn't tall enough to fit it. Other than that everything looked good.

smcgregor:
Hot take: I don't think I would connect the open/close button to opening and closing the side menu if it is in the middle of the Discover/Submit buttons.

clari182:
Got it! Thanks @lmcnulty I'll include those changes on my PR

lmcnulty:
> Hot take: I don't think I would connect the open/close button to opening and closing the side menu if it is in the middle of the Discover/Submit buttons.

I agree that it's not immediately obvious that the button in the middle will collapse the sidebar, but I think that it adds more visual interest and that the functionality will become apparent once the arrow flips around when the user visits a page that starts out collapsed. And since being able to manually collapse the sidebar is not necessary to use any functionality of the site, I think that's ok.

Some other options:

1. Put it inset in the bottom corner:
   image

2. Put it inset in the middle, and maker smaller / less prominent:
   image


3. Set `cursor: col-resize` when the cursor approaches the edge of the sidebar and allow manually resizing it, like in Slack:

   image

   Maybe this could also include a narrow handle to communicate that the edge can be grabbed.

cesarvarela (commented):
Not sure if this is ready for review but received a notification so:

Looks like when collapsed there is an option for Discover but not on for Submit.

This seems broken:

image


Maybe use overflow hidden when expanding the menu so you don't see all the cramped-up letters while the expand animation runs?

clari182:
@lmcnulty I went with a solution that I think fits with what you were pointing out and what Sean was.
The collapse button is an orange arrow that replaces the Discover/Submit btns when collapsed. Ready for another round of review!
@cesarvarela can you shoot a video of what is happening? I tried in Safari, Chrome and Firefox to reproduce it but no luck. Thanks!

cesarvarela:
@clari182 happens when you click on the discover menu item (not the big button)

lmcnulty:
Maybe we could put the button at the bottom and make the icons slide up. E.g.


https://user-images.githubusercontent.com/25443411/202039347-4071e438-540d-4395-bc94-aedbbb652dc9.mp4

Another draft pr: #1387.

clari182:
@lmcnulty I moved the arrow down, let me know what you think.
@cesarvarela fixed the discover issue

lmcnulty:
We still need to do something about the case where the screen is not large enough to fit all the sidebar icons. Currently they're not always possible to reach, and with the collapse arrow at the bottom, it sometimes overlaps.

https://user-images.githubusercontent.com/25443411/202521253-1f57c7a3-c10a-4705-b133-e0e3311beea1.mp4

I'm going to try to come up with a solution on my branch, if that's ok.

clari182:
@lmcnulty sounds good Luna. Maybe we can remove the sticky for smaller screen sizes. I think you had already worked on this previously in another branch. I'll put this on hold if you want to take the small screen issue and then incorporate it into this PR.

cesarvarela (changes requested):
The padding on the left is smaller than on the right. Not sure if it is intentional or not:

image


The tooltip text getting wrapped look funny:

image

When loading the discover app, the spinner, and some UI components get shifted to the left and then to the center after loading:

image

clari182:
@cesarvarela I fixed the 3 changes you requested

cesarvarela:
What's the "official" one? this or #1387

cesarvarela (commented):
What's the "official" one? this or https://github.com/responsible-ai-collaborative/aiid/pull/1387

clari182:
@cesarvarela I think at this point it's best to continue with #1387
Looks better IMO and Luna continued based of my initial work

PR #1317, opened 2022-10-25
News Digest Update

lmcnulty:
[Deploy preview with data](https://deploy-preview-13--lmcnulty-staging.netlify.app/apps/newsdigest/).

- Goes with https://github.com/responsible-ai-collaborative/nlp-monitoring/pull/1.
- Updates title and description
- Changes url to /apps/newsdigest and adds redirect. I'm not sure if the redirect works. 
- Adds harm-related keywords in red
- Adds entities in gray
- Sorts based on age, cosine similarity, and number of keywords
- If the cosine similarity to the mean embedding of reports with a given classification is substantially greater than the cosine similarity of all report embeddings, it displays that classification below the tags. This never seems to occur, however.

localhost_8000_apps_newsdigest_


site/gatsby-site/tailwind.config.js
```diff
@@ -58,6 +58,7 @@ module.exports = {
     extend: {
       screens: {
         '1300px': { min: '1300px' },
+        '3xl': { min: '1920px' },
```
cesarvarela, line 61: What's this for? Looks like it may have sitewide implications

cesarvarela (commented):
It bothers me a bit that the skeleton has nothing to do with what is shown after the page loads. 

Maybe remove the avatar part removing it?

site/gatsby-site/tailwind.config.js
lmcnulty: The largest [breakpoint](https://tailwindcss.com/docs/responsive-design) provided by Tailwind is 2xl at 1536px. If I set the grid of cards to 3 columns at 1536px, it looks too cramped. Whereas if just leave it at 2, the cards get too wide on a larger screen. 3xl follows the pattern, and 1920px seems like a natural place to set it. It's not overriding a default, so it won't have effects anywhere except here. I'd rather have it set here rather than a one-off because if we need a larger breakpoint that 1535px again we should use the same one.

PR #1320, opened 2022-10-26
Deploy to Production

cesarvarela:
Key updates:

- new submit form
- new incident subscriptions
- entities collection (not "visible" but might break things) 

Issues 

~- migrations will trigger tons of incident updates (deleting and pushing the trigger before and after the migration might work)~

This PR deletes the MongoDB `incidents` trigger, so the included migration doesn't trigger spurious incident updates.

After it is merged and the build finishes https://github.com/responsible-ai-collaborative/aiid/pull/1357 should be merged.

(we are looking for a better workaround, but in the meantime...)


cesarvarela:
@smcgregor this is ready for review

cesarvarela:
Hopefully, it is just flakiness, but I'll take a look.

PR #1322, opened 2022-10-26
Add incident_title and editors to submissions for editors

### LINKED ISSUE #1266: Editor Submission Form – Add `Editors` & `Incident Title` fields
Legacy requests from #1111

`Editors` and `Incident Title` fields are available in the Table View Edit form:

image

But they are not available in the Submission form *for a new incident* in the Tell us more:

image

These fields should _only_ be available to editors, not submitters, and for new incidents. This saves editor time spent on manually changing these entries for each incident.

lmcnulty:
Resolves #1266.

- I think ultimately these fields should go outside of "Tell us more" along with any other fields that are mandatory. However, that change belongs in another PR.
  - Maybe for editors we should just not collapse the tell us more section at all ?

## Testing / Breakage Attempts

- As editor, make and promote submission with incident ID
  - Can't add incident title or editors
  - Check MongoDB that new report content is correct

- As editor, make and promote submission without incident ID
  - Don't add incident title or editors
    or other optional information on submission
  - Blocked from promoting until more info is added
    - Form validation requires title. Add it. 
    - Doesn't require editor - leave it blank.
  - Check MongoDB that new incident and report content is correct

- As editor, make and promote submission without incident ID
  - Add everything *except* the incident title
  - **Found bug**: When only the incident title is missing,
    it allows promotion, even though it requires it in the form.
    Fixed by adding `!submission.incident_title` to missing data check.
  - Check MongoDB that new incident and report content is correct

- When logged out, make submission without incident ID
  - **Found bug**: Displays editors, but not title, to logged-out users
  
- When logged out, make submission with incident ID

- As editor, promote both of the above

### DIFF
--- a/site/gatsby-site/cypress/e2e/submit.cy.js
+++ b/site/gatsby-site/cypress/e2e/submit.cy.js
@@ -2,6 +2,8 @@ import parseNews from '../fixtures/api/parseNews.json';
 
 import semanticallyRelated from '../fixtures/api/semanticallyRelated.json';
 
+import { maybeIt } from '../support/utils';
+
 describe('The Submit form', () => {
   const url = '/apps/submit';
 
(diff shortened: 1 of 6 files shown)


cesarvarela (changes requested):
Please add/update tests as appropiate

lmcnulty:
@cesarvarela I've added a `maybeIt(…)` to test submissions with the new fields. Since they only appear for editors, I don't think the other tests should be affected.

PR #1324, opened 2022-10-27
Feature/similar indicents cards

### LINKED ISSUE #1251: Redesign similar incidents
Redesign the similar incident section to use badges

### Current
image

### Option 1: 
image

### Option 2:
image

@smcgregor and @lmcnulty please let me know what option you like best and share your thoughts on the matter.

Thanks!

clari182:
- closes #1251


lmcnulty (changes requested):
Looks nice! A few design comments:

- There should be padding above and below "incidents reports matched by published date" (and the spinner that appears before it)
  image

- It looks weird to have two buttons in primary blue right next to each other. I think that "not sure" should be light gray instead.

  image

- I would put the incident number and title on one line, or if that turns out not to look good change the top line to `Incident #45` to reduce the empty space. I do like that the number is bigger, though.

  image

- The button text wraps when the card doesn't have enough space

  image

- When the cards wrap, it seems like the one on the next line should either be the full width, or the same width as the other cards

  
  image

clari182:
@lmcnulty ready for another review round! I fixed styling based on your comments

lmcnulty:
Awesome! A couple more things:

image

- The button group and the "Use Id" button should be vertically aligned
- The bottom padding should be the same as the left padding

- Black still looks kind of weird for the "not sure" option. I would prefer light gray, like what you get when you hover over the light buttons. However it looks like that would require us to customize the component.

  The yellow `color="warning"` might work as an intermediate between red and green. 

  image

  That still seems a bit strong though. Perhaps the best option would be to have all the buttons start with color="light" initially, then set the "not sure" to yellow if it's actually pressed?

  image

lmcnulty (commented):
Thanks! Visually the only issue I see is that there should be space between the spinner and the label:

image

It also looks like the translations are missing?

image

clari182:
@cesarvarela conflicts fixed

cesarvarela:
@clari182 another one!

cesarvarela (changes requested):
Big changes, and not one test was added/updated.

clari182:
@cesarvarela tests updated!

clari182:
@cesarvarela test added!

cesarvarela:
@clari182 please merge staging to fix conflicts and get tests to run

clari182:
@cesarvarela done!

cesarvarela (commented):
image

conflicts!

smcgregor:
Merging this after the next deployment.

Issue #1064, opened 2022-09-07
Add Editor Standard Pages


smcgregor:
I have been discussing these with @LKchemposer , who as brought together an excellent starting guide for the editor standards.

https://docs.google.com/document/d/1JbRC5IMehpcH-Ivp8dZojUPam3ToXKRxGL3ycUIrNxo/edit#heading=h.3bcojmxvmbqm

smcgregor:
@lmcnulty thanks for typesetting, my initial take is that this is difficult to read because our styles are not designed for this sort of documentation content. I am also hesitant to spend too much time on this beyond the first take typesetting.

smcgregor:
(sorry to accidentally bring the conversation to the issue instead of the PR, we were also combing over the contents one more time today since the `issue` functionality is about to land)

Issue #1326, opened 2022-10-28
List decorators separated from list content when `<li>` contains `<p>`


lmcnulty:
On production: https://incidentdatabase.ai/blog/representation-and-imagination

<img src="https://user-images.githubusercontent.com/25443411/198668635-de5c89aa-8942-41a2-89f7-c5c2dd13a5f8.png" width="500"/>

This issue goes away when we remove `list-style-position: inside` added in https://github.com/responsible-ai-collaborative/aiid/pull/1226. However, it would re-introduce the problem from issue #1182. I'd prefer to handle that with `ol, ul { padding-left: 2em }` (which is the default browser style, but overridden to by Tailwind). It would fix this problem in a simple way, and it's also more readable, in my opinion:

<table>

  <tr><th><code>li { list-style-position: inside }</code></th><th><code>ol { padding-left: 2em }</code></th>
<tr>
<td>List items don't really stand out from each other</td>
<td>The outside marker emphasizes where one item starts and the other ends</td>
</tr>
  <tr>
    <td><img src="https://user-images.githubusercontent.com/25443411/198673784-443dc3c1-4815-4e75-a09d-ed44a42d9d2c.png"/></td>
    <td><img src="https://user-images.githubusercontent.com/25443411/198673632-5f413a8a-0744-4e77-9122-021923a611d9.png"/></td>
  </tr>

</table>

What do you think, @clari182 ?

clari182:
Sounds good @lmcnulty ! Assigning to myself to work on the fix

cesarvarela:
This line is doing things:

https://github.com/responsible-ai-collaborative/aiid/blob/master/site/gatsby-site/src/tailwind.css#L60

PR #1327, opened 2022-10-28
Add editor's guide

lmcnulty:
Per #1064. [Deploy Preview](https://deploy-preview-1327--staging-aiid.netlify.app/editors-guide)

I've made some [content edits](https://github.com/responsible-ai-collaborative/aiid/pull/1327/commits/a8c7106e1f433d9915f5da1fcc80751e877636ba):
- Added links to https://en.wikipedia.org/wiki/Title_case and https://daringfireball.net/projects/markdown/
- Added explanation of confirmation bias
- Reworded explanation of cultural bias
- Fixed ambiguous pronouns
- Reworded reference to "non-technical" users. [I've written before about why I dislike this term](https://lmcnulty.me/words/non-technical/), but briefly it's because:
  - it conflates different skills (someone may know a lot about system programming but very little about AI).
  - it's almost always used in a way that ignores technical skills unrelated to computing.
  - it discourages people who are taught to think of themselves as "non-technical" from learning more.


smcgregor:
@LKchemposer can you take first review?

site/gatsby-site/content/editors-guide/index.es.mdx
```diff
@@ -0,0 +1,474 @@
...
+## Definitions
+
+The AI Incident Database contains records of AI “incidents” and “issues”.
+
+- **AI incident**: 
+  an alleged harm or near harm event to people, property, or the environment 
+  where an AI system is implicated.
+
+- **AI issue**: 
+  an alleged harm or near harm by an Al system
```
smcgregor, line 17: (not doing a review, but I just noticed this says "eh el", not "eh aye")

site/gatsby-site/src/global.css
```diff
@@ -53,8 +53,15 @@ a {
   text-decoration: none !important;
 }
 
-li {
-  list-style-position:inside;
+ol, ul {
```
cesarvarela, line 56: This changes all the lists on the site, I would say you should create custom components and add them here: https://github.com/responsible-ai-collaborative/aiid/blob/master/site/gatsby-site/src/components/ui/MdxComponents.js

cesarvarela: Or maybe we should be using this https://flowbite.com/docs/components/typography/ instead?

cesarvarela (changes requested):

site/gatsby-site/src/global.css
lmcnulty: This a solution to the general problem in #1326, included in this PR for now because the page is unreadable without a fix. It restores the default style that got overridden with the introduction of Tailwind. There is no reason for this to be a custom component – it's a completely vanilla list item.

This particular page has heavily nested lists, which Flowbite Typography [might not support and isn't designed for](https://flowbite.com/plugins/typography/#:~:text=i'm%20not%20sure%20if%20we'll%20bother%20styling%20more%20than%20two%20levels%20deep.).

cesarvarela: We need to let tailwind/flowbite overwrite vanilla styles and work from there. This way, any developer that copies/pastes elements from Flowbite's docs will have them working out of the box. If we undo these overrides, this stops working.

I mentioned the custom component because it was the fastest way to deal with this, but ideally, I think we should be using Flowbite's typography defaults here, too; it should be just a matter of setting the `prose` class at the top of our markdown-based content, and that should be it.

```
<article class="prose lg:prose-xl">
  {{ markdown }}
</article>
```  

Are you sure it will look too bad?

cesarvarela: To add more context, if we undo global styles, we defeat the purpose of this: https://tailwindcss.com/docs/preflight

cesarvarela (changes requested):
Missing translations

lmcnulty:
[Sean says](https://github.com/responsible-ai-collaborative/aiid/issues/1456#issuecomment-1337960333) we should hold on translating this for now.

> The guide is likely to develop fairly quickly over the next few months, so let's hold this one until it begins to stabilize. We should link to the English version from the translated pages.

cesarvarela (approved):

PR #1328, opened 2022-10-31
Update list style to avoid hardcoded revert style

### LINKED ISSUE #1326: List decorators separated from list content when `<li>` contains `<p>`

clari182:
- closes #1326


lmcnulty (approved):
This works! I'm not sure how it works, though. :thinking: What does `list-style: auto` do? I don't see it in the [MDN docs for list-style](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style).

cesarvarela (commented):
This change made all lists numbered, it seems:

image

clari182:
@lmcnulty it was working because it was overriding the `list-style: none` with auto but it was not correct, since auto is nothing apparently. At first I thought it was working because it was resetting the list styles to default but it wasn't that.
Apparently, tailwind has a `list-style: none` set in their default theme which was messing up the mdx files.
I set a `ul, ol { list-style: revert; }` which should revert that, but the catch is if we create a `<li>` we need to remember to set the class `list-none` to it, otherwise it will have a bullet.
I checked the rest of the site and everything seems to work as expected.
But let me know if you spot something I didn't!

lmcnulty (changes requested):
This re-introduces the problem solved by https://github.com/responsible-ai-collaborative/aiid/pull/1226.

image

Lists also need to have padding, like `ol, ul { padding-left: 2em }`, as discussed in #1326. You might also be able to do something like `ol, ul { padding-left: revert }` to undo Tailwind's 0-padding rule.

clari182:
@lmcnulty done!

lmcnulty (approved):

site/gatsby-site/src/tailwind.css
```diff
@@ -125,6 +120,11 @@
         @apply text-white bg-transparent relative order-1 ml-auto opacity-0 w-0 h-0 overflow-hidden focus:opacity-100 focus:py-[0ch] focus:px-[1ch] focus:w-[unset] focus:h-[unset]  767px:text-xs;
     }
 
+    ul, ol {
+        list-style: revert;
+        padding-left: 2em;
```
cesarvarela, line 125: I mentioned this in another pr, but we can't override global styles like this, it kind of defeats the purpose of using Tailwind. It already has a CSS reset https://tailwindcss.com/docs/preflight.

clari182: The issue is that tailwind has a `list-style: none` set in their default base settings. So for markdowns in the `mdx` files, it will never show the right list style.
If we choose to use tailwind's `list-style: none` we accept the fact that all markdown list will have no style (no bullets or numbers). 

For example:
image
is defined with markdown as with a number list but they will not be displayed as such. My latest commit has that change, let me know what you think.

cesarvarela: Yeah, I think we should use the plugin typography there; it feels like we are reinventing the wheel.

clari182: Done! I included the typography plugin for blog posts now. Let me know what you think

lmcnulty:
If we're using Tailwind Typography for this, we need to apply it in more places than just the blog posts.

That includes the MDX pages:

image

As well as the citation pages:

image

lmcnulty (approved):

site/gatsby-site/src/tailwind.css
cesarvarela: Nice, seems to be an improvement 🥳

PR #1331, opened 2022-11-01
Change pending notification process

pdcp1:
This PR changes the way that we process pending notifications.

### Context
Before this change, we sent email notifications on New Incidents as soon as the Incident was inserted into the DB. The problem with that approach is that the Incident page is not yet generated.

### Solution
Send email notifications after the build is done (`onPostBuild` gatsby event).

### Testing link
https://deploy-preview-1331--staging-aiid.netlify.app/


site/realm/data_sources/mongodb-atlas/aiidprod/notifications/rules.json
```diff
@@ -0,0 +1,37 @@
+{
+    "collection": "notifications",
+    "database": "aiidprod",
```
cesarvarela, line 3: I would move this to the `customData` database because it is not related to the actual "data" of the database. Also, it might get us in trouble if we do a restore of the `aiidprod` database.

cesarvarela (commented):
I wonder if the `postBuild` event is late enough for the new Incident page to be published since after the build completes, Netlify takes a few minutes to upload and publish the site.

(I still think that it is a good compromise to avoid doing something more complicated)

site/realm/data_sources/mongodb-atlas/aiidprod/notifications/rules.json
pdcp1: Done! `notifications` collection moved to `customData` DB

PR #1353, opened 2022-11-08
update reports count on landing

smcgregor:
We have more than 2k now!


smcgregor:
@cesarvarela this one should be quick and easy.

PR #1368, opened 2022-11-10
Fix/Form error handling

clari182:
On form error, prevents form from going to the last step
[Deploy Preview #1368](https://deploy-preview-1368--staging-aiid.netlify.app/)

### DIFF
--- a/site/gatsby-site/src/components/submissions/SubmissionWizard.js
+++ b/site/gatsby-site/src/components/submissions/SubmissionWizard.js
@@ -15,15 +15,25 @@ const SubmissionWizard = ({ submitForm, initialValues }) => {
 
   const [steps, setSteps] = useState([]);
 
+  const [submissionFailed, setSubmissionFailed] = useState(false);
+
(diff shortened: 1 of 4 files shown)


site/gatsby-site/src/components/submissions/SubmissionWizard.js
```diff
@@ -15,15 +15,25 @@ const SubmissionWizard = ({ submitForm, initialValues }) => {
...
   const handleNextStep = async (newData, final = false) => {
+    setSubmissionFailed(false);
     setData((prev) => ({ ...prev, ...newData }));
 
     if (final) {
-      await submitForm({ ...data, ...newData });
+      try {
+        await submitForm({ ...data, ...newData });
+      } catch (error) {
+        setTimeout(() => {
```
cesarvarela, line 32: This will result in a race condition. Why is it necessary? Does `0` work? It is a bit better.

clari182: 0 works!

Issue #1346, opened 2022-11-04
Future Form improvements


clari182:
- [ ] Hitting the fetch button triggers all validations to run immediately, rather than waiting for the fetch to complete.
- [ ] `Incidents reports matched by URL:` -> `Checking for duplicate report addresses...`. Also, I don't think the URL needs to be restated in the header of that check area. The value being queried doesn't need to be in the header for any of the duplication checks if the duplication check is right next to the form element.
- [ ] The duplication check boxes are the only remaining hot point that jumps out to me in terms of design. I think one way to resolve this may be to eliminate the headers and attach the body of the duplication check boxes directly to the form element.
- [ ]  I don't think there should be both a toast and the green notification on the submit page -- one or the other. Also keep in mind that some people will be submitting the form multiple times in succession, so I don't think we should hide the form after submission.

cesarvarela:
![image](https://user-images.githubusercontent.com/1172479/200399035-18ab1cab-8844-44ee-9d44-20d13826e2ce.png)

The one-hour thing is not accurate, submissions appear right away.

cesarvarela:
- [ ] submitting only the first page is not working

smcgregor:
@clari182 are you working the `submitting only the first page is not working` issue? This is a blocking issue for the deployment.

clari182:
Yes @smcgregor it's ready for review here https://github.com/responsible-ai-collaborative/aiid/pull/1364

smcgregor:
- [ ] When adding a report for an existing incident, the `description` field is still visible. It should be hidden if the incident already exists.

PR #1369, opened 2022-11-10
Submit Form / don't show errors when clicking fetch info

clari182:
- resolves Hitting the fetch button triggers all validations to run immediately, rather than waiting for the fetch to complete. in #1346


smcgregor:
👍

smcgregor (approved):

PR #1374, opened 2022-11-12
Fix cypress tests

cesarvarela:
Fixes all broken tests:

image

Even if they are flaky on CI, on a local machine, they run very consistently; let's not forget about them <3

Soon we'll be running them on every PR as before.

Interesting read:

https://testing.googleblog.com/2016/05/flaky-tests-at-google-and-how-we.html

> Almost 16% of our tests have some level of flakiness associated with them! This is a staggering number; it means that more than 1 in 7 of the tests written by our world-class engineers occasionally fail in a way not caused by changes to the code or tests.


cesarvarela:
FYI: @pdcp1 @clari182 @lmcnulty

pdcp1:
@cesarvarela This test fails on my local environment using the DB dump from Staging.

```
1) Cite pages
       Should show proper entities card text:

      Timed out retrying after 30000ms
      + expected - actual

      -'Alleged: Tesla developed an AI system deployed by Tesla and Motorist, which harmed motorists.'
      +'Alleged: Tesla developed an AI system deployed by Tesla and Motorist, which harmed Motorists.'
```

This is because one of the incidents has `motorists` instead of `Motorists` before the update migration, so it creates an entity with the following:
image

The solution is to update `motorists` to `Motorists` entity name on Staging DB. I see that it was already updated on Prod.

cesarvarela:
Fixed the name on staging and as you mention it is already fixed on production

PR #1399, opened 2022-11-17
Display response badge on incident response reports

### LINKED ISSUE #1395: Incident response - #1 tag reports
Part of #1282 

An [example](https://incidentdatabase.ai/cite/51#r1765) of a incident response is on incident 51. While it lacks some important information, it does give the company's perspective on the incident. We will treat this response as our test case. This report should be tagged with `response`, then when rendering the card associated with the report there should be some change to the manner in which it is rendered so it is clearly identified as an incident response. One idea here is to add a badge to the headline saying something like `{author} post-incident response`.

clari182:
- resolves #1395 

You can view the badge on https://deploy-preview-1399--staging-aiid.netlify.app/cite/51#r1765


smcgregor:
This looks good, make sure it handles multiple authors.

clari182:
Yes, this is how it would look with multiple authors:
image

pdcp1 (changes requested):
@clari182 

- We should add tests to this new feature

clari182:
@pdcp1 I added the test but we should add the `response` tag to that report before this is merged into production otherwise it will fail.

pdcp1:
> I added the test but we should add the `response` tag to that report before this is merged into production otherwise it will fail.

@clari182 please add this note to the PR description so we don't forget to add the `response` tag to that report on Prod. Actually, this is something that we can do now, adding a tag doesn't break anything.

PR #1403, opened 2022-11-17
Reorder snapshot page and fix space

smcgregor:


smcgregor:
@cesarvarela this should be a quick and easy one -- it only reorders and puts a space into the text.

cesarvarela:
Merged, looks like this page is missing translations.

PR #1414, opened 2022-11-24
Redesign cite report cards

### LINKED ISSUE #1398: Incident response - #2 redesign report cards
Part of #1282 

The incident cards should likely be reworked since they are very large. Each card can have a "Read More" button that expands the textual contents, while the image can be displayed next to the text similar to the "Latest Incident Report" component on the landing page.

clari182:
- resolves #1398 

Reports are now represented in a shorter version with a Card (similar to `Latest Incident Report `on the landing). They only show 240 chars of the report text. 
They have an `expand` button that expands the text to the full content and hides the left image (to allow better reading).

I haven't included a `collapse` button I didn't think it was necessary, and if the text is very long and you collapse it, users might lose track of where they were located on the `cite` page. But if you think it is I can add it with no problem.


cesarvarela (commented):
> I haven't included a collapse button I didn't think it was necessary, and if the text is very long and you collapse it, users might lose track of where they were located on the cite page. But if you think it is I can add it with no problem.

I agree with this, but at the same time, the first thing I expected was a collapse button 🤔 

My vote is to add them and also to reduce their visual weight by using the outline style, like in the entity pages: https://deploy-preview-1414--staging-aiid.netlify.app/entities/facebook

What do you think @lmcnulty ?

cesarvarela:

cesarvarela:
@clari182 please merge staging to get tests to run

clari182:
@cesarvarela done!

PR #1422, opened 2022-11-29
Fix/Submission Form: hide description when incident id

clari182:
- part of #1346
Solves:

> When adding a report for an existing incident, the description field is still visible. It should be hidden if the incident already exists.


cesarvarela (changes requested):
This change warrants a new test or an update to an existing one.

clari182:
@cesarvarela test added

cesarvarela:
@clari182 ditto from other prs 😆

clari182:
@cesarvarela conflicts fixed!

cesarvarela (approved):

Issue #1344, opened 2022-11-03
Blog navigation


lmcnulty:
When viewing a blog post on mobile, to get back to the blog index, you have to open the menu. Often, you'll go to another blog post after opening a first one, so there should be a link back to the index – probably at both the top and bottom of the page.

PR #1425, opened 2022-11-29
Style Entity and Blog Post Headers

### LINKED ISSUE #1404: Entity Subscription Button Size
The button to subscribe to incidents involving an entity is very large for a button, especially if the name of the entity is several words.

image

It's good that it clearly explains what it does, but that doesn't need to be inside the button itself. For instance, we could have it under the heading and say something like:

> [ ✉️ Notify me ] of new incidents involving *UK Office of Qualification and Examinations Regulation*

lmcnulty:
<details>
<summary>Old screenshot</summary>

image
</details>

Resolves #1404 and #1344.


smcgregor:
I agree the giant button is not a good look, but the way it displays now (at least on wider screens) feels cluttered to me without good visual separation. The encapsulation of the button made it a bit more visually appealing/separated and the white space between the header content and the first body content is a bit weird.

One idea: move the button right by the entity name, make it a bit less salient, and change the wording to just "Subscribe" or "Notify".

lmcnulty:
I like "subscribe," but I think that would confusing given that it's the same as the login button. "Notify" doesn't really work on its own -- it's not clear what one is being notified of. I think "follow" plus a tooltip works:

image

image

cesarvarela (changes requested):

cesarvarela:
And please fix conflicts!

cesarvarela (approved):

PR #1426, opened 2022-11-29
Deploy to Production

cesarvarela:
- incident responses


cesarvarela:
@smcgregor ready for review.

(I'll be checking on some failing tests)

smcgregor:
I am not sure how it would look making this change, but I think it would be better to drop the right sidebar below the line separating the title from the contents of the incident. It feels more cluttered having it on the top. It is also on the page twice at the moment.

Screen Shot 2022-12-04 at 7 07 10 PM

I believe everything else looks good here, but I have not test run everything else yet. Once we resolve the layout choice, I am good to merge.

cesarvarela:
@lmcnulty can you take alook at the comment above, please?

lmcnulty:
#1458 fixes the double-appearance of the similar incidents.

I like having the right sidebar title aligned to the top because it balances against the left sidebar. Does anyone else have a preference?

smcgregor:
@lmcnulty Even if the header of the content goes all the way to the right side (i.e., the title in the top row would be on a single line and the horizontal rule would go across)? I don't like content breaking into the header there since people will scan from left to right and the similar incident panel should be read after looking at the main contents.

PR #1441, opened 2022-12-01
Change defaults on progress plot

smcgregor:
Before:

Screen Shot 2022-11-30 at 4 34 49 PM

After:

Screen Shot 2022-11-30 at 4 31 02 PM


lmcnulty:
My fault, but this results in the "y-axis starts at zero" checkbox being checked when the y-axis *doesn't* start at zero. If you merge [this fix](https://github.com/smcgregor/aiid/pull/1) into your branch it will be good to go.

PR #1446, opened 2022-12-01
Feature/empty alt submit

### LINKED ISSUE #539: Image descriptions
- Not something most users would notice, but this page is not fully ADA compliant.  It's difficult to do programatically, but the alt text needs to describe what the image itself is about, not what the context of a link click is e.g. "Starbucks store sign with trees" as opposed to "Is Starbucks changing its baristas?".  I also don't see anything for the icons below each entry (but I"m not at a full laptop so I can't quite inspect and see)

Issue from Joshua Lugo

clari182:
- resolves #539


clari182:
@cesarvarela conflicts fixed!

PR #1457, opened 2022-12-05
Add Sorting option on discover page

### LINKED ISSUE #747: Discover page - sorts and filters
- User story: "I want the incidents or reports to be sorted by incident date, submitted date, and/or published date of report, ascending or descending, within a time range, because I'd like to see how the incidents or reports progress over time."

Current report display on the Discover page are not sorted by time

image

clari182:
- resolves #747

PR preview: https://deploy-preview-8--clarayoudale-staging.netlify.app/

###  New `Sorting` dropdown which allows sorting by
- Incident Date (asc & desc) - Default is `Incident Date desc`
- Published Date (asc & desc)
- Submitted Date (asc & desc)

### Before merging/testing
Before merging to Staging, Algolia updater needs to be triggered


clari182:
Hey @cesarvarela this is ready for review. However, the tests won't work until the new replicas are added because this updates the discover results list.

smcgregor:
I don't think we made all the replica changes manually before, this setting is not in Algolia's capacity to configure from code?

cesarvarela (commented):
I agree with Sean, I would be surprised if these changes can't be done using the API.

cesarvarela (changes requested):
Tests won't pass on this branch because the AlgoliaUpdater only runs when `process.env.CONTEXT == 'production,` which only happens when the current branch is `staging` or `master.` But you should be able to get them running in your environment.

I think the newly added timeouts shouldn't be necessary since this is just sorting stuff., unless Algolia is particularly slow when doing so 🤔

clari182:
@cesarvarela this is now running on my environment: https://deploy-preview-8--clarayoudale-staging.netlify.app/
Ready for another review

cesarvarela (commented):
Not sure if this is related to this PR, but it looks like we have reports from the future:

image

The `desc` and `asc` options inside the category filter were unexpected to me. The usual UI for this is to keep them separated.

image

More of a UX thing, but it is unclear what the `Relevance` sort means. Also, when I select relevance, the `asc` and `desc` options get unselected, but you can toggle them again 🤔

clari182:
The reports from the future are not related to this PR, but I can take a look separately. 
Regarding the _relevance_ option, I added this to keep the current order of reports, otherwise, it was defaulted to "Incident date desc" which would show a lot of reports from the same incident at once, and it looked confusing IMO. So I added the _Relevance_ option as default which is basically the current order of reports.
But let me know if you think this is necessary or not, or if you have another suggestion.

I did fix the issue that you could select _asc_ or _desc_ while on _Relevance_, now you can only choose those on the dates type sorts. 

Ready for another review.

cesarvarela (changes requested):
Please fix conflicts!

cesarvarela:
This looks weird to me:
image

but if @smcgregor or @lmcnulty are ok I'll merge.

clari182:
@cesarvarela conflicts fixed! I also added a `-` to separate the name and sort direction
image

Let me know if you think it's better this way

lmcnulty:
I personally find that when sites use the terms "ascending" and "descending", I have to think for a moment about what that means in context. I suspect that that's even more likely to be true for users who aren't as engaged in UI patterns and vocabulary. I think it would be better to encapsulate the ascending/descending toggle in the options. So for instance, the options could be:

- Relevance
- Oldest incidents
- Newest incidents
- Oldest submissions
- Newest submissions
- Oldest publications
- Newest publications

I don't think there's any need to allow sorting by "least relevant." That might be too much duplication, though. Maybe when "newest" or "oldest" are selected, there could by another dropdown

```
[▼ Newest        ]  [▼ By incident date    ]
[  Oldest        ]  [  By publication date ]
[  Most relevant ]  [  By submission date  ]
```

I also wonder whether you can fit the dropdowns in the row above? Right now it pushes the content down, leaving some unused space above the fold.

image

clari182:
@cesarvarela ready for another review!
@lmcnulty I went with the first option :
Relevance
Oldest incidents
Newest incidents
Oldest submissions
Newest submissions
Oldest publications
Newest publications

It looked better than the two dropdowns.
I also moved the dropdown to the line above

lmcnulty:
image

Can you move the dropdown to the middle? I think "filter search" should be at the end.

clari182:
Done @lmcnulty !

lmcnulty:
Thanks! Two more things:

image

- It looks the filters need a bit more margin on top, since the sort-by dropdown takes a bit more vertical space.

- I think we should make the styles of the "incident/reports/incident reports" and the "sort by" dropdowns match.

cesarvarela:
From luna's screenshot, these look off-center:

image

lmcnulty:
@cesarvarela The off-center icons only seem to occur in my main, highly-customized Firefox profile and not in a clean one. That weirdness also predates this PR.

clari182:
@lmcnulty The "Display option" dropdown has now the same style as the sorting.
Eventually we would need to update all dropdowns from the filter section to have a similar look&feel, but should be addressed in another PR.
@cesarvarela I can address that styling issue in a future PR also.

cesarvarela (approved):

Issue #1283, opened 2022-10-17
Variant Support


smcgregor:
We are going index high-volume data associated with several system types that tend to produce a large number of incidents. One example includes chatbots that can produce high volume of abusive content. This type of data is generally very structured, and thus is difficult to index in a generic manner. Initially, we will ask for inputs into two text areas: `inputs and circumstances` and `outputs and outcomes`. We will provide instructions on what these mean to the users (please design with 2 sentences of Lorem text).

**Design notes:**

1. Users should have the ability to add variants only via the citation pages for an incident. For now, this should not require a user account, but if we start getting flooded with trash we can lock it down to user accounts.

2. The form should have two text areas for input. One is prompted to be the `input and circumstances`, while the other is prompted to be the `output and outcomes`. People can put whatever they like into these and we can support markdown.

3. Submissions to this form should be flagged as "unreviewed" and editors should have the ability to change the flag to "approved" and "rejected", as well as have the ability to edit/destroy the submission.

4. Regardless of the flag status, the entries in the collection can be displayed on the citation page to which they were submitted, with appropriate display on the UI indicating when something has not been reviewed.

5. In the future, we will likely extend this variant submission process to provide for more explicit structure of inputs and outputs. For example, we want to have the ability to download all GPT-3 prompts and outputs as strings separated from any discussion about the contents of those prompts/responses.

6. The editors will need some user interface to quickly page through and approve variants. I think this can be a table view with the incident number, title, and description next to each variant's inputs/outputs.

7. Most incidents will not have variants, but for those that do, there should be some mention within the statistics panel of the associated citation page.

@pdcp1 I am assigning you on here to lead the feature to delivery. Please feel free to bring in me/other people as may be appropriate.

smcgregor:
Test case: https://incidentdatabase.ai/cite/118

pdcp1:
@smcgregor Please confirm if we are on the same page about this feature:

1. Variants will use the same Report form but with some extra optional fields, like Inputs and Outputs
2. Variants will be stored in the `reports` collection
3. Variants will have a `report_number`, `input` (text field) and `output` (text field)
4. The idea is that Algolia indexes Variants in the same way that it indexes Reports, that's the main reason to store it in the same collection. So users can search for Variants inputs/outputs on the Discovery page.
5. We should add a simpler way to add an Incident Variant. It could be a small form on the citations page that is opened when the user clicks on `Add Variant`
[...]

PR #1467, opened 2022-12-06
Add Variants list section on Incident page

pdcp1:
This PR is related to the issue #1283

### DIFF
--- a/site/gatsby-site/src/graphql/variants.js
+++ b/site/gatsby-site/src/graphql/variants.js
@@ -1,0 +1,65 @@
+import gql from 'graphql-tag';
+
+export const FIND_INCIDENT_VARIANTS = gql`
+  query FindIncidentVariants($report_numbers: [Int]!) {
+    reports(
+      query: {
+        AND: [
+          { report_number_in: $report_numbers }
+          { text_inputs_exists: true }
+          { text_outputs_exists: true }
+          { text_inputs_ne: "" }
+          { text_outputs_ne: "" }
+        ]
+      }
+    ) {
+      report_number
+      title
+      date_published
+      tags
+      text_inputs
+      text_outputs
+    }
+  }
+`;
+
+export const INSERT_VARIANT = gql`
+  mutation InsertVariant($report: ReportInsertInput!) {
+    insertOneReport(data: $report) {
+      report_number
+    }
+  }
+`;
+
+export const UPDATE_VARIANT = gql`
+  mutation UpdateVariant($query: ReportQueryInput!, $set: ReportUpdateInput!) {
+    updateOneReport(query: $query, set: $set) {
+      url
+      title
+      authors
+      submitters
+      date_published
+      date_downloaded
+      date_modified
+      epoch_date_published
+      epoch_date_downloaded
+      epoch_date_modified
+      image_url
+      text
+      plain_text
+      tags
+      flag
+      report_number
+      editor_notes
+      language
+    }
+  }
+`;
+
+export const DELETE_VARIANT = gql`
+  mutation DeleteOneVariant($query: ReportQueryInput!) {
+    deleteOneReport(query: $query) {
+      report_number
+    }
+  }
+`;
--- a/site/gatsby-site/src/templates/cite.js
+++ b/site/gatsby-site/src/templates/cite.js
@@ -23,8 +23,9 @@ import Pagination from '../elements/Pagination';
 import SocialShareButtons from '../components/ui/SocialShareButtons';
 import { useLocalization } from 'gatsby-theme-i18n';
 import useLocalizePath from '../components/i18n/useLocalizePath';
-import { useMutation } from '@apollo/client';
+import { useMutation, useQuery } from '@apollo/client';
 import { UPSERT_SUBSCRIPTION } from '../graphql/subscriptions';
+import { FIND_INCIDENT_VARIANTS } from '../graphql/variants';
 import useToastContext, { SEVERITY } from '../hooks/useToast';
 import Link from 'components/ui/Link';
 import { graphql } from 'gatsby';
@@ -34,6 +35,7 @@ import AllegedEntities from 'components/entities/AllegedEntities';
 import { SUBSCRIPTION_TYPE } from 'utils/subscriptions';
 import { faEnvelope, faPlus, faEdit, faSearch } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import VariantList from 'components/cite/VariantList';
 
 const sortIncidentsByDatePublished = (incidentReports) => {
   return incidentReports.sort((a, b) => {
@@ -61,6 +63,7 @@ function CitePage(props) {
       nlp_similar_incidents,
       editor_similar_incidents,
       editor_dissimilar_incidents,
+      report_numbers,
     },
     data: {
       allMongodbAiidprodTaxa,
@@ -127,6 +130,10 @@ function CitePage(props) {
     title: 'Incident Occurrence',
     mongodb_id: 0,
     isOccurrence: true,
+  });
+
+  const { data: variants, loading: loadingVariants } = useQuery(FIND_INCIDENT_VARIANTS, {
+    variables: { report_numbers },
   });
 
   const taxonomies = useMemo(
@@ -445,6 +452,11 @@ function CitePage(props) {
             </Col>
           </Row>
         ))}
+
+        <VariantList
+          loading={loadingVariants}
+          variants={variants ? variants.reports : []}
+        ></VariantList>
 
         <SimilarIncidents
           nlp_similar_incidents={nlp_similar_incidents}
--- a/site/gatsby-site/src/components/cite/VariantList.js
+++ b/site/gatsby-site/src/components/cite/VariantList.js
@@ -1,0 +1,116 @@
+import React from 'react';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faQuestionCircle } from '@fortawesome/free-solid-svg-icons';
+import Card from '../../elements/Card';
+import { Trans, useTranslation } from 'react-i18next';
+import { Spinner, Tooltip } from 'flowbite-react';
+import { getVariantStatus, VARIANT_STATUS } from 'utils/variants';
+import Markdown from 'react-markdown';
+
+const VariantStatusBadge = ({ status }) => {
+  let badgeClass;
+
+  switch (status) {
+    case VARIANT_STATUS.approved:
+      badgeClass = 'bg-green-100 text-green-800 dark:bg-green-200 dark:text-green-800';
+      break;
+
+    case VARIANT_STATUS.rejected:
+      badgeClass = 'bg-red-100 text-red-800 dark:bg-red-200 dark:text-red-800';
+      break;
+
+    default:
+      badgeClass = 'bg-blue-100 text-blue-800 dark:bg-blue-200 dark:text-blue-800';
+      break;
+  }
+
+  return (
+    <div className={`${badgeClass} text-xs font-semibold px-2.5 py-0.5 rounded capitalize`}>
+      <Trans>{status}</Trans>
+    </div>
+  );
+};
+
+const VariantCard = ({ variant }) => {
+  const { t } = useTranslation();
+
+  return (
+    <Card data-cy="variant-card" className="relative p-4 overflow-hidden flex">
+      <div className="flex">
+        <VariantStatusBadge status={getVariantStatus(variant)} />
+      </div>
+      <div className="flex w-full flex-row items-center mt-2 gap-4">
+        <div className="w-1/2">
+          <div className="font-bold flex items-center gap-2">
+            <Trans>Input and circumstances</Trans>
+            <Tooltip
+              content={t(
+                'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.'
+              )}
+              trigger="click"
+            >
+              <FontAwesomeIcon
+                icon={faQuestionCircle}
+                style={{ color: 'rgb(210, 210, 210)', cursor: 'pointer' }}
+                className="far fa-question-circle"
+              />
+            </Tooltip>
+          </div>
+          <div className="border-1 rounded-lg px-3 mt-2">
+            <Markdown>{variant.text_inputs}</Markdown>
+          </div>
+        </div>
+        <div className="w-1/2">
+          <div className="font-bold flex items-center gap-2">
+            <Trans>Output and outcomes</Trans>
+            <Tooltip
+              content={t(
+                'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.'
+              )}
+              trigger="click"
+            >
+              <FontAwesomeIcon
+                icon={faQuestionCircle}
+                style={{ color: 'rgb(210, 210, 210)', cursor: 'pointer' }}
+                className="far fa-question-circle"
+              />
+            </Tooltip>
+          </div>
+          <div className="border-1 rounded-lg px-3 mt-2">
+            <Markdown>{variant.text_outputs}</Markdown>
+          </div>
+        </div>
+      </div>
+    </Card>
+  );
+};
+
+const VariantList = ({ variants, loading }) => {
+  return (
+    <div className="my-6">
+      {loading && <Spinner />}
+
+      {!loading && variants.length > 0 && (
+        <>
+          <h1>
+            <Trans>Variants</Trans>
+          </h1>
+          <div>
+            <Trans>
+              Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor
+              incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud
+              exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
+            </Trans>
+          </div>
+          <div className={'flex flex-col gap-5 mt-5'}>
+            {variants.map((variant) => (
+              <VariantCard variant={variant} key={`variant-${variant.report_number}`} />
+            ))}
+          </div>
+        </>
+      )}
+    </div>
+  );
+};
+
+export default VariantList;
--- a/site/gatsby-site/src/utils/variants.js
+++ b/site/gatsby-site/src/utils/variants.js
@@ -1,0 +1,26 @@
+const VARIANT_STATUS = {
+  unreviewed: 'unreviewed',
+  approved: 'approved',
+  rejected: 'rejected',
+};
+
+/**
+ * Variant review status
+ * @readonly
+ * @enum {string}
+ */
+module.exports.VARIANT_STATUS = VARIANT_STATUS;
+
+module.exports.getVariantStatus = (variant) => {
+  if (variant.tags.includes(VARIANT_STATUS.approved)) {
+    return VARIANT_STATUS.approved;
+  }
+  if (variant.tags.includes(VARIANT_STATUS.rejected)) {
+    return VARIANT_STATUS.rejected;
+  }
+  if (variant.tags.includes(VARIANT_STATUS.unreviewed)) {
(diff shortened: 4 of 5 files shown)


cesarvarela (commented):
Please merge staging and set it as ready for review!

pdcp1:
@cesarvarela the staging preview build passes 🙌  but I couldn't add the https://deploy-preview-1467--staging-aiid.netlify.app link to the Allowed Request Origins.

pdcp1:
@cesarvarela @smcgregor This PR is ready for testing on this preview link:
- The variant list page: https://deploy-preview-15--pablocosta-staging.netlify.app/apps/variants/
- The variant Incident list: https://deploy-preview-15--pablocosta-staging.netlify.app/cite/10

smcgregor:
Quick front end review

# Blocking

These block deployment to production.

- [ ] When I submitted a new variant for review, it displays in the list of unreviewed variants, then it disappears entirely after I refresh the page. I see the submissions in the review queue though.

# Non-blocking

These need to change shorter term, but the future could land in production before these changes are made.

- [ ] A lot of the variants will initially be input/output pairs for chatbots. Both inputs and outputs can be quite long (see below). I think we need to put the inputs above the outputs rather than side to side.
- [ ] When I added a variant, it pops into the list of variants but the submission form is still displayed. The submission form has poor visual separation from the list of variants so it looks a bit messy and difficult to parse. Maybe two changes would be good here. (1) hide the form after submission (2) present the form slightly differently when it is displayed.
- [ ] We need to add some messaging talking about the variants (e.g., they have not been independently validated). Let's soft launch this feature, then get a blog post talking about the feature that includes all the background information. We will link to the blog post in the variant presentation.
- [ ] The modal window for editing a variant is small. It could also use a "save" button that does not change its review status.
- [ ] The spinner when loading variants on the citation pages doesn't indicate what is or is not loading. Can you shift the rendering of variants to the server side rendering process just as the reports are currently rendered? I know this will make the content out of date periodically, but it will ensure the page still loads very quickly for search engine purposes. We would then display something like this after a user submits a variant: `Your variant has been added to the review queue and will appear on this page within 12 hours. Please continue submitting when you encounter more variants." Most of the time we won't review it in the same day, but it will appear within a day as `unreviewed`.

Screen Shot 2022-12-16 at 3 37 30 PM

smcgregor:
- [ ] The "latest incident" component should render the latest full report, not the latest variant.

pdcp1:
> Can you shift the rendering of variants to the server side rendering process just as the reports are currently rendered?

@smcgregor I agree with you on this. The disadvantage of this server-side approach is that the list will not be updated all the time when a new Variant is created, and also the Variant status/texts will be outdated when the user edits from the citation page. If that sounds good to you, please tell me if you want to add a text explaining this after the user update (edit, approve or reject) a Variant.

smcgregor:
Sounds good to me. At present we are heavily optimizing for the UX of the user submitting content/reading it, but the editors are people that generally know what the system is doing.

Adding text to the effect of, "Your edits will be live within 24 hours" is good here.

pdcp1:
@smcgregor All feedback provided by you is done and ready to review on https://deploy-preview-15--pablocosta-staging.netlify.app/cite/21

@clari182 Please do the CR when you have time

clari182:
Before doing CR I have some thoughts on the front-end review:

- Instead of showing the spinner with `Fetching variants` we should use the Skeleton loader like on the "Table View"
- The variants table should expand to the full width like on the [entities page](https://deploy-preview-15--pablocosta-staging.netlify.app/entities) 
image

But code looks good!

smcgregor:
A few more comments from discussion during the editor's meeting. Most of these comments are for resolving the tension of "what is a variant report" vs "what is a standard report".

The current change set treats `variants` as being different from reports (e.g., by presenting a special form that only has the two variant-associated form elements), but variants are stored in the reports collection and a report is a `variant` if the variant field(s) are specified.

## Changes requested

1. [ ] Present the variant-associated fields in the submission form (last, as optional)
2. [ ] There should not be separate variant/non-variant card components for rendering the variant reports -- the card should just render the variant information if that is all there is
3. [ ] When you edit a report, you should be able to edit all the variant-related fields and adding a variant-related value to a report moves it to the variant listing

We should discuss this comment set before you implement.

smcgregor:
## Editing Guide Changes

I am going to track these and make these changes before we merge to prod.

1. [ ] If someone submits an incident without specifying a variant field, the editor is responsible for specifying the variant field and pointing the report to the existing incident
2. [ ] The variant fields should be as exacting as the incident type allows. For example, the exact prompt that is given to a text-to-text model and the exact output. Discussion/context/description should go in the report text.

pdcp1:
@smcgregor I think a non-blocker solution to this could be to add an `Edit more fields` link into the edit Variant modal that redirects to the existing Report edit page https://staging-aiid.netlify.app/cite/edit?report_number=217&incident_id=21

image

In this form I can add the two Variant fields `text_inputs` and `text_outputs` at the bottom as optional fields.

pdcp1:
@smcgregor This is ready for another round of review https://deploy-preview-15--pablocosta-staging.netlify.app/cite/21

site/gatsby-site/src/components/variants/VariantsTable.js
```diff
@@ -0,0 +1,440 @@
...
+import VariantEditModal from './VariantEditModal';
+import styled from 'styled-components';
+import useToastContext, { SEVERITY } from '../../hooks/useToast';
+import { Trans, useTranslation } from 'react-i18next';
+import { VariantStatusBadge } from './VariantList';
+import { VARIANT_STATUS } from 'utils/variants';
+import { DELETE_VARIANT, UPDATE_VARIANT } from '../../graphql/variants';
+import { LINK_REPORTS_TO_INCIDENTS } from '../../graphql/reports';
+
+const Table = styled.div`
```
clari182, line 19: We're trying to reduce the use of `styled-components`, is there a way you can replace this with regular html (like on the entities page)? Otherwise in the future we'll need to replace it.

clari182 (changes requested):
Code looks good, the only thing I thinks needs change is the table on the variants page VariantsTable.js because it uses `styles-components` and we're trying to ease-out of the practice.

site/gatsby-site/src/components/variants/VariantsTable.js
pdcp1: Done!

smcgregor:
I need to spend a bit more time in review before I merge to staging.

## Blocking

- [ ] I tried to add the full incident report fields to a variant report and it rejected the change: `{"data":{"updateOneReportTranslation":null},"errors":[{"message":"Cannot return null for non-nullable field Report.report_number.","locations":[{"line":3,"column":5}],"path":["updateOneReportTranslation","report_number"]}]}`

## Not Blocking

- [ ] The question marks on the page have different behavior than those on the submit page. I believe the behavior on the submit page is more sensical.
- [ ] "Rather than index variants as entirely separate incidents, we list variations of incidents under the first incident submitted to the database." -> "Rather than index variants as entirely separate incidents, we list variations of incidents under the first similar incident submitted to the database."
- [ ] The spinner when accept/rejecting a variant on the `/apps/variants` page does not gray out the whole background.
- [ ] The "Back to Report" button does not go back to the report. https://deploy-preview-15--pablocosta-staging.netlify.app/cite/edit?report_number=2312&incident_id=21

## TODO

To complete my review I need to do the following,

- [ ] Add a new incident and issue
- [ ] Skim the source code
- [ ] Add a new variant to an existing incident
- [ ] Promote a variant to a full new incident
- [ ] Add full report fields to an existing variant
- [ ] Run a full build with all those changes being made
- [ ] Look into how translations may be handled/not handled in current change set

clari182 (approved):

site/gatsby-site/src/graphql/variants.js
```diff
@@ -0,0 +1,85 @@
...
+      incident_id
+      report_number
+    }
+  }
+`;
+
+export const UPDATE_VARIANT = gql`
+  mutation UpdateVariant($query: ReportQueryInput!, $set: ReportUpdateInput!) {
+    updateOneReport(query: $query, set: $set) {
+      url
```
smcgregor, line 57: You don't need to list `text_inputs` here?

smcgregor:
(more)

## Not Blocking

- [ ] I produced this Screen Shot 2023-01-04 at 10 02 43 AM

by submitting this content:
```
There are a few types of queries that chatbots may struggle to answer:

Complex or open-ended questions: Chatbots may have difficulty understanding and providing a useful response to questions that are very open-ended or that require a high level of comprehension.

Questions with multiple layers of meaning: Chatbots may struggle to understand questions that contain multiple layers of meaning or that are intentionally ambiguous.

Questions about subjective topics: Chatbots may have a hard time providing useful responses to questions about subjective topics, such as opinions or personal preferences.

Questions that require common sense: Chatbots may struggle to provide useful responses to questions that require a certain level of common sense or understanding of the real world.

Questions that require real-time data: Chatbots may have difficulty answering questions that require access to real-time data or information that is constantly changing.
```

## Implementation commentary

When you have a moment, let's discuss a few comments I made inline. Specifically, I am still puzzling through the schema relationship between variant reports and incident/issue reports. I think there are things we should do here to make it easier to manage -- particularly as we introduce more fields that indicate something is a variant.

site/gatsby-site/src/graphql/variants.js
pdcp1: There is no need to return the `text_inputs` and `text_outputs` fields as part of the UpdateVariant mutation. Both fields are part of the `ReportUpdateInput` object but not part of the returned mutation fields.

site/gatsby-site/src/components/cite/SimilarIncidents.js
```diff
@@ -108,8 +108,8 @@ const SimilarIncidents = ({
   editor_similar_incidents,
   editor_dissimilar_incidents,
   flagged_dissimilar_incidents,
-  orientation,
-  className,
```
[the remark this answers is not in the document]
pdcp1: I marked these two fields as optional with default values to avoid Eslint errors on the `cite.js` code when it uses the `<SimilarIncidents>` component without these two fields.

image

site/gatsby-site/page-creators/createWordCountsPage.js
```diff
@@ -26,7 +26,11 @@ const createWordCountsPage = async (graphql, createPage) => {
         }
       }
       latestReport: allMongodbAiidprodReports(
-        filter: { is_incident_report: { eq: true } }
+        filter: {
+          is_incident_report: { eq: true }
+          text_inputs: { in: [null, ""] }
```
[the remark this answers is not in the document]
pdcp1: Changed

pdcp1:
@smcgregor I can't mark your checkboxes as checked, so I added an additional comment with the follow-up. All of them are completed. The only remaining item to finish is your comment about the Variants schema. We should define how we can identify a Variant and a Report. Do you have any thoughts on it being in the same page?

smcgregor:
`Do you have any thoughts on it being in the same page?` Do you mean being in the same card component?

smcgregor:
I was trying to keep this feature a bit reined in since I think it could easily turn into a beastly change set, but I think trying to defer this discussion was causing more problems than it was solving. Let's define each of the types of report we now have with this pull request. They differ in the following properties:

* (Associated with incident): The report has one or more incidents tied to it
* (Is complete report): Has all the fields traditionally associated with a report
* (Has non-variant data): Text, image, submitters, authors, dates, etc.
* (Has variant data): The report has variant data. This is exclusively input/output text at the moment.

1. Incident report: (Associated with incident) and (Is complete report) and (Has non-variant data) and may OPTIONALLY (Has variant data)
2. Issue report: Not (Associated with incident) and (Is complete report) and (Has non-variant data) and may OPTIONALLY (Has variant data)
3. Variant report: (Associated with incident) and NOT (Is complete report) and OPTIONALLY (Has non-variant data) and (Has variant data)

I think this covers all cases where we would want to render the report. An **invalid** report combinations would be:

Issue Variant: Not (Associated with incident) and (Has variant data)

The current variant-related fields that determine (Has variant data) are then:

`text_inputs` and `text_outputs`

In the future this field list could grow, for example, with `image_output` for text-to-image models. This is where the design requirement I am trying to introduce is coming from -- we want to have the ability to introduce new variant-related fields without needing to hunt down all the text_input/output fields. I think the way to do this is to look at `(Is complete report)` and `(Associated with incident)`, since they differentiate incident reports and variant reports. From here, the complete set of changes to the pull request would be:

1. Assume anything that (Is complete report) and (Associated with incident) is an incident report and anything that is NOT (Is complete report) but is (Associated with incident) is a variant report.
2. When (Is complete report) and (Associated with incident) are true, treat it as a normal incident report and render it to the page.
3. For all incident report cards, add the variant data to the end of the cards if it exists. Label each report card that has variant data with some badge indicating the data is present in the card.
4. Also render the complete collection of variant data in the variant section of the `cite` page.

Perhaps an easier way of completing this change set would be to make the UI changes indicated above, but to determine the Algolia update on the basis of the `(Is complete report)` fields? We may need to go verbal on this one to discuss.

pdcp1:
> to determine the Algolia update on the basis of the `(Is complete report)` fields?

@smcgregor I was thinking on the Algolia side of things and maybe we can only index the documents that match the `(Is complete report)` condition.
These documents should have all of these fields set:
- authors
- description,
- epoch_date_downloaded
- epoch_date_modified
- epoch_date_published
- epoch_date_submitted
- image_url
- language
- source_domain
- submitters
- title
- text
- plain_text
- url
- cloudinary_id

What do you think?

smcgregor:
Sounds good to me!

site/gatsby-site/cypress/e2e/unit/AlgoliaUpdater.cy.js
```diff
@@ -133,10 +133,44 @@ describe('Algolia', () => {
       }),
     };
 
+    const filter = {
+      text_inputs: { $in: [null, ''] },
+      text_outputs: { $in: [null, ''] },
```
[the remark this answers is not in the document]
pdcp1: Resolved

site/gatsby-site/src/graphql/variants.js
cesarvarela: I think it is  a good idea to query those fields anyway because it allows the Apollo client to update its cache (even if it is not being used at the moment)

site/gatsby-site/src/templates/cite.js
```diff
@@ -564,16 +572,23 @@ export const query = graphql`
...
         report_number
         title
+        description
         url
         image_url
         cloudinary_id
         source_domain
         mongodb_id
         text
+        plain_text
```
cesarvarela, line 582: Adding these fields almost doubles the amount of data required to build a citation page, which, multiplied by the number of incidents * number of languages, is a lot more data being used.

I wonder if build time changed much after this change 🤔

site/gatsby-site/src/utils/AlgoliaUpdater.js
```diff
@@ -193,12 +193,47 @@ class AlgoliaUpdater {
...
 
-    const reports = await this.mongoClient
+    let reports = await this.mongoClient
       .db('aiidprod')
       .collection(`reports`)
       .find({}, { projection })
       .toArray();
 
+    // Only index Incident Reports
+    reports = reports.filter(
```
cesarvarela, line 203: I would prefer to have only one function that decides if a report is an incident report or not, it looks like this should use the "isCompleteReport" function.

site/realm/data_sources/mongodb-atlas/aiidprod/incidents/rules.json
```diff
@@ -32,25 +32,48 @@
             "search": true,
             "additional_fields": {}
         },
+        {
+            "name": "Add Variant",
```
cesarvarela, line 36: Is this a new role? Should this be part of the default role instead?

site/realm/data_sources/mongodb-atlas/aiidprod/reports/rules.json
```diff
@@ -32,19 +32,35 @@
             "search": true,
             "additional_fields": {}
         },
+        {
+            "name": "Variant",
```
cesarvarela, line 36: I think this rule allows any report with `text_inputs` and `text_outputs` fields set to be created/deleted by anonymous users.

Looks like a security issue 🤔

site/gatsby-site/netlify.toml
```diff
@@ -10,7 +10,7 @@ TERM = "xterm"
   [plugins.inputs]
     enable = false
   [plugins.inputs.preBuild]
-    enable = true
+    enable = false
```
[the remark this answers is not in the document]
pdcp1: This is disabled to disable the tests in order to generate the preview link. It should be enable after merge.

site/gatsby-site/netlify.toml
```diff
@@ -10,7 +10,7 @@ TERM = "xterm"
...
     record = false
     spa = true
     start = 'CI=true gatsby build && gatsby serve -p 8080'
```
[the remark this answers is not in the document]
pdcp1: Yes, that's true. This is the only way that I found to execute the tests before (`preBuild`) the build. So we have to `build & serve` before start testing

site/gatsby-site/src/templates/cite.js
pdcp1: I confirmed that adding these extra fields doesn't affect the build time of the citation page. Anyways, I removed them since they are not needed anymore.

site/gatsby-site/src/utils/AlgoliaUpdater.js
pdcp1: Done

site/realm/data_sources/mongodb-atlas/aiidprod/reports/rules.json
pdcp1: I just added another condition to this rule
```
"title": {
    "$eq": ""
}
```
to only match Variants.
The idea is to allow anonymous users to create Variants

site/realm/data_sources/mongodb-atlas/aiidprod/incidents/rules.json
pdcp1: I prefer to add a separate role to be more precise on the purpose of these new rules. If we want to remove the anonymous Variant creation in the future we can just remove this role and keep the `Default` role unmodified.

pdcp1:
@cesarvarela I answered all of your comments. I will work on the build issue in another PR since is not related to this PR.

cesarvarela (approved):
I'm okay to merge this and get it out of the way, but I prefer to see tests passing either here or in Pablo's env.

pdcp1:
> I'm okay to merge this and get it out of the way, but I prefer to see tests passing either here or in Pablo's env.

@cesarvarela This build on my environment passed all tests but failed for the build timeout, please take a look https://app.netlify.com/sites/pablocosta-staging/deploys/63c809f1a62cb200081b9d1b
Feel free to merge.

cesarvarela:
Before merging this, we have to:

- run build only once
- enable tests
- make sure notification emails are sent after a successful deploy

pdcp1:
@cesarvarela Now it works with tests enabled and a new Netlify plugin that sends the notifications after the deployment succeeds.
This is the last build https://app.netlify.com/sites/pablocosta-staging/deploys/63c869aadfa4bb2cc9a41fcc

image

The process now is:
- Build
- Run tests
- If tests pass, execute the `netlify-plugin-process-notifications` plugin that processes the pending notifications

cesarvarela:
Looks good. It is weird that the build crashes because the schema customization should take care of the missing `text_*` fields 🤔 

Merging this will crash staging as it is.

Is there a report that we know will have a variant?

pdcp1:
> Is there a report that we know will have a variant?

This incident can be treated as a test case with a variant https://incidentdatabase.ai/cite/118

cesarvarela:
Wait, I'm not seeing those fields in the diff, but it should look like this:

`
    type mongodbAiidprodReports implements Node {
      cloudinary_id: String
      tags: [String]
      plain_text: String
      embedding: reportEmbedding 
      text_inputs: String
      text_outputs: String
    }
`

cesarvarela:
🤞

pdcp1:
@cesarvarela the deploy passes! 🙌

PR #1483, opened 2022-12-13
Fix table entity filters

### LINKED ISSUE #1480: Table View – cannot search via entity fields
Search of entities not working likely has something to do with the entities being hyperlinked to the Entities page

image

lmcnulty:
related to #1480.

The search was not working because the value returned by the column accessor was an array of objects, not an array of strings. The react-table didn't know how to filter those. The easy solution, used here, is just to include the entity id and name in a string, then display it as the name with a link. The more principled solution would be to use [a custom filter function](https://react-table-v7.tanstack.com/docs/api/useFilters#:~:text=filter%3A%20string%20%7C%20function(rows%3A%20array%3Crow%3E%2C%20columnids%3A%20array%3Ccolumnid%3A%20string%3E%2C%20filtervalue)%20%3D%3E%20rows%5B%5D) instead of a string. Not sure if that's worth it right now.

## Breakage attempts
- Use all the other filter options
- Filter for entities, e.g. "Youtube"
- Filter for entities with a hyphen, e.g."Arabic-speaking"
- Filter by just a hyphen
  - This includes anything with multiple words. That seems fine.


smcgregor:
Merging this. @LKchemposer can you play with staging and comment on the issue again? I think you are likely looking for the search to search the entity names rather than the description. This fixes the description search, but it is a bit counter intuitive to me that it works this way.

PR #1504, opened 2022-12-19
Feature/discover incident reports

### LINKED ISSUE #936: Discover incidents vs reports
The results of searches in the Discover app correspond to reports, not incidents. Several people have reported that they found this confusing, including Janet, myself, and the subject in the [think-aloud](https://github.com/responsible-ai-collaborative/aiid/issues/902). It seems as if this design was inherited from the time when reports were the top-level entries in the database and incidents were just an id field shared among reports. This is no longer the case.

I think we should rework the Discover app to search incidents rather than reports. This is what people will naturally expect from something called an AI _Incident_ database. We could just try to explain it better, but I think that would be swimming against the current. Thinking about what it would mean to search incidents rather than reports, lets go through the fields that currently exist in the Algolia index:

<table>
<tr>
  <td><tt>title</tt> </td>
  <td> We now have these for incidents, and they're generally higher-quality than the report titles.</td>
</tr>
<tr>
  <td><tt>authors</tt> </td>
  <td> I suspect think very few people actually care about this, but if we want to keep it, we could just join the authors of all the reports. We could call it something like "reporters" or "reported by" since incidents don't have authors.</td>
</tr>
<tr>
  <td><tt>description</tt> </td>
  <td> We have these for incidents now.</td>
</tr>
<tr>
  <td><tt>epoch_date_downloaded</tt> </td>
  <td> I doubt that anyone wants to search by this – it can probably be omitted.</td>
</tr>
<tr>
  <td><tt>epoch_date_modified</tt> </td>
  <td> ^</td>
</tr>
<tr>
  <td><tt>epoch_date_submitted</tt> </td>
  <td> ^</td>
</tr>
<tr>
  <td><tt>epoch_date_published</tt> </td>
  <td> This could be interesting if there were a significant number of incidents that were reported on long after they occurred. I haven't checked, but I suspect that this is rarely the case. It can probably be omitted too.</td>
</tr>
<tr>
  <td><tt>image_url</tt> </td>
  <td> We don't have these for incidents – for the TSNE visualization I just used the first report's image, which I think is fine.</td>
</tr>
<tr>
  <td><tt>language</tt> </td>
  <td> We could make this an array that contains all languages for which reports of the incident exist.</td>
</tr>
<tr>
  <td><tt>source_domain</tt> </td>
  <td> Some of the "authors" are just the sites that the reports comes from, so I think this could be combined with authors.</td>
</tr>
<tr>
  <td><tt>submitters</tt> </td>
  <td> I don't think many people want to search by this.</td>
</tr>
<tr>
  <td><tt>url</tt> </td>
  <td>^</td>
</tr>
<tr>
  <td><tt>editor_notes</tt> </td>
  <td> ^</td>
</tr>
<tr>
  <td><tt>text</tt> </td>
  <td> I think we should just concatenate the report texts and let Algolia search those. It shouldn't show up in the UI – if people want to read about a report beyond the title and description, they'll click the result and go to the citation page.</td>
</tr>
<tr>
  <td><tt>incident_date</tt> </td>
  <td> This is the main date that matters, and I think we should highlight it more. You should be able to sort results by it.</td>
</tr>
<tr>
  <td><tt>classifications</tt> </td>
  <td>I believe these are already associated with incidents, not reports. They should also probably be emphasized more – the user in the think-aloud said that they were the most interesting part.</td>
</tr>
</table>

An additional bonus to using incidents instead of reports is that the index will be smaller and we'll be able to stay on Algolia's free tier longer.

clari182:
- closes #936 

This feature requires `node src/scripts/algolia-update.js` to be triggered prior to being tested. Please fork this branch and test locally on your environment.

On the discover page, when choosing "Incidents" in the dropdown, it will display the incident's title and description, as opposed to the report's.

Deploy preview: https://deploy-preview-5--clarayoudale-staging.netlify.app/


site/gatsby-site/cypress/e2e/integration/discover.cy.js
```diff
@@ -180,4 +180,15 @@ describe('The Discover app', () => {
 
     cy.contains('button', 'Clear Filter').should('be.disabled');
   });
+
+  it.only('Should display incidents instead of reports when selection Incidents view', () => {
```
cesarvarela, line 184: Leftover `it.only`

cesarvarela (changes requested):

clari182:
Hey @cesarvarela, ready for another review. The preview link is: https://deploy-preview-5--clarayoudale-staging.netlify.app/

cesarvarela (approved):

PR #1505, opened 2022-12-20
Add blog link to post

### LINKED ISSUE #1344: Blog navigation

lmcnulty:
Resolves #1344.

image

<details>
<summary>Old Screenshots</summary>

image

image

</details>


site/gatsby-site/src/tailwind.css
```diff
@@ -544,6 +544,10 @@
         @apply grid grid-cols-1 gap-[6px];
     }
 
+    .tw-header-back-link {
```
cesarvarela, line 547: We should avoid adding these. Best practice is just to add classes to components https://tailwindcss.com/docs/reusing-styles

cesarvarela (changes requested):

lmcnulty:
This will conflict with #1425, so I'm combining them.

PR #1508, opened 2022-12-22
Fix tests for related reports

### LINKED ISSUE #1490: Fix related reports tests
Not sure if broken or simply outdated.

- Should show a list of related reports
- Should show a preliminary checks message
- Should show related reports based on semantic similarity
- Should not show semantically related reports when the text is under 256 non-space characters

clari182:
- closes #1490


site/gatsby-site/cypress/support/commands.js
```diff
@@ -64,6 +64,7 @@ Cypress.Commands.add('setEditorText', (value, selector = '.CodeMirror') => {
   return cy
     .get(selector)
     .first()
+    .click()
```
cesarvarela, line 67: Does this make a difference?

Issue #994, opened 2022-08-22
Placeholder image variety


lmcnulty:
Currently all reports without images use a placeholder with tiled "AI Incident Database>AI Incident Database" text. It would be nice to add more variety to placeholders, and perhaps allow submitters to pick from a set of images that could broadly represent various AI incidents. We've also talked about using AI-generated images, which could be interesting given our subject matter.

From #903.

PR #1510, opened 2022-12-22
Add generated thumbnails

lmcnulty:
Per #994. Keeps the textual look for placeholders, but makes them more distinct and flavorful.

image

## Breakage attempts

- Generated thumbnail shows on submit form.
  - Entering an image url, allowing the image to load, then removing the URL results in the generated image reappearing.
  - Entering a url that doesn't point to an image results in the generated image.
- Generated thumbnails appear in the discover app in multiple colors and with correct report numbers.
- Generated thumbnails appear in the Spatial View with correct incident numbers.
- OpenGraph image is correct on [citation page](https://www.opengraph.xyz/url/https%3A%2F%2Fdeploy-preview-1510--staging-aiid.netlify.app%2Fcite%2F1).
  - There still seem to be cases where the Cloudinary image won't load. Even though I added a second og:image as a fallback, [opengraph.xyz](https://deploy-preview-1510--staging-aiid.netlify.app/cite/300) won't show it in such cases. This might be different depending on the site, however. I'm not sure we can do any better.


smcgregor:
Looks great, a few preliminary points and I'll let someone else take final review:

1. I think the image generation function should likely be its own component that is imported by the Cloudinary rather than exported from the cloudinary script.
2. The image is generated by the scripting environment, which likely doesn't work when sharing a link. I don't think we are sharing the report images, so this might be fine, but can you see if there are any issues here and indicate 👍 or 👎 ?
3. I am a bit torn on whether we give the report number in the image, which is not something that we make front and center unless the report is an "issue" and not an "incident". I am good on proceeding here as-is, particularly since we can change it later without it being a Cloudinary issue.
4. Translations?

smcgregor:
I am not following "We'll have to escape the public id, which is automatically done by the cloudinary AdvancedImage component, which we can't use for a metaImage url.". Maybe it is faster to "show" on this than "tell"?

lmcnulty:
The Algolia URL we we've used so far were generated by the AdvancedImage component. That's a component which renders an image client-side, so it can't be used to serve an image directly. That was mostly just a note to myself that I needed to find how to get the cloudinary url otherwise. It turns out it's just: 
```javascript
new CloudinaryImage(publicID, {…}).createCloudinaryURL()
```

smcgregor:
You are saying we can't render the image at the moment regardless?

lmcnulty:
We can.

cesarvarela (commented):
Before doing an in-depth review, what's the reason to do this client side instead of using Cloudinary's image generation features?

generation:  https://cloudinary.com/documentation/layers
fallbacks: https://cloudinary.com/documentation/advanced_url_delivery_options

We can *probably* build a cloudinary url that generates a colorful fallback image that shows if there is no uploaded image, with minimal client side code, that also works on meta tags.

I wonder how heavy the image generation algorithm it is, especially considering that the website is already heavy on cpu usage, and the lighthouse score is kind of low already:

image

lmcnulty:
To be honest, using Cloudinary for this never occurred to me. However, I'd still prefer to go with this solution, at least for now.

I did some unscientific testing, and for the small images like the ones in discover app, the canvas operations to generate one image take about 50ms on my 2015 laptop, whereas it takes a couple hundred ms to download each image on my fairly fast connection. On the other hand, for the larger images, e.g. on the cite page, in can take more like 500ms for a generated image to render, whereas it still only takes a couple hundred milliseconds for the network-accessed images -- surprisingly, it seems like most of time is actually consumed by the overhead of the HTTP request rather than the data transfer itself. I don't think there's an obvious winner when it comes to performance. I will say that, if necessary, we can improve the performance of the image generation, whereas we can't improve the speed a network request.

One reason to generate the fallback client-side is that sometimes Cloudinary will respond with a 400 error and no fallback, for instance, with this url: https://res.cloudinary.com/pai/image/upload/d_fallback.jpg/f_auto/q_auto/c_fill,h_480/v1/reports/s3.reutersmedia.net/resources/r/?m=02&d=20150422&t=2&i=1042301179&w=1200&r=LYNXMPEB3K115. I tried to solve this [back in July](https://github.com/responsible-ai-collaborative/aiid/pull/792) and the best solution I could find then was detecting when an image fails to load and then switching to the fallback url. Reflecting on that some more, the issue there is probably not how we display the images, but how we select the Cloudinary IDs. However, even if we fix that, this is all a specific instance of the general fact that making a network request to a third party is less reliable than doing a computation locally. Client-side image generation will work just the same if Cloudinary has an outage, or if it changes its API or pricing, or if the user has an unreliable connection or very zealous firewall.

Having a url we could use for the meta tags would be nice, but I think just showing our logo in social media previews is fine.

cesarvarela:
Yes, most of the problems lie in how we upload images to Cloudinary, which ends up with weird public IDs that cause errors everywhere. Doing proper uploads should fix most of those 404 you mention.

I'm not sure about the Cloudinary going down thing, considering that offering reliable/fast image hosting is their reason for existing, and their SLA is 100%, but I degrees.

I'll merge this one, and let's keep an eye on the PageSpeed score, especially the Time to Interactive metric, which is crazy high for a site that's primarily static text and images.

cesarvarela:
Please fix conflicts!

PR #1518, opened 2023-01-04
Use local names for languages in switcher

### LINKED ISSUE #939: The items in the language switcher should be localized
Currently, when the language switcher displays "Spanish" in the menu even when viewing the site in Spanish.

In general, it's preferable for language switchers to always display the language's name for itself (e.g. "English, Español), since some people might not know the name of their language when displayed in another language. This probably isn't as much of a problem with just English and Spanish, which are fairly mutually intelligible in this context, but I think we should go along with best practices.

lmcnulty:
Resolves #939.

image


smcgregor:
Wow. I love that this is in the locale object.

Issue #963, opened 2022-08-18
Image Cropping


lmcnulty:
Report images get cropped badly for especially tall or wide images. An example can be see, ironically, in report 1528.

<img src="https://user-images.githubusercontent.com/25443411/185445771-b9206cb9-4c64-4ec8-9cc2-b143a828df44.png" width="400"/>

The report's image is: 
<img src="https://user-images.githubusercontent.com/25443411/185446121-63d93bf9-a4cc-4e55-8423-3e636b9c1779.png" height="400"/>

This is not a very good choice of image. But since many of the images are scraped from the article pages and submitted unchanged by users, we need should try to handle such images gracefully, or else reject them. This particular case in which the middle of the image is completely blank is unusual, but there are more common cropping failures. A lot of reports have the logo of the site that published them as their image. This is usually cropped badly. E.g.

![image-sizing-rules](https://user-images.githubusercontent.com/25443411/185446985-4010dd88-2ec0-495c-929f-b14d2d34c5ba.png)

I think that for images with a white background or with an unusual aspect ratio, say 3 or more times as large in one dimension as the other, it's probably better to "fit" rather than "fill" the image into the available space. If we wanted to be clever, we could also crop images to the region with the most entropy, although I'm not sure if Cloudinary will allow us to do so.

smcgregor:
@lmcnulty if you release an issue that was in the backlog, make sure you put it back in the backlog.

PR #1519, opened 2023-01-04
Thumbnail cropping

lmcnulty:
Per #963. Builds on #1510.

Before:

image

After:

image

Even though this uses Cloudinary's SDK, it seems to increase the number of images that fail to load. That may or may not be worth it.


lmcnulty:
As I said, this one causes more loading failures. I think we should find a solution for #1709 before coming back to this.

PR #1522, opened 2023-01-07
Deploy to Production

smcgregor:


smcgregor:
@lmcnulty it looks like the fetch info button is broken in staging now. Can you take a look?

smcgregor:
I'll merge something small to staging and see what happens.

smcgregor:
No joy. I am not sure where the server-side log for this function lives -- it is not coming up in Netlify/MongoDB. Anyone know where to look?

lmcnulty:
`/api/semanticallyRelated` also seems to be immediately 500ing.

smcgregor:
Merging this. I'll peg the deploy to an older version and make the API breakage a daybreaker if it then also breaks prod.

PR #1532, opened 2023-01-11
Avoid send notifications on `gatsby build` for Cypress

pdcp1:
This PR avoids sending notifications on the `gatsby build` fired by Cypress process setting a env variable `CI=true`


smcgregor:
ha, it was checking the send queue twice then... Merge?

pdcp1:
> ha, it was checking the send queue twice then... Merge?

@smcgregor Exactly 🤦  In the previous PR I added the tests to the prebuild process, but I forgot to avoid sending the notifications at that moment. So the notifications are processed on pre-build and post-build.
Now I use a `CI=true` env variable to avoid that.

Please merge.

PR #1547, opened 2023-01-17
Prevent gatsby i18n theme to duplicate pages

### LINKED ISSUE #1546: Duplicated report pages
This is the same issue fixed in #1503, but for report pages, so the fix should be the same.

i.e.

https://incidentdatabase.ai/es/fr/reports/2292

cesarvarela:
This prevents URLs like these from being generated: https://incidentdatabase.ai/fr/es/reports/217

The reason why these changes work is obscure and dependent on internal implementation,https://github.com/gatsbyjs/themes/issues/104 but it is the easiest way: 

closes #1546

### DIFF
--- a/site/gatsby-site/cypress/e2e/unit/pageCreators/createReportPages.cy.js
+++ b/site/gatsby-site/cypress/e2e/unit/pageCreators/createReportPages.cy.js
@@ -10,14 +10,6 @@ const response = {
         },
         {
           report_number: 2,
-          language: 'en',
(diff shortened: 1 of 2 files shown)


pdcp1:
@cesarvarela One related test is failing, please review.

pdcp1 (changes requested):
Please fix failing test

pdcp1 (approved):
It looks good 👍

Issue #803, opened 2022-07-14
Multi lingual support for dates


cesarvarela:
Dates should be localized across the site.

 Format should be pulled from https://github.com/responsible-ai-collaborative/aiid/blob/master/site/gatsby-site/i18n/config.json

smcgregor:
I think the fastest thing here is to adopt `yyyy-mm-dd` universally since it won't require per-language setting, is generally interpretable by most cultures, and the sort order matches chronological order.

PR #1564, opened 2023-01-25
Standardize date formatting

cesarvarela:
closes https://github.com/responsible-ai-collaborative/aiid/issues/803

### DIFF
--- a/site/gatsby-site/src/components/blog/PostPreview.js
+++ b/site/gatsby-site/src/components/blog/PostPreview.js
@@ -1,13 +1,11 @@
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 import { GatsbyImage as Img } from 'gatsby-plugin-image';
 import { format } from 'date-fns';
 import { LocalizedLink } from 'gatsby-theme-i18n';
 import { Trans } from 'react-i18next';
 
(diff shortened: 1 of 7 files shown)


lmcnulty (commented):
I grepped for `format(` and there are a couple places still using other formats:

```
src/pages/apps/classifications.js
244:      picker.startDate.format('MM/DD/YYYY') + ' - ' + picker.endDate.format('MM/DD/YYYY')

src/pages/apps/reports.js
208:      picker.startDate.format('MM/DD/YYYY') + ' - ' + picker.endDate.format('MM/DD/YYYY')
360:    return <>{format(new Date(s.props.cell.value * 1000), 'MMM d, yyyy')}</>;

src/pages/summaries/incidentsOverTime.js
58:    title: (d) => format(d.date, 'M/d/yy') + ' – ' + d.count + ' ' + d.series,
95:            value={format(startDate, 'yyyy-MM-dd')}
```

Those pages aren't localized yet though, so it's not inconsistent for them to use month-day-year.

cesarvarela:
You are right; I went with the easy ones :P I dug a little deeper and updated some more. Much more than this is probably not worth it at the moment.

lmcnulty (approved):

PR #1581, opened 2023-02-02
Deploy to Production

cesarvarela:
- new fallback thumbnails
- collapsable menu
- fix discover app trailing slash


smcgregor:
# Blocking Issue

* You can no longer scroll to the right on the table view. I think this was introduced with the sidebar changes.

# Non-Blocking issues

* The minified react error is back, but this is also on the current prod :(
* The tooltip on the sidebar appears only when the mouse is over the icon, not the whole button.

PR #1585, opened 2023-02-06
Fix issue when submitting multiple forms in a row

### LINKED ISSUE #1579: Report submission – Cannot submit reports _sometimes_
All required fields in step 3 were filled out, but it ran into this issue:

Screen Shot 2023-02-01 at 4 13 00 PM

Refreshing and submitting only at step 2 is the work-around, so there may be some issue in step 3.

clari182:
- closes #1579 

To test, submit two incidents in a row, without refreshing. The incidents should be created correctly.

Also fixes Editor issue where, after the first submit, the editor had a "Text required" error displayed.

https://deploy-preview-1585--staging-aiid.netlify.app/


cesarvarela (changes requested):
Please add a test!

site/gatsby-site/cypress/e2e/integration/submit.cy.js
```diff
@@ -1324,4 +1324,65 @@ describe('The Submit form', () => {
...
+      `https://www.arstechnica.com/gadgets/2017/11/youtube-to-crack-down-on-inappropriate-content-masked-as-kids-cartoons/`
+    );
+
+    cy.get('[data-cy="fetch-info"]').click();
+
+    cy.wait('@parseNews');
+
+    cy.get('input[name="authors"]').type('Something');
+
+    cy.setEditorText(
```
cesarvarela, line 1354: The fetch info button fills the text field, you shouldn't need to set a text manually 🤔

site/gatsby-site/src/components/forms/SubmissionWizard/StepOne.js
```diff
@@ -321,7 +321,11 @@ const FormDetails = ({
               value={values.text}
               onChange={(value) => {
                 setFieldValue('text', value);
-                setFieldTouched('text', true);
+                if (!submissionComplete) {
```
cesarvarela, line 324: What's this for, looks very specific, like we are misusing formik 🤔

cesarvarela (commented):
A couple of comments before merging

site/gatsby-site/src/components/forms/SubmissionWizard/StepOne.js
clari182: Unfortunately, formik does not work well with the `Editor component` and the function `resetForm` doesn't reset tits errors, and touched values.
So this entails that once the form is submitted the text editor appears surrounded by red with the text "Editor is required" because it has not been reset by Formik.

site/gatsby-site/cypress/e2e/integration/submit.cy.js
clari182: You're right... Although it was like this before on other tests, but I tested and it works without the manual set. So maybe this wasn't working right on a previous version of Cypress. I removed the manual setEditorText when using the fetch Info option

cesarvarela:

clari182:
@cesarvarela ready for review

cesarvarela (approved):

Issue #1365, opened 2022-11-09
Update to Gatsby v5


cesarvarela:
Possible benefits

- partial hydration: faster load times, faster time to interactive
- slice API: reduce build times
- ...

(This is a reminder/suggestion, maybe it is not worth it)

smcgregor:
I think we should likely wait until there is a 5.0.1 or 5.1.0 version, unless there is a pressing reason.

smcgregor:
They are now on 5.2, so we can proceed here if we want to.

PR #1590, opened 2023-02-08
Fix rel alternate meta tags 

cesarvarela:
This one got a little bit out of hand, but I need to patch one of the `gatsby-i18n-theme`, so I converted it to a local plugin.

The good thing is that Gatsby abandoned the package anyway, and we'll have to update it to support #1365  

builds on top of #1604


cesarvarela:
closed in favor of #1624

PR #1604, opened 2023-02-10
Fix and test for hydration errors

### LINKED ISSUE #1512: Production Minified React Error
Production has a series of what are likely hydration errors. The app behaves fine, but this has been popping up periodically for a while and we need to take care of it.

`react-dom.production.min.js:131 Uncaught Error: Minified React error #418; visit https://reactjs.org/docs/error-decoder.html?invariant=418 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.`

cesarvarela:
This should fix most (all?) hydration errors and adds a test that should catch them if they reappear.

closes #1512


cesarvarela:
I hadn't had so much fun in years.

clari182 (approved):

PR #1608, opened 2023-02-11
Fix staging build

cesarvarela:


lmcnulty:
I didn't see this – I think this is the same as #1609, except that the `value_json` needs to be parsed. I'm not sure why GitHub is showing that PR as if the new taxonomies one hadn't been merged.

cesarvarela:
That's more correct yeah, let's use yours

PR #1605, opened 2023-02-11
Deploy to Production

cesarvarela:
- new taxonomies

**Because of migrations, we'll need to disable/enable triggers before merging into production**

PR #1609, opened 2023-02-11
Update gatsby node location as attribute

lmcnulty:
Per #1605, updates the maps integration in gatsby-node.js to match the new classifications system.


cesarvarela:
You could try to checkout out directly from upstream and force-push the commit.

cesarvarela:
It passed anyway, merging 💥

PR #1610, opened 2023-02-11
Make schema dummy_fields list of objects instead of object

lmcnulty:
This fixes the broken classifications app from https://github.com/responsible-ai-collaborative/aiid/pull/1605#issuecomment-1426559484.


smcgregor:
@cesarvarela merging since I think you are 👍 , please feel free to revert if mistaken

Issue #1618, opened 2023-02-14
TSNE Visualization title loading


lmcnulty:
From Nikiforos' feedback on the the TSNE visualization:

> Probably a backend issue: there's a ~2 second delay for loading the title and image of the incident which reduces the visualization's utility for the use case above. If at least the titles could be cached and loaded instantaneously-ish, that would be great.

Issue #1600, opened 2023-02-09
TSNE Visualization Width


lmcnulty:
There are some screen sizes/zoom levels at which the tsne visualization is partly cut off:

![image](https://user-images.githubusercontent.com/25443411/217648989-0e583260-db6f-4452-9375-7d18f02e433b.png)

On large screens, the visualization does not fill all the available space:

![image](https://user-images.githubusercontent.com/25443411/217649232-428978d4-f3a3-4d92-b806-610a2a3f45b0.png)

Instead, it should always fit the total available space.

PR #1621, opened 2023-02-14
TSNE: Pre-query titles, use full width, set z-indices

### LINKED ISSUE #1601: TSNE Visualization colors hidden
The TSNE visualization currently shows incidents with higher incident numbers above ones with lower numbers. That's preferable to the reverse, but since newer classification are less likely to be classified, the colorful, classified incidents tend to get buried in colorless ones, which looks bad. Instead, classified incidents should appear above unclassified ones.

lmcnulty:
Resolves #1601, #1600, and #1618.


cesarvarela (commented):
The behavior is a little weird using the MacBook trackpad, but that's another issue.

PR #1624, opened 2023-02-15
Fix hydration errors

cesarvarela:
- upgrade dependencies and remove the need for using `--legacy-peers-flag`
- update to mdx v2, which required tons of changes everywhere (reference: https://mdxjs.com/migrating/v2/#update-mdx-files)
- newest Gatsby plugins require node v18 which has been add to the supported runtimes on Netlify recently
- ...


cesarvarela:
Closing, so it doesn't take a build box on netlify until it is working here: https://github.com/cesarvarela/aiid/tree/fix-hydration-errors

PR #1626, opened 2023-02-16
Update candidates schema

lmcnulty:
This needs to merged before https://github.com/responsible-ai-collaborative/nlp-monitoring/pull/1. Then, probably after we wait to see how it does after the workflow runs for some time, we can merge https://github.com/responsible-ai-collaborative/aiid/pull/1317.


smcgregor:
@lmcnulty should this still be merged?

lmcnulty:
@smcgregor Yes. It didn't actually need to be merged before the nlp-monitoring one because the schema is not enforced on the upload side. However, it does need to be merged before the frontend updates.

lmcnulty:
With that being the case, there is no need for this to be separate from #1317.

PR #1628, opened 2023-02-16
Fix report ID link

### LINKED ISSUE #1583: Report's title anchor href is missing report number
So instead of only updating the address bar it sends you to the homepage:

image

lmcnulty:
Resolves #1583.

### DIFF
--- a/site/gatsby-site/src/components/reports/ReportCard.js
+++ b/site/gatsby-site/src/components/reports/ReportCard.js
@@ -55,7 +55,10 @@ const ReportCard = ({ item, className = '', incidentId }) => {
       >
         <div className="flex items-center w-full justify-between">
           <div>
-            <LocalizedLink to={`#r${item.report_number}`} className="max-w-full cursor-pointer">
+            <LocalizedLink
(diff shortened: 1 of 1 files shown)


site/gatsby-site/src/components/reports/ReportCard.js
```diff
@@ -55,11 +54,11 @@ const ReportCard = ({ item, className = '', incidentId }) => {
       >
         <div className="flex items-center w-full justify-between">
           <div>
-            <LocalizedLink to={`#r${item.report_number}`} className="max-w-full cursor-pointer">
+            <a href={`#r${item.report_number}`} className="max-w-full cursor-pointer">
```
cesarvarela, line 57: `LocalizedLink` doesn't handle the hash?

cesarvarela:
Please fix conflicts!

site/gatsby-site/src/components/reports/ReportCard.js
lmcnulty: It adds a trailing slash `/cite/1/#r1/` which breaks the scrolling.

Issue #1235, opened 2022-10-09
Discover, Table View – Add export options


LKchemposer:
User: "After configuring and filtering my search results for incidents I'm interested in, I want to download current results into a CSV or Excel to analyze them, or into a PDF for easy attachment."

pdcp1:
Closing this since it was merged into Staging

PR #1629, opened 2023-02-17
Add export to CSV option to Discover page

pdcp1:
This closes https://github.com/responsible-ai-collaborative/aiid/issues/1235

image


site/gatsby-site/src/components/discover/Controls.js
```diff
@@ -36,6 +37,10 @@ const Controls = () => {
         </div>
 
         <div className="flex">
+          <div className="bootstrap place-self-center">
```
cesarvarela, line 40: I think we should stop using `bootstrap` stuff?

cesarvarela (commented):
This only exports the first page. Is that what's expected?

site/gatsby-site/src/components/discover/Controls.js
pdcp1: Changed to flowbite button

pdcp1:
> This only exports the first page. Is that what's expected?

That's what the Algolia React component support. Only the visible page results are accessible to export.
To export all results we need to change the export component to get the results in another way, maybe a direct Algolia API call.
What do you think @LKchemposer? Is it OK to only export the visible page results? You can play with it [here](https://deploy-preview-1629--staging-aiid.netlify.app/apps/discover/?display=details&is_incident_report=true&page=1).

cc: @smcgregor

cesarvarela:
I'll merge after we get a response to the questions above.

LKchemposer:
@pdcp1 @cesarvarela This looks great! I think this is OK to roll out for now as a demo, although the feature might benefit from a few things from the UX perspective, in order of importance:

1. If only results of the first page is exported, there needs to be some communication about that, maybe via a notification?
2. The fields in the CSV does not contain `incident_id` (and `incident_title`) which are most likely what people want to have when exporting
3. Naming the button `Export` with the icon is friendly to a wider range of users
4. Not sure if the `text` field is needed in the CSV, as it takes a large chunk of the data. my guess is that people who would use the export feature do so to keep the db's indexes of the reports, rather than the text of the reports themselves. for the text, we can direct them to the `Download Complete Database` link at the bottom of the site instead, maybe also via the notification?

Please let me know your thoughts on these. I'd imagine some of these are easier to implement than others.

pdcp1:
@LKchemposer Thanks for your feedback.
This is how it looks like after the changes:

Screenshot 2023-02-23 at 17 25 53

Note: We couldn't add the `incident title` field to the result set since it's not an Algolia field.

pdcp1:
@cesarvarela the build passed after a second run here https://app.netlify.com/sites/staging-aiid/deploys/63f7da6a25b0bc1846698f04

cesarvarela (commented):
The test failing is passing on staging with no problem. I think there is something in this changeset that's creating some flakiness 🤔

cesarvarela:
Scratch that, it is failing here too: https://github.com/responsible-ai-collaborative/aiid/pull/1651, I'll take a look at why it is flaky

pdcp1:
@cesarvarela the tests pass now

Issue #1631, opened 2023-02-21
Editor's Form – Hyperlinked cards get easily misclicked


LKchemposer:
The hyperlinks for the cards which open the page on the same tab makes it really easy to misclick, when they take some time to load and expand out the form suddenly. No progress on the editor's form is saved when the form is reloaded.

Suggestions, in order of preference:

1. Open hyperlinks in a new tab
2. Disable hyperlinks (for this form only)
3. Prompt confirming going to the page

<img width="625" alt="image" src="https://user-images.githubusercontent.com/39468345/220316406-e9c69216-b91d-4360-bc84-c52f57775cfb.png">

lmcnulty:
The easiest change is making those links open in a new tab.

With a normal HTML form, trying to navigate away from the page while there are unsubmitted field entries would prompt a confirmation. But since we're using Gatsby, internal links don't cause a full page reload – they replace the current page content with that of the linked page. It seems that prompting before switching to a different page is unsupported https://github.com/gatsbyjs/gatsby/issues/10726. Instead they recommend saving the form state in localStorage. That seems like the most correct solution, it will just be a bit more work. @LKchemposer, if you would regard opening the links in a new tab as resolving this issue, then I think we should open another for saving the form state.

I think this also reveals that we shouldn't allow the similar incidents to appear suddenly and change the position of the other elements on the page. I'm going to open another issue for this.

lmcnulty:
Closing because related incident links now open in a new tab, and remaining concerns are covered by #1633 and #1677.

PR #1632, opened 2023-02-21
Make submission form related incident links open in new tab

lmcnulty:
This is the first-line solution to https://github.com/responsible-ai-collaborative/aiid/issues/1631.

### DIFF
--- a/site/gatsby-site/src/components/RelatedIncidentsArea.js
+++ b/site/gatsby-site/src/components/RelatedIncidentsArea.js
@@ -1,6 +1,6 @@
 import React, { useState, useEffect } from 'react';
 import { Button, Spinner } from 'flowbite-react';
-import { LocalizedLink } from 'gatsby-theme-i18n';
+import { useLocalization } from 'gatsby-theme-i18n';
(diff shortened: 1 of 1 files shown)


site/gatsby-site/src/components/RelatedIncidentsArea.js
```diff
@@ -80,19 +82,21 @@ const RelatedIncidentsArea = ({
               <span>
                 {val?.incident_id && (
                   <span className="text-lg">
-                    <LocalizedLink
-                      to={'/cite/' + val.incident_id}
+                    <a
```
cesarvarela, line 85: I'm guessing `LocalizedLink` doesn't support the `target` prop?

lmcnulty: I don't believe so – it didn't work until I changed it to an `<a>`.

cesarvarela: Seems they fixed it on Gatbsy v5 but we aren't there yet.

PR #1644, opened 2023-02-23
Fix/discover language switch

### LINKED ISSUE #1642: Cannot Select Language When on Discover App
It redirects back to the language you are switching from. It appears to only happen on the discover application.

clari182:
- closes #1642 

This is caused because of a double re-render of the discover page when switching languages.
The first render is due to the language switch which is correct, but this leads to the `searchState` being updated and the useEffect being rendered a second time, but this time with the previous language (for some reason I cannot comprehend).

More specifically the issue happens because of line 216, which is necessary to update the URL params and query.

This is not the best solution to the issue, so let me know if you can think of something else that fixes this.

[Deploy Preview #1644](https://deploy-preview-1644--staging-aiid.netlify.app/)


cesarvarela (approved):
It looks like the failing test is unrelated to this change.

It also may be a good idea to always to a page refresh when switching language, but it can be done in another pr.

PR #1655, opened 2023-02-27
Save sortBy in URL and use that param to sort

### LINKED ISSUE #1650: Sort Settings Not Saved
The sort is not being saved to the URL, so when you share or bookmark your current view of the page it is not re-loaded according to your current view.

It also looks like there may be an issue in the parameters when you set the display option to `Incident and Issue Reports`.

`https://incidentdatabase.ai/apps/discover/?display=details&is_incident_report=false%7C%7Ctrue&page=1`

Specifically, this doesn't look right to me even though it works: `is_incident_report=false%7C%7Ctrue`.

clari182:
- close #1650 

Sort is maintained when reloading the page or sharing the URL after a sort has been applied.


site/gatsby-site/cypress/e2e/integration/discover.cy.js
```diff
@@ -257,7 +257,7 @@ describe('The Discover app', () => {
 
     cy.location('search', { timeout: 8000 }).should(
       'equal',
-      '?display=details&is_incident_report=true&page=1'
+      '?display=details&is_incident_report=true&page=1&sortBy=relevance'
```
pdcp1, line 260: We should add another test to check if the sort selector is set with the value from the URL

pdcp1 (changes requested):

clari182:
@pdcp1 done!

pdcp1 (approved):

PR #1672, opened 2023-03-01
Remove /en prefix from related incident links

### LINKED ISSUE #1670: Related incidents links in English have non-existent /en/ prefix
image

lmcnulty:
Resolves #1670.

### DIFF
--- a/site/gatsby-site/src/components/RelatedIncidentsArea.js
+++ b/site/gatsby-site/src/components/RelatedIncidentsArea.js
@@ -5,6 +5,7 @@ import { Trans, useTranslation } from 'react-i18next';
 import SimilaritySelector from './SimilaritySelector';
 import { faArrowDown } from '@fortawesome/free-solid-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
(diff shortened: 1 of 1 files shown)


smcgregor:
I think this might be a bigger conversation, but at some point we should look at where we have the implied English assumption. I am good with this solution for now though.

site/gatsby-site/src/components/RelatedIncidentsArea.js
```diff
@@ -83,9 +83,10 @@ const RelatedIncidentsArea = ({
                 {val?.incident_id && (
                   <span className="text-lg">
                     <a
-                      href={`/${locale}/cite/${val.incident_id}`}
+                      href={`${locale == 'en' ? '' : '/' + locale}/cite/${val.incident_id}`}
```
cesarvarela, line 86: Is this because `LocalizedLink` doesn't handle the `target` prop?

There are 2 path localization utils that would make a more robust solution: `localizedPath` (built-in)  and `switchLocalizedPath` (util made by me) please use any of those.

cesarvarela (changes requested):

cesarvarela (approved):

PR #1674, opened 2023-03-01
Deploy to Production

cesarvarela:
- allow multiple submits
- save discover sorting on url 
- remove language in related reports links


cesarvarela:
I just realized that uBlock origin blocks the rollbar.js script 🤔

cesarvarela:
@smcgregor ready for review

smcgregor:
The discover app is redirecting to english when you navigate to it. Production doesn't have this problem.

PR #1680, opened 2023-03-02
Feature/cite improvements

### LINKED ISSUE #1639: Citation Page Improvements
* Move "Suggested citation format" content into the "Incident Stats" panel. It doesn't need its own panel
* Add the "Notify me of updates" button to the header in addition to the tools panel
* Move the "Reports Timeline" into the "Reports" section. It doesn't need to be as high on the page as it is.
* Any other look and feel improvements that may come up

### LINKED ISSUE #1679: Report card image overflow
Discovered in https://github.com/responsible-ai-collaborative/aiid/pull/1594#issuecomment-1442516873

image

This happens in smaller screen resolutions

clari182:
- closes #1639 

- [x] Add the "Notify me of updates" button to the header in addition to the tools panel (https://github.com/responsible-ai-collaborative/aiid/issues/1639#issue-1595738836) -> **for this I added a Badge next to the incident's title (like the responded Badge), since the `Tools card` moved up before the `description`**
- [x] Move the "Reports Timeline" into the "Reports" section. It doesn't need to be as high on the page as it is.(https://github.com/responsible-ai-collaborative/aiid/issues/1639#issue-1595738836)
- [x] Regarding the "Suggested Citation Format" (above): could we instead make this part of the "BibTex Citation" link in the "Tools" block? So the title of that link in the Tools Block would say "Citation Info" and the content would include both the "Suggested Citation Format" and the "BibTex Citation". (https://github.com/responsible-ai-collaborative/aiid/issues/1639#issuecomment-1442299525)
- [x] Bump the "Tools" block up to just under the Description. As a user, I don't notice these tools when I start scrolling because by then I'm looking for the content itself.(https://github.com/responsible-ai-collaborative/aiid/issues/1639#issuecomment-1442299525)
- [x] Size down the Carousel image block so it doesn't take up so much vertical space.(https://github.com/responsible-ai-collaborative/aiid/issues/1639#issuecomment-1442299525)

- closes #1679
- [x] Fix report overflow image on big screen sizes


smcgregor:
@clari182 thanks for the checklist/details. My design review below, I have not looked at the implementation yet.

* @Janetbananet and I differ in our preferred orderings. I prefer the description at the top of the page, while she expressed a preference for the tools being higher. If anyone else would like to weigh in on this one, please do so! Otherwise, I'll ask you to move the description back up. In the event the description is lower, then it needs to be in a card or otherwise improved in its display. Going from card to content without a card, then back to card, looks messy. When the description is at the top, it appears more like a subtitle and looks better.
* The big image component very often does not present the image particularly well due to proportion differences. Maybe this needs to have its own issue and we roll it back on this one?
* The big image component shows reports...I think it should likely go after the timeline or at the very least under the `## Incident Reports` section.
* The spinner on page load is distracting. The user will naturally look at the spinner until it resolves and our system response time is long enough for that to be distracting. I don't think we are going to speed up the server response time here, but we can make it so that users don't stare at it. At worst case, I think we should make it so the button is not dynamically checking subscribe status.
* The unevenness on the cards looks a bit messy to me. For example, here there is a dangling line below the image that stops mid sentence while there is still a bunch of whitespace to the right. Ideally, either the content would run to the end of the card or it would never have that last line that goes beyond the image. This may be a new issue rather than something on this PR.
Screen Shot 2023-03-04 at 3 42 24 PM

* The citation info modal has a smaller header than the contents within the modal.
* There is a random line below the BibTek block's copy button that looks like a rendering error.
* There is extra whitespace below the close button on the modal
* The copy buttons both have equal saliency to the close button. I think we typically use buttons without color fill for these sorts of actions.
* Let's add details to "why" we are asking people to use a specific citation format. Something like a context popup or tooltip next to "Suggested Citation Format" indicating, "Citing this specific format will make it possible for the Incident Database to find your research and include it on this page."
Screen Shot 2023-03-04 at 3 47 32 PM

* It looks like this branch conflicts with `staging` at the moment.

clari182:
> @clari182 thanks for the checklist/details. My design review below, I have not looked at the implementation yet.
> 
> * @Janetbananet and I differ in our preferred orderings. I prefer the description at the top of the page, while she expressed a preference for the tools being higher. If anyone else would like to weigh in on this one, please do so! Otherwise, I'll ask you to move the description back up. In the event the description is lower, then it needs to be in a card or otherwise improved in its display. Going from card to content without a card, then back to card, looks messy. When the description is at the top, it appears more like a subtitle and looks better.

**Done: I agree, the description looks better first, without a card and the Tools right below. I think adding one more card for description is not necessary, it's more readable as it is now, at the top.**

> * The big image component very often does not present the image particularly well due to proportion differences. Maybe this needs to have its own issue and we roll it back on this one?

**With this are you referring to the carousel? I changed the carousel now, to use the flowbite one (instead of the bootstrap). I think it looks better. For the sizing of the image I'll take another look but fix in another PR.**

> * The big image component shows reports...I think it should likely go after the timeline or at the very least under the `## Incident Reports` section.

**Done**

> * The spinner on page load is distracting. The user will naturally look at the spinner until it resolves and our system response time is long enough for that to be distracting. I don't think we are going to speed up the server response time here, but we can make it so that users don't stare at it. At worst case, I think we should make it so the button is not dynamically checking subscribe status.

**Done**

> * The unevenness on the cards looks a bit messy to me. For example, here there is a dangling line below the image that stops mid sentence while there is still a bunch of whitespace to the right. Ideally, either the content would run to the end of the card or it would never have that last line that goes beyond the image. This may be a new issue rather than something on this PR.

> 
> Screen Shot 2023-03-04 at 3 42 24 PM
> 

**Maybe the solution here is to display less text when it's collapsed. Since the redesign of this card implied that the text float around the image, it's a little bit tricky to check if a line is not going under the image or not on the collapsed view. This all depends on how the text has been written on the issue. But I'll create a separate issue for this to come up with an improved solution.**

> * The citation info modal has a smaller header than the contents within the modal.

**Done**

> * There is a random line below the BibTek block's copy button that looks like a rendering error.

**Done**

> * There is extra whitespace below the close button on the modal

**Done**

> * The copy buttons both have equal saliency to the close button. I think we typically use buttons without color fill for these sorts of actions.

**Done**

> * Let's add details to "why" we are asking people to use a specific citation format. Something like a context popup or tooltip next to "Suggested Citation Format" indicating, "Citing this specific format will make it possible for the Incident Database to find your research and include it on this page."
> 
> Screen Shot 2023-03-04 at 3 47 32 PM
> 

**Done**

> * It looks like this branch conflicts with `staging` at the moment.

**Done**

@smcgregor ready for another review

smcgregor:
It looks like you addressed all my points. I think someone else can do final code review then :shipit:

site/gatsby-site/src/components/cite/ImageCarousel.js
```diff
@@ -15,39 +15,33 @@ const ImageCarousel = ({ nodes }) => {
...
+              </a>
+            </h3>
+            <p className="bg-black/50 m-0 text-white px-1 rounded text-center">
+              {value.source_domain}
+            </p>
+          </div>
+        </div>
+      ))}
+    </Carousel>
+    // </div>
```
cesarvarela, line 44: looks like a leftover

site/gatsby-site/src/components/cite/NotifyButton.js
```diff
@@ -0,0 +1,33 @@
+import { faEnvelope } from '@fortawesome/free-solid-svg-icons';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import React from 'react';
+const { Button, Spinner } = require('flowbite-react');
```
cesarvarela, line 4: Why are you using `require` instead of `import` not sure if it makes a difference, but I would try to keep it consistent.

site/gatsby-site/src/components/cite/Tools.js
```diff
@@ -0,0 +1,64 @@
...
+import React from 'react';
+import { Trans, useTranslation } from 'react-i18next';
+import { RESPONSE_TAG } from 'utils/entities';
+import CitationFormat from './CitationFormat';
+import NotifyButton from './NotifyButton';
+
+function Tools({ incident, incidentReports, isSubscribed, subscribeToNewReports, subscribing }) {
+  const { t } = useTranslation();
+
+  const { isRole, loading } = useUserContext();
```
cesarvarela, line 16: This has a high chance of causing hydration errors, in my experience.

site/gatsby-site/src/components/cite/citationTypes/BibTex.js
```diff
@@ -0,0 +1,70 @@
...
+          color={'gray'}
+          onClick={() => {
+            navigator.clipboard.writeText(bibTex);
+            addToast({
+              message: 'BibTeX copied to clipboard',
+              severity: SEVERITY.success,
+            });
+          }}
+        >
+          <FontAwesomeIcon icon={faCopy} className="fas fa-times" style={{ marginRight: '1ch' }} />
```
cesarvarela, line 62: This missing the `titleId` prop causes hydration errors too.

site/gatsby-site/src/components/cite/citationTypes/Citation.js
```diff
@@ -0,0 +1,91 @@
...
+          color={'gray'}
+          onClick={() => {
+            navigator.clipboard.writeText(text);
+            addToast({
+              message: 'Citation format copied to clipboard',
+              severity: SEVERITY.success,
+            });
+          }}
+        >
+          <FontAwesomeIcon icon={faCopy} className="fas fa-times" style={{ marginRight: '1ch' }} />
```
cesarvarela, line 83: ditto

cesarvarela (commented):
Please ensure the `DEV_SSR` flag is turned on in your gatsby-config file. You should be seeing hydration errors (other than the existing ones)

cesarvarela (changes requested):
This still has the old hydration errors. Please merge staging.

clari182:
@cesarvarela ready! Latest [deploy](https://app.netlify.com/sites/staging-aiid/deploys/6410b2e77a30d2026fb39e05) worked
Deploy preview: https://deploy-preview-1680--staging-aiid.netlify.app/

cesarvarela:

cesarvarela (commented):
This makes changes related to a test that failed. I wonder if it is because it is introducing flake 🤔

clari182:
Hey @cesarvarela conflicts fixed.
About the test, I tried several times and it hasn't failed for me anymore. But how do you suggest I can do to make sure it doesn't happen again?
Otherwise this is ready for review.

cesarvarela:

clari182:
@cesarvarela done!

cesarvarela (approved):

PR #1699, opened 2023-03-06
Fix date-related hydration errors

cesarvarela:
Some hydration errors were tough to reproduce locally because, when building, the "server" date is the same as the "client" date. But on Netlify, the server date (I suppose UTC) will differ from the client date based on where whoever is accessing the page is.


cesarvarela:
Finally a landing page without hydration errors:  https://deploy-preview-1699--staging-aiid.netlify.app/

smcgregor:
Yay!

pdcp1 (approved):

Issue #1697, opened 2023-03-06
Blog preview images at strange widths


smcgregor:
Go to the blog listing and narrow the width of the page until the below happens.

<img width="963" alt="Screen Shot 2023-03-06 at 12 44 11 PM" src="https://user-images.githubusercontent.com/64780/223226434-6ae4a588-8ea4-450f-85e6-a711cdbcf64e.png">

PR #1700, opened 2023-03-07
Fix blog index width

lmcnulty:
Resolves https://github.com/responsible-ai-collaborative/aiid/issues/1697.

This is the result of the classes `max-50rem:max-w-full` in the Layout component, which don't match the breakpoints we use elsewhere and is just the Tailwind port of [a media query dating back to time immemorial](https://github.com/responsible-ai-collaborative/aiid/blob/ae4f49dcc3d72d8a9ca950e50c16d7941860aa49/site/gatsby-site/src/components/layout.js#L51).


cesarvarela (changes requested):
I'm getting errors and a blank page, please merge staging.

cesarvarela (approved):

PR #1708, opened 2023-03-07
Default sorting to featured replica index

### LINKED ISSUE #1706: Default sort "relevance" not sorting based on config
It looks like the default sorting is using index `instant_search-en` instead of `instant_search-en-featured`

### LINKED ISSUE #1719: "Clear Filters" button is broken
When you apply a filter then click "clear filters" the UI removes the filter, but the data returned does not change.

clari182:
- closes #1706 
- closes #1719 

The issue here is that the `instant_search-en-featured` index wasn't being used because it's only used if there are no params in the URL (no refinement list). But now, by default the discover has `page`, `is_incident_report`, `details` and `sortBy` in its URL which then defaults to `instant-search-en` index.

In order to fix this issue, we need to change between indexes depending on the selected filters, the params in the URL and the selected sort.

Deploy preview: https://deploy-preview-13--clarayoudale-staging.netlify.app/

Expected behavior:

## If no filters are applied and the sort is set to "Relevance"
You should see the **featured** results. (It should match the results of the Replica Index `instant_search-en-featured` from your Algolia

## If no filters are applied and the sort is set to something else than "Relevance"
You should see all results, not the featured ones

## If filters are applied, regardless of the applied sort
You should see all results, not the featured ones


cesarvarela (changes requested):
There is no reason for tests to fail. Also, please add a new one ensuring the correct items are shown since it is easy to access them from the `config.js` file.

clari182:
@cesarvarela tests won't go through on staging because there is an Algolia update that is reflected in my environment.
I also had to change the `config.js` format of the featured results, since it was a key-value object, and when used it's automatically ordered by key. I had to change this to run the test that checks that the reports are displayed in the same order as the config.

This PR also fixes #1719 

Ready for another round of review.

cesarvarela (approved):

PR #1716, opened 2023-03-08
Cite page -> Report Card title click should expand card

### LINKED ISSUE #1714: Cite page -> Report Card title click should expand card
When clicking on the report card inside the cite page, all it does is scroll up to that card. I think it's better if it expands/collapse the card instead.

The other option would be to link to the report URL but I think it's not the most expected user experience.

clari182:
- closes #1714


site/gatsby-site/src/components/reports/ReportCard.js
```diff
@@ -76,11 +73,11 @@ const ReportCard = ({ item, className = '', incidentId }) => {
           onClick={(e) => e.stopPropagation()}
           onKeyDown={(e) => e.stopPropagation()}
         >
-          <a href={`#r${item.report_number}`} className="max-w-full cursor-pointer">
-            <h5 className="text-xl font-bold tracking-tight text-gray-900 dark:text-white w-full hover:text-primary-blue">
+          <div role="presentation" onClick={toggleReadMore} onKeyDown={toggleReadMoreKeyDown}>
```
lmcnulty, line 76: I'm guessing that the `role="presentation"` is there because the linter complained about having event handlers on a div with no aria role. This silences the warning, but it doesn't solve the problem that the element can't be focused in keyboard or screen-reader navigation. It's not a big deal since there's another button that does the same thing at the bottom, but to be semantically correct I'd prefer if this were a `<button>` styled as ordinary text.

Also, I noticed now that cursor switches to a pointer when hovering over the text, even though nothing happens when you click there. If possible, it would be best to the cursor to a pointer only when it's the margins, which can be clicked.

image

lmcnulty (approved):

cesarvarela (changes requested):
The preview URL has hydration errors; it seems like it is running behind staging or has introduced new ones.

cesarvarela (approved):

PR #1727, opened 2023-03-13
Add copies of CSETv1 taxonomy for 3 annotators

lmcnulty:
This is needed because CSET intends to have multiple annotators annotate the same incident for reliability purposes.


smcgregor (approved):

cesarvarela (changes requested):
I think this justifies adding/updating some tests.

cesarvarela (approved):

PR #1730, opened 2023-03-14
Add Rollbar to backend functions

pdcp1:


cesarvarela (commented):
I think some functions should still let errors go through after logging to Rollbar because, with this changeset, they will fail silently.

For example, `onNewUser` before this change, if there were an error, it would return a 500 error, and you'll know that something went wrong, but now it will always return 200, even when it fails.

The above doesn't necessarily apply to all of them though.

pdcp1:
@cesarvarela I just add a `throw error;` statement on some of the functions that need to fail if an error occurs. Please review.

PR #1736, opened 2023-03-14
Deploy to Production

cesarvarela:
- small ui fixes
- live data toggle on static incidents table
- login redirects


cesarvarela:
@smcgregor ready for review

Issue #1732, opened 2023-03-14
Remove Bootstrap from Site


kepae:
Today's blog post is great and has a ton of supporting link references. However, I almost missed them because they appear as slightly bolded text of the same color as the standard text. Hovering doesn't highlight or change the appearance in any way, either (though the cursor changes to the "click" format).

https://incidentdatabase.ai/blog/chatgpt-incidents-and-issues/

<img width="791" alt="image" src="https://user-images.githubusercontent.com/4238598/225107017-3cf50c70-d2ac-4d34-a940-af65d6be52aa.png">

I'm not familiar with the styling resources for the site at large, but other links, such as the variant paragraph description, are blue and more clearly distinguishable.

<img width="965" alt="image" src="https://user-images.githubusercontent.com/4238598/225107357-dc0f78b6-8351-4436-bb72-187d3a38bcb2.png">

lmcnulty:
This arose when we added the Tailwind Typography plugin. The [Tailwind site itself](https://tailwindcss.com/docs/typography-plugin) has bold. uncolored links, but they're underlined. None of the links on our site have underlines, so this might be a conflict between a pre-existing global style and Tailwind Typography.

kepae:
Seems so! Perhaps the best option is to tackle it at the config level with [a custom theme for `--tw-prose-links`](https://tailwindcss.com/docs/typography-plugin#element-modifiers).

Otherwise, can maybe use the [`prose-a` element modifier](https://tailwindcss.com/docs/typography-plugin#element-modifiers) in the template. But, the config level sounds better for consistency.

(I'm trying to prototype it locally, but managing my node versions first...)

kepae:
The config looks more involved and has site-wide impact (?), but modifying the template allows a quick look.

Changing the template: `site/gatsby-site/src/templates/post.js`

```javascript
<StyledMainWrapper className="prose prose-a:text-blue-600 hover:prose-a:text-blue-500">
  <MDXProvider components={MdxComponents}>{children}</MDXProvider>
  <Author>By {mdx.frontmatter.author}</Author>
</StyledMainWrapper>
```
-->
<img width="659" alt="image" src="https://user-images.githubusercontent.com/4238598/225136844-60edf07c-9845-4718-a354-24ce9eb690da.png">

(These are perhaps _too blue_? Not my call.)

Issue #1617, opened 2023-02-14
Complete removal of styled components


cesarvarela:
There are some Components still using styled-components here and there. They should be removed before #1365 to make the upgrade easier.

cesarvarela:
They are also causing some hydration errors:

![image](https://user-images.githubusercontent.com/1172479/218624769-75f76585-b5cf-4eb8-9a79-404c3889c97a.png)

cesarvarela:
It is also not possible at the moment to use partial hydration with styled-components.

PR #1749, opened 2023-03-20
Remove bootstrap/styled components from discover page

clari182:
- Remove bootstrap and styled-components from the apps/discover page

- Part of https://github.com/responsible-ai-collaborative/aiid/issues/1732

- Helps with #1617


cesarvarela (changes requested):
This is something that's already happening, but since you are at it:
image
(images overflow discover cards round corners)

The text should be left aligned, and the items list shouldn't scroll horizontally:

image

Mobile view filters look a bit weird. Items look like they are disabled:
image

clari182:
Fixed @cesarvarela

cesarvarela (approved):

PR #1778, opened 2023-03-29
Fix discover page warnings

### LINKED ISSUE #1717: Fix warnings on Discover app
There is an Algolia warning and a few invalid props being passed to inputs:
image

clari182:
- closes #1717

### DIFF
--- a/site/gatsby-site/src/components/discover/SORTING_LISTS.js
+++ b/site/gatsby-site/src/components/discover/SORTING_LISTS.js
@@ -4,6 +4,7 @@ const SORTING_LIST = [
   {
     default: true,
     name: 'relevance',
+    value: 'instant_search-en',
     value_en: 'instant_search-en',
     value_es: 'instant_search-es',
     value_fr: 'instant_search-fr',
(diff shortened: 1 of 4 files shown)


site/gatsby-site/src/components/discover/SORTING_LISTS.js
```diff
@@ -4,6 +4,7 @@ const SORTING_LIST = [
   {
     default: true,
     name: 'relevance',
+    value: 'instant_search-en',
```
cesarvarela, line 7: By default, it should be `en`. Why do you need this?

clari182: The Algolia [sortBy widget](https://www.algolia.com/doc/api-reference/widgets/sort-by/js/) needs its items to have a value for its first iteration. Since that wasn't existing it was returning the "undefined" warning.

cesarvarela: Yeah, but language is never empty, on the first render it already knows the language, it should go directly to `value_en`. I think there is some executive order that needs to be tweaked 🤔

clari182: Yeah that's what I thought also, but after debugging I realized the issue happens when the Sorting Algolia widget is mounted, and it uses the SORTING_LIST as its `items`. So the first time, when it mounts, it automatically tries to find the `value` field that doesn't exist. I'll dig more into this next week but I couldn't find another way to fix this.

cesarvarela:
I'll give it a try and if I can't get it working like I mention in the comment above I'll merge.

PR #1784, opened 2023-03-29
Deploy to Production

cesarvarela:
- log backend function errors to Rollbar 
- bootstrap cleanups
- dynamic citation pages for new incidents (DSG support is coming in another pr)

### configuration update

A new Atlas Secret needs to be added called `rollbarAccessToken ` with the appropriate value to get this deployment to work.

image


cesarvarela:
@clari182, it looks like the related reports are a bit flaky, they failed twice.

clari182:
@cesarvarela looks like that when there are more than 3 related reports, they don't always appear in the same order. For example the same results appear in different order in staging and prod, so the test fails. I'm working on a fix now [here](https://github.com/responsible-ai-collaborative/aiid/pull/1800)

cesarvarela:
It looks like the new incident routes are not working 🤔

smcgregor:
I just created a new fake incident and navigated to the dynamic page: https://staging-aiid.netlify.app/cite/470

One non-blocking rough point and one potentially blocking issue.

1. The next/previous buttons don't work. This doesn't need to be fixed immediately, but perhaps short-term the previous button should always be enabled.

2. It is showing a bunch of taxonomy classifications for the CSET taxonomy. Is this a ghost associated with incident 470 or a bug? I think it may be a blocking bug.

cesarvarela:
Merged a potential fix 🤞

smcgregor:
@LKchemposer newly ingested incidents will now be immediately editable!

LKchemposer:
This is a great upgrade!

I noticed that I can edit the newly ingested report data immediately (in addition to the incident data), although the translations are not available immediately (just want to make sure that this is expected behavior).

image

Also, follow-up question: since the new report can be immediately editable, could this (in the near future) be applied for reports associated with an existing incident as well?

smcgregor:
* Translations run at build time. We are not likely to change that in the short term.
* I believe it would not be too difficult to make the reports immediately editable. It may be part of the DSG change set that will make more pages dynamic soon.

PR #1819, opened 2023-04-03
Fix cloudinary fallback logic

### LINKED ISSUE #1797: Fallback Image Not Displaying
https://incidentdatabase.ai/cite/6/

I suspect what is happening here is that the fallback image is rendering faster than the 404 is returning, so it renders the fallback then the 404 replaces the fallback image.

Screen Shot 2023-03-30 at 9 39 27 PM

clari182:
- Fixes #1797 

I rollbacked the changes made previously (checking if URL is valid), and re-used the current logic. 
I only added that if the image is placeholder, it isn't consider valid.


cesarvarela (commented):
Merging because this is kind of broken atm, but I would like to see some tests.

PR #1824, opened 2023-04-04
Reuse table component

### LINKED ISSUE #1811: Sorting on Incident Table
https://incidentdatabase.ai/apps/incidents/

We should be able to sort on the columns.

cesarvarela:
closes #1811


clari182 (changes requested):
Hey @cesarvarela I have a few comments, maybe this isn't meant to be fixed on this PR, let me know!

## The classifications, reports & variants tables don't look the same as the incidents table:

### **Classifications /apps/classifications**
image
image

- Table looks a bit different, with no borders between rows 
- Move the actions to the last column and change the edit icon button to an "Edit" light color button
- Pagination is a little bit different

### **Reports table /apps/reports**
image
image

- Borders between rows are missing
- Column order is missing
- Pagination is different than the one in incidents

### **Variants table**
image

- We should add a header to this page with "Variants" or similar
- The table's header is missing the cols order
- The search inputs don't have a bold placeholder (although personally, I prefer it this way, so maybe we could take out the bold from the incident's table placeholders).
- I'm not sure if this should be treated on this PR but the Actions col needs some work IMO. Maybe a dropdown so you can choose the action you want to do, but 4 buttons look a little odd.

### For reference this is the **Incidents table /apps/incidents**
image


Maybe these tables are meant to be fixed later on, if that's the case, Incident's table looks great and code is good so approved :)

cesarvarela:
Oh, the classifications one was on purpose, but I forgot about the others.

clari182:
@cesarvarela, ok, ignoring classifications. Reports' and variants' tables still look the same, does this mean it's meant to be fixed in another PR or on this one?

PR #1828, opened 2023-04-05
Cloudinary images edge cases

clari182:
- Sometimes the Cloudinary on error wasn't triggering because it uses the `<AdvancedImage` component ref `imageElement`. When on first render there was an error, the `<PlaceholderImage` was rendered instead making the ref null.
- This caused that, on window resize, the on error wasn't triggering again so no error was detected.


cesarvarela (commented):
I'll merge, but someone, someone please https://github.com/responsible-ai-collaborative/aiid/issues/1833

Issue #1830, opened 2023-04-05
User Account Signup Error


smcgregor:
I just received this a 404 on trying to confirm a new user account created via the "Subscribe to Major Updates" button.

Path visible in the console: `https://realm.mongodb.com/api/client/v2.0/app/aiidstitch2-fuwyv/auth/providers/local-userpass/confirm`

Replication:

1. visit: https://incidentdatabase.ai/signup/
2. enter a valid email
3. click "Subscribe to Major Updates"
4. click the confirmation email confirmation link
5. the page that loads will have a spinner

Also note, when signing up for the newsletter, I don't believe there is supposed to be a confirmation email.

See also #1829

smcgregor:
I think Major Updates is supposed to be a manual process where we sign them up to the Google Groups after the fact. Is that not the case?

pdcp1:
Exactly, the Major Updates creates a user with the email entered and a dummy password (because it's required by Realm). Then we should go to the `App Users` page on Realm, collect all new emails and add them manually to the Google Group.

smcgregor:
`App Users` indicates who has dummy passwords? I think the `Major Updates` case likely should not have an email confirmation, while the New Incidents subscription should.

smcgregor:
Got it, thanks!

pdcp1:
@smcgregor I forgot to mention that you should also check the `Pending` users since they are also subscribers.

<img width="1123" alt="Screenshot 2023-04-05 at 13 30 38" src="https://user-images.githubusercontent.com/6564809/230145087-7943083b-1ab3-40d1-a7ab-b0ef06e28c5d.png">

pdcp1:
> I think the `Major Updates` case likely should not have an email confirmation, while the New Incidents subscription should.

@smcgregor regarding this, there is no way to send the email confirmation to one case and don't send the email in the other case since both are threated as the same type of users (Email/Password users).
In my opinion, when a user subscribe to `Major Updates` a toast message "Thanks for subscribing to our Newsletter!" should be displayed and is up to the user to confirm or not the email.
If he/she doesn't confirm the email, it will still be there as a `Pending Confirmation` and it can be added to the Google Group:

<img width="995" alt="Screenshot 2023-04-05 at 15 35 48" src="https://user-images.githubusercontent.com/6564809/230173782-ec0a750b-687b-418f-9551-f4c6df6fc136.png">

smcgregor:
[...]

PR #1831, opened 2023-04-05
Fix user email confirmation

pdcp1:
This closes https://github.com/responsible-ai-collaborative/aiid/issues/1830


cesarvarela (changes requested):
Since there are no changes to the `package.json` file, the lock file shouldn't be included in the PR.

cesarvarela (approved):

Issue #1788, opened 2023-03-29
Add autocomplete to Developers, Deployers, and Harmed Parties


Janetbananet:
The Developers, Deployers, and Harmed Parties fields in the Submission Form are all currently free-form text fields. This can cause a splintering effect in the Entities table because people may use slightly different names (ex: "Meta" vs "Meta Platforms") which will degrade the built in analytics and counts in the Entities table.

Adding an autocomplete feature based upon existing entries will help alleviate this issue.

smcgregor:
Added this to the backlog

PR #1835, opened 2023-04-06
Add autocomplete to entities on submission form

lmcnulty:
Resolves https://github.com/responsible-ai-collaborative/aiid/issues/1788.

This currently works by statically querying all entities and passing them to the tag input as options. We might need to query entities dynamically if the number increases dramatically, but this seems to work fine as a first pass.

image

The tag input is a bootstrap component, and the menu breaks if we don't have it in a bootstrap class.


cesarvarela (changes requested):
Please update at least one of the existing tests to use this feature.

cesarvarela (changes requested):
Please update the test with staging again. `incident_id` is now `incident_ids`, and it is an array.

cesarvarela (approved):

PR #1841, opened 2023-04-06
Deploy to Production

cesarvarela:


cesarvarela:
@smcgregor ready for review

smcgregor:
Not blocking:

* I think when you navigate to the admin page rather than reload the page at the admin page, the data doesn't load. Log out, visit the admin page, then click "subscribe", login, and you will be directed to the admin page but no contents will load.

I'll deploy this Thursday AM unless you let me know I should wait.

PR #1858, opened 2023-04-12
Remove bootstrap & styled-components from toast component

clari182:
- Helps with #1732 
- Helps with #1617 

Uses tailwind toasts instead of bootstrap

Where can you test the toasts:
- On [cite/10](https://deploy-preview-1858--staging-aiid.netlify.app/cite/10/), click `Subscribed to Updates` btn inside tools. This will display the info toast.
- When subscribing to notifications on an incident. This will display the success toast.
- Trying to login with a wrong username/password. This will display the fail toast


pdcp1 (approved):
Everything looks good.
We may change the size of the close button, but it's just a personal opinion.

Screenshot 2023-04-13 at 15 20 01

PR #1874, opened 2023-04-14
Language updates

smcgregor:
Draft

### DIFF
--- a/site/gatsby-site/src/components/landing/QuickSearch.js
+++ b/site/gatsby-site/src/components/landing/QuickSearch.js
@@ -14,11 +14,11 @@ export default function QuickSearch() {
     const width = window.innerWidth;
 
     if (width >= 350 && width < 450) {
-      setSearchPlaceholder(t('Search 2000+ reports'));
+      setSearchPlaceholder(t('Search 2500+ reports'));
     } else if (width >= 450 && width < 500) {
-      setSearchPlaceholder(t('Search 2000+ AI harm reports'));
+      setSearchPlaceholder(t('Search 2500+ AI harm reports'));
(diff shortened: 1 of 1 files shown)


smcgregor:
@cesarvarela this build error has presented twice now. Any idea what it is?

cesarvarela:
Yeah, we mongorestored staging, builds work fine now.

smcgregor:
got it, I hit retry here.

site/gatsby-site/src/components/landing/QuickSearch.js
```diff
@@ -14,11 +14,11 @@ export default function QuickSearch() {
     const width = window.innerWidth;
 
     if (width >= 350 && width < 450) {
-      setSearchPlaceholder(t('Search 2000+ reports'));
+      setSearchPlaceholder(t('Search 2500+ reports'));
```
smcgregor, line 17: @cesarvarela this wasn't translating so I added keys to the locales. However, it is still not translating. Is `landing.json` not the proper locale for QuickSearch?

cesarvarela: If you look at a few lines below, it is fetching translations from any of these three:
```
  const { t } = useTranslation(['translation', 'landing', 'actions']);
```
It should work 🤔 though I don't like that it is being used before being declared.

smcgregor: Hmm, you are thinking reordering the declaration might fix it?

cesarvarela: yes

Issue #1840, opened 2023-04-06
Spatial View - selected incident has black background


lmcnulty:
If you click the "Similar Incidents" header on the citation page, it takes you to the spatial visualization with the incident in question highlighted. Right now it's showing up with a black background.

![image](https://user-images.githubusercontent.com/25443411/230429678-b5e6ac40-dec2-4b99-888d-a770d6f94570.png)

Issue #1838, opened 2023-04-06
Spatial visualization classification selector does not match initial selection


lmcnulty:
Starts at "CSET:Near Miss" when "CSET:Sector of Deployment" is displayed

![image](https://user-images.githubusercontent.com/25443411/230427538-b1e8bf38-2c03-441f-8d05-8e7af4b6a126.png)

PR #1876, opened 2023-04-14
Fix spatial view initial view and selected color

lmcnulty:
Resolves https://github.com/responsible-ai-collaborative/aiid/issues/1840 and https://github.com/responsible-ai-collaborative/aiid/issues/1838.


lmcnulty:
This has a tricky bug whereby if you go to `/summaries/spatial/`, then manually change the path to `/summaries/spatial/?incident=…`, the incident will get focused but not change color. It does get colored correctly if you reach the page from "Similar Incidents" link or include the url parameter on the initial page load. I don't think this is worth spending more time on now, so I'm marking this as ready for review.

PR #1880, opened 2023-04-17
Video autoplay

jaegeun1393:


cesarvarela:
It looks like it's being addressed by #2143

PR #1885, opened 2023-04-17
Test tool to compare CSET classifications

### LINKED ISSUE #1879: CSET Taxonomy Consistency Check Tool
CSET has multiple people that annotate incidents independently of one another in copies of their new taxonomy. Once the annotators reach a reasonable degree of consistency between each other, they no long need to have as many checks of their annotations and the extra taxonomies are no longer used. However, CSET _does_ need a tool that will help them identify when there are inconsistencies. When they identify inconsistencies talk with one another and resolve the differences.

While the ideal tool is one that compares each instance of the CSET taxonomies and outputs only those cases where there are differences, this is also a pretty big implementation effort since it would require doing a bunch of things like sorting and matching tags, filtering fields that are full text and should not match, etc. All this work is likely to be one-off and not serve any general case. I am going to set this option aside and consider the following options.

**Add to Incident Table View**

A better approach is likely to just make it easy to look at the classification columns for each of the taxonomy replicates and visually check them for differences. The CSET columns could be collated and displayed on the current incident table view.

**New Extensible Table View**

Another way to table inspect classification consistency would be to build a table view for the taxonomies. Specifically, a table view where you can turn on/off columns across all taxonomies for each incident. CSET would then be able to visually inspect collections of columns for consistency. As an added benefit, it would be possible compose views of the tabular data related to incidents. We could (eventually) put the column configurations in the query string of the page to save state.

Why is this better than adding to the current table? We can ship it faster while it is still messy since CSET can use this now. We can then tighten it down, then eventually deprecate the old table.

**Hook tableau or similar into API**

Direct export to highly dynamic table view

**Export to CSV**

Easy enough...

**Priorities?**

I think we do the extensible table view if we can get it into prod within a week without displaying it to the public. Otherwise the CSV export is likely the way to go.

cesarvarela:
closes #1879

### DIFF
--- a/site/gatsby-site/cypress/e2e/integration/apps/csettool.cy.js
+++ b/site/gatsby-site/cypress/e2e/integration/apps/csettool.cy.js
@@ -1,0 +1,26 @@
+import { maybeIt } from '../../../support/utils';
+import cssettool from '../../../fixtures/classifications/cssettool.json';
+
+describe('CSET tool', () => {
+  const url = '/apps/csettool/52/';
+
+  maybeIt('Successfully loads CSET annotator classifications', () => {
+    cy.login(Cypress.env('e2eUsername'), Cypress.env('e2ePassword'));
+
+    cy.conditionalIntercept(
+      '**/graphql',
+      (req) => req.body.operationName == 'FindClassifications',
+      'FindClassifications',
+      cssettool
+    );
+
+    cy.visit(url);
+
+    cy.wait(['@FindClassifications']);
+
+    cy.waitForStableDOM();
+
+    cy.get('tbody > tr').should('have.length', 62);
+    cy.get('thead th').should('have.length', 3);
+  });
+});
--- a/site/gatsby-site/cypress/fixtures/classifications/cssettool.json
+++ b/site/gatsby-site/cypress/fixtures/classifications/cssettool.json
@@ -1,0 +1,636 @@
+{
+    "data": {
+        "classifications": [
+            {
+                "__typename": "Classification",
+                "_id": "6422fb4ad6ebced21049f7a2",
+                "attributes": [
+                    {
+                        "__typename": "ClassificationAttribute",
+                        "short_name": "Physical Objects",
+                        "value_json": "\"yes\""
+                    },
+                    {
+                        "__typename": "ClassificationAttribute",
+                        "short_name": "Entertainment Industry",
+                        "value_json": "\"no\""
+                    },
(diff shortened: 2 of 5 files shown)


cesarvarela:
@smcgregor quick draft:

https://deploy-preview-1885--staging-aiid.netlify.app/apps/csettool/52

The path is `/apps/csettool/{incident_id}.`

If you are okay with this direction, I've been thinking about adding next:

- custom cell renders for complex attributes like the entities list
- highlight rows with differing values

smcgregor:
Thanks @cesarvarela. Let's ship this to CSET and see if there are any high-importance asks and engineer as needed there. We can return to the more general case later. Can you get this mergeable to prod as a "secret menu" item? Maybe a single smoke test and review of code quality. Functionality can advance later.

smcgregor:
I received the alert below in email. Probably related?


Query Targeting: Scanned Objects / Returned has gone above 1000
--
 
The ratio of documents scanned to returned exceeded 1000.0 on  aiiddev-shard-00-01.seam4.mongodb.net, which typically suggests that un-indexed queries are being run. To help identify which query is problematic, we recommend navigating to the Query Profiler tool within Atlas. Read more about the Profiler here.

Query Targeting: Scanned Objects / Returned has gone above 1000
 
The ratio of documents scanned to returned exceeded 1000.0 on [ aiiddev-shard-00-01.seam4.mongodb.net](https://cloud.mongodb.com/email/link/643dcbe4403af9266ef48535/viewHostMetricsStatus?url=%2Fv2%2F6153eea15b27c6494839aa4e%23%2Fmetrics%2Fhost%2F417b90db86ee12b56fa0b75b475cd87d%2Fstatus), which typically suggests that un-indexed queries are being run. To help identify which query is problematic, we recommend navigating to the [Query Profiler](https://cloud.mongodb.com/email/link/643dcbe4403af9266ef48535/viewHostMetricsProfiler?url=%2Fv2%2F6153eea15b27c6494839aa4e%23%2Fmetrics%2Fhost%2F417b90db86ee12b56fa0b75b475cd87d%2Fprofiler) tool within Atlas. Read more about the Profiler [here](https://docs.atlas.mongodb.com/tutorial/profile-database/index.html).

cesarvarela:
I think it is the `candidates` collection:

image

cesarvarela:
@smcgregor added a button that points to the table in the citation page:
image

It is only visible to users with the `taxonomy_editor` role.

PR #1887, opened 2023-04-18
Remove CSET publish field

### LINKED ISSUE #1883: Duplicate `Publish` on CSET taxonomy form
image

lmcnulty:
`publish` is now a top-level member of all classifications, so the CSET-specific one is not necessary. Resolves #1883.

### DIFF
--- a/site/gatsby-site/migrations/2023.04.17T23.15.19.remove-cset-publish-field.js
+++ b/site/gatsby-site/migrations/2023.04.17T23.15.19.remove-cset-publish-field.js
@@ -1,0 +1,56 @@
+const config = require('../config');
+
+/** @type {import('umzug').MigrationFn<any>} */
+exports.up = async ({ context: { client } }) => {
+  // We're getting duplicate publish fields because
+  // publish was a field of the CSET taxonomy,
+  // but it has been upgraded to a top-level member of all taxonomies.
+  // To remove the duplicate we need to delete the original publish field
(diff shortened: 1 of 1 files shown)


site/gatsby-site/migrations/2023.04.17T23.15.19.remove-cset-publish-field.js
```diff
@@ -0,0 +1,56 @@
...
+    const attributePublish = classification.attributes.find(
+      (attribute) => attribute.short_name == 'Publish'
+    );
+
+    const attributePublishValue =
+      attributePublish?.value_json && JSON.parse(attributePublish.value_json);
+
+    // Ensure that we don't have a situation where one is explicitly set to true
+    // and the other explicitly set to false.
+    console.assert(Boolean(classification.publish) == Boolean(attributePublishValue));
```
cesarvarela, line 40: To be precise, this doesn't ensure what you describe. It only prints an error message, but the next sentences still execute. I don't know if that's what you want here.

cesarvarela (commented):
One question before merging.

site/gatsby-site/migrations/2023.04.17T23.15.19.remove-cset-publish-field.js
lmcnulty: I don't think this condition will actually be met, but I changed it to throw an exception instead.

PR #1898, opened 2023-04-19
Restore outline in blog

### LINKED ISSUE #1872: Blog outline is gone
(again)

image

clari182:
- closes #1872


lmcnulty (commented):
So it looks like the issue was that the outline was rendering but not visible. Can you update [the test we have on this](https://github.com/responsible-ai-collaborative/aiid/blob/master/site/gatsby-site/cypress/e2e/integration/blog.cy.js#L13) to check that it's visible?

clari182:
@lmcnulty so it looks like the tests were checking that the outline existed and it did, but it wasn't visible. So I updated the test to check if it's visible in a bigger viewport and hidden on smaller ones. Ready for another review!

lmcnulty (approved):

Issue #1893, opened 2023-04-18
Deployment Issues


smcgregor:
These are the issues I turned up when testing staging. None of these are blocking, but they are bugs that are in prod and have a high or critical priority.

1. The [classifications view](https://staging-aiid.netlify.app/apps/classifications/) is broken when you change the taxonomy to a taxonomy that is not the default. It presents a bunch of empty columns for each of the taxonomies.
2. One of the links on the [new blog post](https://staging-aiid.netlify.app/blog/incident-report-2023-march/) goes to `https://staging-aiid.netlify.app/cite/354/%23r2903`, which is a dead link.

PR #1915, opened 2023-04-20
Fix broken taxonomy switching on /apps/classifications

lmcnulty:
Resolves item 1 of #1893.


cesarvarela (changes requested):
Please add some tests!

cesarvarela (approved):

PR #1928, opened 2023-04-24
Add indexes as recommended by mongodb profiler

### LINKED ISSUE #1927: Create database indexes as migrations
The Mongo db cloud profiler suggests creating a few indexes, and although we could add them through the UI, we should do it in migration so it gets reflected on every env.

image

cesarvarela:
closes #1927


pdcp1 (approved):
It looks good to me

Issue #1814, opened 2023-04-03
Autocompletion on Incident ID or the Title of the Incident in Submission form


Janetbananet:
This function is already available in the "edit report" form under the "Incident IDs" field (see screen shot below). This should also be available in the submission. Note that the submission form should only allow for one incident ID.

![image](https://user-images.githubusercontent.com/28989465/229572574-3312858d-f831-4f43-aa46-4381e943c53b.png)

Janetbananet:
This capability is also needed in the "ingest form" (aka Edit Submission form), but in that case we DO want the ability to allow for multiple incident IDs. This is a separate GitHub issue #1822

PR #1929, opened 2023-04-24
Multiple Incidents linking on submit forms

### LINKED ISSUE #1822: Multiple incident association to a report in the ingest form
Add the capability to associate submissions with multiple incidents in the ingest form (aka "edit submission" form) and add autocompletion on incident ID or the title of the incident.

Below are two screen shots. The first shows which form specifically needs these changes, and the second shows the multiple incident association and autocomplete from the "edit report" form which already has both of these capabilities built-in.

Screen Shot 1:
image

Screen Shot 2:
image

cesarvarela:
closes #1822 and #1814

This can be seen working here:  https://github.com/cesarvarela/aiid/pull/45

### DIFF
--- a/site/gatsby-site/src/components/forms/SubmissionWizard/StepOne.js
+++ b/site/gatsby-site/src/components/forms/SubmissionWizard/StepOne.js
@@ -350,7 +350,11 @@ const FormDetails = ({
               <Label popover="incident_id" label={t('Incident IDs')} />
             </div>
             <div className="mt-1">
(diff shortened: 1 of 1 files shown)


clari182 (changes requested):
Looks good! 
I'd only add a placeholder on the Incident IDs field otherwise it's the only one without one.
image

I'd also change the wording a little bit of the tooltip to let the user know it can be linked to multiple incidents
image

PR #1932, opened 2023-04-25
Add last page button to Table component

### LINKED ISSUE #1848: Table View – Reinstate "show last page" button
The current button layout doesn't include a way to quickly jump to the last page, which is used most often to rectify newly ingested incident data (appearing on the last page).

**Proposed: add a `Last` button next to `Next`**

image

cesarvarela:
closes #1848


clari182:
Looks good! However I'd add a `first` also. Because when there are a lot of pages and you go to the last, if you want to go back to the beginning fast you need to reload.

clari182 (approved):

Issue #1643, opened 2023-02-23
UX Report from Responsible AI professional


smcgregor:
I had a conversation yesterday indicating something to the effect of:

1. "I didn't know the discover app supports filters". I think doing something to make it slightly more prominent would be good. Maybe adding the filter icon would be enough.
2. "The headline link on the citation page just scrolls -- it doesn't link to the content itself. I expected it to take me to the source publication." Some options on changes: Go to the article as hosted by the source, expand/collapse the card, produce individual pages for every report and link to that, remove the link entirely, or something I have not thought of

Janetbananet:
@smcgregor I'm curious about #2. Is this the photo carousel that is just above where all the Incident Reports are listed?

smcgregor:
No, click on the headline of the card presenting the report.

kepae:
Bumping this for potential frontend improvements, @lmcnulty.

Looks like there's a PR that could still work? Screenshot results/deploy preview would also help a quick review.
(looks like there may have been a test flake on the last deploy. hopefully better now.)

PR #1936, opened 2023-04-25
Move filters to sidebar

lmcnulty:
Making filters more prominent per https://github.com/responsible-ai-collaborative/aiid/issues/1643.


lmcnulty:
I would still like to consider doing this sometime but this PR is stale.

PR #1942, opened 2023-04-27
Submit/fallback image warning

### LINKED ISSUE #1799: Rendering fallback image in the submission form
Background: Currently, when a user is entering an image URL into the submission form, the image preview automatically shows the fallback image if the URL from the original source does is not able to load. This will cause people to believe that the URL worked even in cases when it did not. 

Change needed: do not render the fallback image on the submission form unless the user elects to display the fallback by default.

clari182:
- closes #1799 

Added a red message under the URL input for images with invalid URLs.

- Submit form
- Submission form (review incident)


site/gatsby-site/src/components/forms/PreviewImageInputGroup.js
```diff
@@ -57,6 +63,11 @@ export default function PreviewImageInputGroup({
           schema,
         }}
       />
+      {imageLoadedFailed && (
+        <span className="text-sm text-red-600">
```
lmcnulty, line 67: It seems to show up even when nothing has been entered. I would add a check that the field has been touched.

Elsewhere when we have red error text, we also put a red border on the input box.

clari182: I added the touched. So now the warning is displayed only after an invalid URL is added.

However, I thought about adding the red border to the input but I think it looks too much like an error instead of a warning. 
I think it would make the user think they are required to enter a valid URL.

site/gatsby-site/src/components/forms/PreviewImageInputGroup.js
```diff
@@ -40,6 +44,10 @@ export default function PreviewImageInputGroup({
     }
   }, [cloudinary_id, values.cloudinary_id]);
 
+  const onImageLoaded = (failed) => {
```
cesarvarela, line 47: I think you can use the `setImageLoadFailed` method instead.

site/gatsby-site/src/components/forms/PreviewImageInputGroup.js
```diff
@@ -83,6 +97,7 @@ export default function PreviewImageInputGroup({
               alt={alt}
               height="300"
               className="inline-block mx-auto min-h-48 min-w-48 max-w-full h-full bg-white"
+              onImageLoaded={onImageLoaded}
```
cesarvarela, line 100: here

cesarvarela:
@clari182 it is conflicts time!

PR #1943, opened 2023-04-28
Replace wordcounts with email form

lmcnulty:


Janetbananet:
@smcgregor What do you think about titling the newsletter something more descriptive/snazzy? I'm thinking "The AI Incident Briefing", but open. I think the "AIID Newsletter" isn't quite right because currently it's not in a newsletter format (especially given that we do not currently have an email marketing client or design) and "AIID" may not be the most descriptive acronym depending on who the audience is.

lmcnulty:
@Janetbananet I didn't update the screenshot, but it currently does say "The AI Incident Briefing". You can check the most recent version from the [Deploy Preview](https://deploy-preview-1943--staging-aiid.netlify.app/).

image.

smcgregor:
Looks great.

Nit: `Subscribe to the AI Incident Briefing and get incident round-ups delivered straight to your inbox, and major organizational updates.` -> `Subscribe to the AI Incident Briefing and get monthly incident round-ups along with occasional major database updates.`

lmcnulty:
@smcgregor @Janetbananet

Does anyone have thoughts about what to show here when a user is already logged in? 

My immediate thought is to either show something like "Thanks for subscribing!" or offer the option to re-subscribe if the user has previously unsubscribed. However, having looked into the history of this feature, it seems that adding major update subscribers [is done manually](https://github.com/responsible-ai-collaborative/aiid/issues/1830)? In that case, I don't think we can programmatically access a user's subscription status. I'm also not sure how a user could re-subscribe from the site.

smcgregor:
I am comfortable on this one not making the subscribe form conditional, at least for now.

lmcnulty:
If we show input field when the user is logged in, if they try to subscribe with the same email, it shows an error, which I'd like to avoid. How about "check your inbox for..." ? That indicates that you were once subscribed but doesn't necessary imply that you're currently subscribed and doesn't require any complicated logic.

smcgregor:
Sounds good. One tweak, before the first comma, "You are currently subscribed to the AI Incident Briefing,"

lmcnulty:
The user can unsubscribe via Google Groups, so it may not be true that a logged in user is currently subscribed. However, they will get at least one email before they unsubscribe, so they can still "check their inbox" for it whether they are currently subscribed or not.

cesarvarela (commented):
@lmcnulty please merge staging, it might fix the deploy preview

PR #1944, opened 2023-04-28
Feature/incidents review table

### LINKED ISSUE #1785: Convert submission queue to a sortable and filterable table

### LINKED ISSUE #1786: Completion submission completion status
We need a way to quickly determine the completeness of a submission in the review queue. This may be judged by the number of completed fields, but that needs more thought/discussion.

The "Quick Add URL's" should automatically be designated a 10% completion, and then be combined with the rest of the submissions so that there is only one central queue.

clari182:
- closes #1785 
- closes #1786 

https://deploy-preview-1944--staging-aiid.netlify.app/


clari182:
Hey @Janetbananet & @LKchemposer 
Would you mind taking a look at this PR and telling me what you think about the submission table as it is?

I'd like your input.
This changes the submission list on apps/submitted. It is now a table and I also added the Completion status on the first column.

Janetbananet:
A few comments from me, although it is crucial to get comments back from @LKchemposer since this could dramatically change his workflow. Depending on his thoughts, we may even want to have a video chat to talk through the design elements. I would recommend holding on more work on this PR @clari182 until we get feedback from Khoa. 

- Compress the table contents so that more of the rows/columns appear without having to scroll.
- Compress the table vertical height so that you can more easily access the horizontal scroll bar.
- I do not see a "review" or "claim" button in the table. Is that because the preview is showing the table if a user is not logged in?
- The "Quick Submissions" queue is still at the bottom of this page instead of integrated into the table.
- Personally, I prefer the UI design element from the original submission queue that shows only a few of the fields, and then expands the selection when you hit "review". Having all of the fields stretched horizontally across the table would be clunky to review.
- In general, I wonder if we could keep more of the original design elements of the submissions queue.

clari182:
@Janetbananet 

- The claim editor is not done yet, this PR is only for the table and the completion status (for now), the editor's claim will be done later.
- Maybe we could show only a few fields and then have a + button to expand the row and display additional fields. So it would expand to show more details. That way we could keep the table but also the previous design.
For that, the only thing I'd need is what fields you prefer to be displayed by default on the table before expanding.
They could be the same as now:
- Title
- Incident date
- Date published
- Date Submitted
- Authors

But I agree, I think it's too much to have all fields on the same row, I'll work on that once we have more comments from @LKchemposer.

LKchemposer:
Sorry about late reply. Will give detailed feedback today.

LKchemposer:
@clari182 not sure why, but the submissions queue page doesn't seem to load for me. do I need an account to access it?

clari182:
@LKchemposer sorry about that, it should work now!

clari182:
I'm closing this as we're going with a better approach discussed in https://github.com/responsible-ai-collaborative/aiid/issues/1785

PR #1933, opened 2023-04-25
Deploy to Production

cesarvarela:
- new Variants UX
- multiple incident linking on Submissions

PR #1953, opened 2023-05-03
Add Variant condition to `linkReportsToIncidents` function

pdcp1:
Don't display Variants on the Latest report component. Fix comment on https://github.com/responsible-ai-collaborative/aiid/pull/1933#issuecomment-1532470734

## Note

This issue will be fixed on Staging once the migration ran. So it won't be fixed on the preview link

### DIFF
--- a/site/realm/functions/linkReportsToIncidents.js
+++ b/site/realm/functions/linkReportsToIncidents.js
@@ -63,8 +63,9 @@ exports = async (input) => {
   await reportsCollection.updateMany(
     {
       report_number: { $in: input.report_numbers },
-      text_inputs: { $in: [null, ""] },
-      text_outputs: { $in: [null, ""] },
+      title: { $nin: [null, ""] },
+      url: { $nin: [null, ""] },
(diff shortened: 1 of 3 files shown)


site/realm/functions/linkReportsToIncidents.js
```diff
@@ -63,8 +63,9 @@ exports = async (input) => {
   await reportsCollection.updateMany(
     {
       report_number: { $in: input.report_numbers },
-      text_inputs: { $in: [null, ""] },
-      text_outputs: { $in: [null, ""] },
+      title: { $nin: [null, ""] },
+      url: { $nin: [null, ""] },
+      source_domain: { $nin: [null, ""] },
```
cesarvarela, line 68: Should this also take into account the new `inputs_outputs` field too?

pdcp1: Actually `inputs_outputs` field is not mandatory for a Variant. Actually, Reports can have a `inputs_outputs` field with some values and it is still a Report.

PR #1954, opened 2023-05-03
Fix similar incident cards on smaller resolutions

### LINKED ISSUE #1952: Broken similar incident cards on smaller resolutions
Small and intermediate resolutions cutoff the text.

Screen Shot 2023-05-02 at 8 37 45 PM

Screen Shot 2023-05-02 at 8 38 03 PM

lmcnulty:
Resolves #1952.

It's kind of easy to break things here because the parent styles are different in each place we use the component.


cesarvarela:
@lmcnulty please fix conflicts!

Issue #1961, opened 2023-05-05
Automatically subscribe the user to reports or incidents the user submitted or made changes to


pdcp1:
Part of issue #1783

PR #1976, opened 2023-05-09
Automatically subscribe the submitter to reports

pdcp1:
On top of this PR https://github.com/responsible-ai-collaborative/aiid/pull/1951
This close https://github.com/responsible-ai-collaborative/aiid/issues/1961


cesarvarela (commented):
I wonder about using `promoteSubmissionToReport` for more than what its name says it will do. I think we could do the subscription step in a separate function, but at the same time, I like the conciseness of only one function call that does everything, 

What do you think?

pdcp1:
> I wonder about using `promoteSubmissionToReport` for more than what its name says it will do. I think we could do the subscription step in a separate function, but at the same time, I like the conciseness of only one function call that does everything,
> 
> What do you think?

@cesarvarela We can add another function `subscribe` and call it from `promoteSubmissionToReport` code. Doing it in that way, we can call the `subscribe` function from another place and keep just one single call to `promoteSubmissionToReport`

site/gatsby-site/src/components/submissions/SubmissionReview.js
```diff
@@ -83,6 +87,29 @@ const SubmissionReview = ({ submission }) => {
       },
     });
 
+  const subscribeToNewReports = async (incident_id) => {
```
cesarvarela, line 90: Is there a rule that checks that only subscriptions of the current user are allowed?

pdcp1: Yes, on `customData/subscriptions/rules.json` we have a rule for that:
```
{
    "name": "Owner of the document",
    "apply_when": {
        "userId": "%%user.id"
    },
    "read": true,
    "write": true,
    "insert": true,
    "delete": true,
    "search": true
}
```

PR #1980, opened 2023-05-10
Feature/form local storage

### LINKED ISSUE #1677: Save form state in sessionStorage or localStorage
> It seems that prompting before switching to a different page is unsupported https://github.com/gatsbyjs/gatsby/issues/10726. Instead they recommend saving the form state in localStorage. That seems like the most correct solution, it will just be a bit more work.

_Originally posted by @lmcnulty in https://github.com/responsible-ai-collaborative/aiid/issues/1631#issuecomment-1438711485_

### LINKED ISSUE #1968: Clear Submit form state on submit

clari182:
- closes #1677 
- closes #1968


lmcnulty:
I think "your changes are saved" messaging in the indicator could cause users to confuse saving and submitting. I would use the word "draft" somewhere. The indicator also takes up more space at the top than I would like. What do you think of moving it to the header Google-Docs style?

image

site/gatsby-site/src/components/submissions/SubmissionWizard.js
```diff
@@ -78,13 +91,23 @@ const SubmissionWizard = ({ submitForm, initialValues, urlFromQueryString }) =>
 
         const cloudinary_id = getCloudinaryPublicID(news.image_url);
 
-        const newValues = {
+        let newValues = {
```
cesarvarela, line 94: Since the variable is not being reassigned, it is appropriate to retain the `const` keyword.

site/gatsby-site/src/components/forms/SubmitForm.js
```diff
@@ -101,27 +89,43 @@ const SubmitForm = () => {
   `);
 
   useEffect(() => {
-    if (!loading) {
-      const submission = { ...query, cloudinary_id: '' };
+    let submission = { ...query, cloudinary_id: '' };
```
cesarvarela, line 92: You can keep the `const` declaration here.

site/gatsby-site/cypress/e2e/integration/submit.cy.js
```diff
@@ -1546,4 +1581,102 @@ describe('The Submit form', () => {
 
     cy.get('@parseNewsSpy').should('have.been.calledOnce');
   });
+
+  it('Should load from localstorage', () => {
```
cesarvarela, line 1585: It would be helpful to have an additional test that simulates form filling and verifies that the `localStorage` data was correctly saved.

cesarvarela (changes requested):

site/gatsby-site/cypress/e2e/integration/submit.cy.js
cesarvarela: example here: https://github.com/responsible-ai-collaborative/aiid/pull/2007/files#diff-8bc19c4a77b6eaa334aa3a8cf9e1f4d0a055a3762ed36db56128bf073c9542b4R32

site/gatsby-site/src/components/forms/SubmitForm.js
clari182: I can't because I'm actually assigning the value on line 109

clari182:
@lmcnulty @cesarvarela ready for another review

cesarvarela (changes requested):
Two last comments before merging.

These forms are getting fun 🙃

site/gatsby-site/src/components/forms/SubmissionWizard/StepOne.js
```diff
@@ -130,22 +135,33 @@ const FormDetails = ({
     resetForm,
   } = useFormikContext();
 
+  const saveInLocalStorage = debounce((values) => {
```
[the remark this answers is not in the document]
cesarvarela: I'm saying that it is probably not doing what you describe because a new function is created on every render. You need to do something like this: https://github.com/responsible-ai-collaborative/aiid/blob/a935caccf588c83f8c7b6cce07f7aaa78a6df5ed/site/gatsby-site/src/components/taxa/TaxonomyForm.js#L36

clari182:
Got it @cesarvarela ! Fixed based on your comments. Ready for another review

cesarvarela (approved):

PR #1997, opened 2023-05-15
CSETv1 update fields 3.3 and 5.3 to "Clear link to Technology", field 9 wording

lmcnulty:
Per change requested by Mia.

3.3 and 5.3 are *not* supposed to be linked, so the casing is different between them.


cesarvarela (approved):
Is there any documentation about this?

I have no idea how we end up with a 2k lines migration.

It also looks like this didn't require any test update, which is a bit scary 😅

cesarvarela:
@lmcnulty I'm going to merge it, but it looks like we could have a test that fetches the CSET taxonomy and then check that the taxonomy form renders correctly and another one that tests that the taxonomy is renderer properly on the citation page. What do you think?

lmcnulty:
@cesarvarela Every time I need to update the taxonomy, I've just been replacing the taxonomy object in full. It seems less error prone than programmatically making changes to deeply nested objects, but the PR diff is less useful. The actual changes are [these](https://github.com/responsible-ai-collaborative/aiid/pull/1997/commits/dbf1aafafb32b51beb075d97052894b6dcab015a) two [commits](https://github.com/responsible-ai-collaborative/aiid/pull/1997/commits/dbf1aafafb32b51beb075d97052894b6dcab015a).

I agree, ideally these should be tested. I probably won't get to that until next week.

PR #2005, opened 2023-05-16
Deploy to Production

cesarvarela:


cesarvarela:
@smcgregor ready for review

smcgregor:
@pdcp1 it looks like this will automatically subscribe the editor to the submission they add to an incident? I didn't notice the subscription work landing. Can you also set the submitting user to subscribe if there is a submitting user?

Merging this now.

pdcp1:
@smcgregor This automatically subscribes the **Submitter** of the submission (if there is a submitter user) to the incident just created (or the existing incident associated with the new report).
Is that what you mean?

smcgregor:
Yes, it looks like I misread the code. Thanks!

PR #2014, opened 2023-05-18
Add missing local netlify server

cesarvarela:


cesarvarela:
FYI @pdcp1  8080 one is needed when tests run on Netlify

Issue #1949, opened 2023-05-02
Submit form crashes if an invalid date is entered


cesarvarela:
![image](https://user-images.githubusercontent.com/1172479/235774702-bb63d92b-57af-49a5-b816-b5cec1c128d9.png)

pdcp1:
Closing this since it was already merged into Staging

PR #2015, opened 2023-05-18
Only allow past dates on `Date Published` and `Date Downloaded`

pdcp1:
This close https://github.com/responsible-ai-collaborative/aiid/issues/1949

### DIFF
--- a/site/gatsby-site/src/utils/date.js
+++ b/site/gatsby-site/src/utils/date.js
@@ -1,3 +1,5 @@
+import { parse } from 'date-fns';
+
 // RegEx for date validation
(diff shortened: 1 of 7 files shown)


site/gatsby-site/src/utils/date.js
```diff
@@ -9,3 +11,14 @@ export const validateDate = (date) => {
   }
   return date;
 };
+
+export const isPastDate = {
+  message: '*Date must be in the past',
```
cesarvarela, line 16: Looks like the translation is missing here, but seems to be missing for other error messages too

PR #2016, opened 2023-05-18
Add Github action to restore Prod DB into staging

pdcp1:


pdcp1:
@cesarvarela I just added an additional step at the end of this workflow to trigger the Netlify Staging build after the DB is restored. Please review.

cesarvarela (commented):
I would use only one variable for the connection string as we do with `MONGODB_CONNECTION_STRING` and the like for consistency and simplicity.

cesarvarela (commented):
We should also make sure that it is not possible for someone to open a PR and access the connection strings 🤔

pdcp1:
> We should also make sure that it is not possible for someone to open a PR and access the connection strings 🤔

@cesarvarela This is ready for review again. Regarding your comment, if someone opens a PR, they can't access the Github secrets values, so we are OK

cesarvarela (commented):
Sorry to be picky, but calling both `_DUMP_` might be confusing because the staging one is also used for restoring :P

> this is ready for review again. Regarding your comment, if someone opens a PR, they can't access the GitHub secrets values, so we are OK

Yeah, but I can open a PR with a GitHub action that has a `console.log(SECRET)`  and get access to it. I think that Gihub checks that the user is already considered a contributor before allowing someone to do this, but if you want to check just to be safe please do.

pdcp1:
> Yeah, but I can open a PR with a GitHub action that has a `console.log(SECRET)` and get access to it. I think that Gihub checks that the user is already considered a contributor before allowing someone to do this, but if you want to check just to be safe please do.

@cesarvarela 
Yeah, the `DUMP` was a copy-paste issue, I just renamed both.
Regarding the Secret value, If anyone opens a new PR, they can only access their Secret environment values, not the `responsible-ai-collaborative` repo secrets values. It can access it only if we merged it into Staging, but it is our responsibility to check a malicious code.
I don't know if we are on the same page. Let me know if I misunderstood your question.

PR #2022, opened 2023-05-19
Deploy to Production

cesarvarela:


cesarvarela:
@smcgregor ready for review

smcgregor:
I am not sure which, if any, of these should block deployment versus which need to be addressed on a timely basis.

* When signing up for a new account, I didn't see a toast after the redirect to the landing page. The user will not know to look in their email.
* After confirming the email, there is nothing telling them where to go next or what they can do with their new account. Since confirming the email does not login, they likely need to be told to login or do a toast on the login page...
* I did not see the "enter your name" screen at any time
* The "enter your name" screen includes the "subscriber" role stated for the user to see. This might be a UX issue since the user has no context for what roles are and whether they should know about them.

PR #2030, opened 2023-05-24
Cset tool updates

### LINKED ISSUE #2023: Support CSETv1 Publication
Engineering work to support CSET publication:

- Write a migration merging the CSETv1 annotator classifications such that...

  - If `CSETv1_Annotator-1` and `CSETv1_Annotator-2` agree,
    the matching value gets copied to `CSETv1`

  - If there's a missing and a non-missing field,
    use the non-missing one.

  - For notes fields, both annotators' texts 
    get concatenated with a textual label. E.g.
    ```
    Annotator 1: Lorem ipsum...

    Annotator 2: Twas brillig...
    ```
  - For array fields, the arrays get merged with duplicates removed.
  - For entities fields, the lists of entities get merged.

- Move the button to delete an entity into the entity header.

- In the comparison tool, color highlight rows
  where the annotators' values don't match.

cesarvarela:
closes #2023

### DIFF
--- a/site/gatsby-site/src/components/classifications/CsetTable.js
+++ b/site/gatsby-site/src/components/classifications/CsetTable.js
@@ -1,7 +1,7 @@
 import React, { useMemo } from 'react';
 import { useFilters, usePagination, useSortBy, useTable } from 'react-table';
 import Table, { DefaultColumnFilter, DefaultColumnHeader } from 'components/ui/Table';
-import { startsWith, uniq } from 'lodash';
+import { startsWith, union, uniqWith, isEqual, filter } from 'lodash';
 
 function ValueCell({ cell, ...props }) {
   const {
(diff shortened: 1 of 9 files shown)


site/gatsby-site/src/components/classifications/CsetTable.js
```diff
@@ -40,12 +292,129 @@ export default function CsetTable({ data, className = '', ...props }) {
...
+      addToast({
+        message: 'Classification updated',
+        severity: SEVERITY.success,
+      });
+
+      setSubmitting(false);
+    } catch (error) {
+      addToast({
+        message: 'Classification updated',
+        severity: SEVERITY.danger,
```
lmcnulty, line 401: I think the message should be different for a failure.

lmcnulty (commented):
Some of the metadata fields are likely to change after merge. We should probably exclude them:

image


It seems like false + true is resolving to true:

image

---

I think we were planning to make this tool read-only and do edits in the  form. I'm a little concerned about the user making changes in both places at once, and then having data overwritten. But this looks good enough that I think we should talk with Mia or Heather about it again.

cesarvarela:
> Some of the metadata fields are likely to change after merge. We should probably exclude them:

What do you mean by that, should I skip those listed in the first pic?

@lmcnulty

lmcnulty:
> What do you mean by that, should I skip those listed in the first pic?

Yes. Sorry for being unclear.

cesarvarela:
@lmcnulty @smcgregor updated, any last words before merging?

smcgregor:
Looking great. We might want to have the entity delete on the merged records, but I think we should get this into prod and get feedback rather than make any more changes.

PR #2052, opened 2023-05-30
Deploy to Production

cesarvarela:
To get the workflow working these variables need to be added to the GitHub staging environment:
```
DB_PRODUCTION_CONNECTION_STRING=[Production connection string with readonly user credentials. ie: mongodb+srv://[DB readonly user]:[DB user password]@aiiddev-xxxxxx.gcp.mongodb.net]
DB_STAGING_CONNECTION_STRING=[Staging connection string with admin user credentials. ie: mongodb+srv://[DB admin user]:[DB user password]@aiiddev-xxxxxx.gcp.mongodb.net]
NETLIFY_BUILD_STAGING_URL=[Netlify Staging build hook. This value is on https://app.netlify.com/sites/staging-aiid/settings/deploys#continuous-deployment]
```

### DIFF
--- a/.github/workflows/mongodb-clone.yml
+++ b/.github/workflows/mongodb-clone.yml
@@ -1,0 +1,29 @@
+name: Restore Prod DB into Staging
+on: [workflow_dispatch]
+jobs:
+  mongo-clone-action:
+    runs-on: ubuntu-latest
+    name: Restore Prod DB into Staging
+    steps:
+      - name: Install Mongo Tools
+        run : |
+          wget -qO - https://www.mongodb.org/static/pgp/server-6.0.asc |  gpg --dearmor | sudo tee /usr/share/keyrings/mongodb.gpg > /dev/null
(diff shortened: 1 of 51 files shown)


cesarvarela:
@pdcp1 we already have a few environment variables that change based on the branch:

image

Maybe call them something like: `MONGODB_RESTORE_CONNECTION_STRING` to avoid confusion? Or should we delete them and reuse the same two for everything?

@smcgregor, what do you like more? A connection string per functionality or a couple used everywhere?

smcgregor:
Connection string per auth level maybe?

smcgregor:
I am having trouble identifying what env changes are needed on this one. Are they only needed if we hit the restore button?

smcgregor:
@clari182 I think there is a plain text attribute you could put in the feed rather than the Markdown. The markdown won't render properly in readers.

I still need to play with the submit form and understand the environment variables, but am otherwise 👍

cesarvarela:
@smcgregor Yes only needed to get the restore button working. Nothing will break if not set.

smcgregor:
Submit form worked fine for me, including the local storage drafts. :)

I don't want to merge this until the RSS feed is updated to the plain text or we decide we are not doing that. @clari182 thoughts?

PR #2070, opened 2023-06-06
Feature/new submission list

### LINKED ISSUE #1785: Convert submission queue to a sortable and filterable table

### LINKED ISSUE #1787: Submission Queue UX Changes
Currently, there is no way to interact with the submissions review queue beyond approve/reject. With the planned increase in submissions and Editors working with them, we need capabilities to sort/group/filter and coordinate Editing efforts in the queue.

Issues (Stories) so far include:

- [x] #1785 
- [ ] #1786
- [x] #1809 1809

### LINKED ISSUE #2164: Full screen editor under left menu
When using the editor full screen, the left menu is not placed under it.
image

clari182:
https://deploy-preview-17--clarayoudale-staging.netlify.app/

- closes #1787 
- closes #1785 
- closes #2164 

Fixes needed:

- [x] Remove actions filtering on table
- [x] Fix Status filtering
- [x] Completion color coding on review page
- [x] Fix submitters sort
- [x] Update from latest changes in staging: editor will now be incident_editors (multiple)
    - [x] This implies having the dropdown with multi selection
    - [x] Remove editor calls from DB
    - [x] Remove editor from schema 
- [x] Anonymous user sees empty/malformed column
- [x] Also, suspicious null null user appears in editor filter. I suspect this is an account without first/last name information. If that's the case, we shouldn't worry about trying to select those users (selecting "all" should suffice).
- [x] Some tags don't stack vertically but aligned horizontally, which renders poorly when there are narrow columns or many entities. In this case, the "submitters"
- [x] Improve color codes for status
- [x] Remove that extra space between Editors, Status, and Add as new sections
- [x] Fix mobile layout
- [x] Fix skipped test
- [x] Divide submissionEdit file into 2: submissionEdit and submissionEditForm


pdcp1:
@clari182 Can you update this PR with the latest version from Staging and generate a preview link to review? We would like to check if everything works as expected and if the changes improves the UX

kepae:
I took the preview for a spin and enjoyed the improvements greatly!! I also had fun trying to break it. :-)
Some issues:

1) Anonymous user sees empty/malformed column.
2) Also, suspicious `null null` user appears in editor filter. I suspect this is an account without first/last name information. If that's the case, we shouldn't worry about trying to select those users (selecting "all" should suffice).

Screenshot 2023-09-07 at 3 29 13 PM

3) Some tags don't stack vertically but aligned horizontally, which renders poorly when there are narrow columns or many entities. In this case, the "submitters"

Screenshot 2023-09-07 at 3 26 08 PM


4) Open question: are there alternative colors or suggestions for the "in review" badge? Green means "all is good" but is correlated with "success". In this UI, something is under review and not "succeeded" yet. In a sense, we have two different types of "pending" (pending and in review) and I'm wondering if we can have colors that are distinct but not blue or red/"failure". Thoughts?

pdcp1:
I think we should improve the layout of the right-side controls on the Submission review page. Maybe removing that extra space between `Editors`, `Status`, and `Add as new` sections

image

pdcp1:
The layout of the mobile version of the Submission review page is broken:

image

site/gatsby-site/src/graphql/users.js
```diff
@@ -45,6 +45,23 @@ export const FIND_USER = gql`
...
+      roles
+      userId
+      first_name
+      last_name
+      adminData {
+        email
+        disabled
+        creationDate
+        lastAuthenticationDate
+      }
```
pdcp1, line 60: Is this query fast enough? Without the filter by role, this query takes ~1 minute to complete, but I want to confirm if the query is faster in this case.

site/gatsby-site/src/components/submissions/SubmissionForm.js
```diff
@@ -366,14 +366,16 @@ const SubmissionForm = () => {
...
+            {!hideEditors && (
+              <UsersInputGroup
+                name="incident_editors"
+                label={t('Editors')}
+                icon={faPenNib}
+                className="mt-3"
+                placeholder={t('Editors')}
+                {...TextInputGroupProps}
+              />
+            )}
```
pdcp1, line 378: I'm just curious when we hide this user's input.

pdcp1 (changes requested):

clari182:
@kepae regarding colors for the `In Review` and `Pending Review` badges, what do you think about this option:
image

kepae:
re: badge colors: LGTM!

site/gatsby-site/src/graphql/users.js
clari182: @pdcp1 the query takes a few seconds to execute. Do you think there is a better way to fetch all editors?

pdcp1: If it takes a few seconds is OK

pdcp1 (approved):

kepae (approved):
I'll merge this next week -- there's already a few changes in staging I'd like to push for tomorrow, and this is a big change I'd like to test in staging more. :-)

cesarvarela (commented):
Is this using the default Table component we are all around the site?

I'm not a fan of resizing columns because they are too easy to break (I mean the current react-table implementation, if there is an implementation that works well somewhere else I'm all for it):

image

clari182:
@cesarvarela Yes, it's using the Table component you implemented, I just added the possibility of resizing. It wasn't a requirement at first but it became one from the editors. Let me know if I should keep it.

cesarvarela:
@clari182 let's keep it yes! We might find a better implementation later.

site/gatsby-site/src/pages/apps/submitted.js
```diff
@@ -79,78 +90,101 @@ const SubmittedIncidentsPage = ({ ...props }) => {
...
-                        fill="currentColor"
-                        viewBox="0 0 20 20"
-                        xmlns="http://www.w3.org/2000/svg"
+      {pageLoading ? (
+        <ListSkeleton />
+      ) : (
+        <>
+          {id ? (
+            <>
+              <AiidHelmet path={props.location.pathname}>
```
cesarvarela, line 99: I think the <helmet> should be outside the `pageLoading` conditional. Otherwise, the initial HTML will render with no title.

kepae:
@clari182 @cesarvarela thoughts on pushing to staging today?

cesarvarela:
@kepae works for me.

PR #2073, opened 2023-06-07
Deploy to Production

cesarvarela:
A `GATSBY_ROLLBAR_TOKEN` env variable needs to be created before merging

### DIFF
--- a/README.md
+++ b/README.md
@@ -561,7 +561,23 @@ The endpoint is implemented as a Gatsby function. In the context where this func
 ```
 REALM_GRAPHQL_API_KEY=xxxxxxxxxx
 ```
+You can generate a new key following these steps:
+1. Go to your Realm App
+2. Go to `Authentication`
+3. Go to `Authentication Providers`
+4. Go to `API keys`
+5. Click on `Create API key` and copy the key value
+
 About Realm API Keys: https://www.mongodb.com/docs/realm/authentication/api-key/
+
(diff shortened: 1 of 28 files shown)


cesarvarela:
@smcgregor can you try to fix conflicts? It is not picking up my latest merge attempting to fix them.

smcgregor:
I can take a look, can you give me more context and details?

cesarvarela:
@smcgregor I think this is some bug with GitHub's diffing getting confused; it should be easy to fix from the web UI.

It is only this line:

image

(we should remove that test or use an entity that changes less often, but the conflicts need to be fixed anyway)

smcgregor:
ok, so I should update my local copy of staging and master, merge master to staging, then push directly to staging?

cesarvarela:
yes exactly that

smcgregor:
Done.

cesarvarela:
@smcgregor ready for review

Among other things, it contains fixes for the current production deploy and CSET tool improvements.

smcgregor:
@cesarvarela will this work if the rollbar token is not set? We need to announce this a bit louder, if so.

Created: https://github.com/responsible-ai-collaborative/aiid/issues/2081

I have not tested the CSET tool updates, but I will ask CSET to give it a test after this merges.

cesarvarela:
@smcgregor it will, it just won't track errors.

PR #2102, opened 2023-06-19
Feature/reports incidents schema

### LINKED ISSUE #1035: Create a Gatsby schema customization to link Reports with Incidents
This will simplify querying Incidents and their Reports' data.

```
diff --git a/site/gatsby-site/gatsby-node.js b/site/gatsby-site/gatsby-node.js
index 924421bc..657d4f6c 100644
--- a/site/gatsby-site/gatsby-node.js
+++ b/site/gatsby-site/gatsby-node.js
@@ -190,6 +190,7 @@ exports.createSchemaCustomization = ({ actions }) => {
       editor_similar_incidents: [Int]
       editor_dissimilar_incidents: [Int]
       flagged_dissimilar_incidents: [Int]
+      reports: [mongodbAiidprodReports] @link(by: "report_number")
     }

     type mongodbAiidprodSubmissions implements Node {
@@ -203,6 +204,7 @@ exports.createSchemaCustomization = ({ actions }) => {
       tags: [String]
       plain_text: String
       embedding: reportEmbedding
+      report_number: [Int]
     }

     type mongodbAiidprodTaxaField_list implements Node {
```

clari182:
- closes #1035 

It also uses this fix to display the latest 5 reports from 5 different incidents.
There is also a layout styling fix for edit incident page (width was too small).

### DIFF
--- a/site/gatsby-site/gatsby-config.js
+++ b/site/gatsby-site/gatsby-config.js
@@ -137,7 +137,7 @@ const plugins = [
 
               const matchingIncidents = allMongodbAiidprodIncidents.edges.find((incident) => {
                 return incident.node.reports.find((report) => {
-                  return report === report_number;
+                  return report.report_number === report_number;
                 });
               });
 
@@ -195,7 +195,9 @@ const plugins = [
               edges {
                 node {
                   incident_id
-                  reports
+                  reports { 
+                    report_number
+                  }
                 }
               }
             }
--- a/site/gatsby-site/src/components/landing/RandomIncidentsCarousel.js
+++ b/site/gatsby-site/src/components/landing/RandomIncidentsCarousel.js
@@ -16,7 +16,9 @@ const RandomIncidentsCarousel = () => {
           allMongodbAiidprodIncidents {
             nodes {
               incident_id
-              reports
+              reports {
+                report_number
+              }
               title
             }
           }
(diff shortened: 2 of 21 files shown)


site/gatsby-site/gatsby-config.js
```diff
@@ -137,7 +137,7 @@ const plugins = [
 
               const matchingIncidents = allMongodbAiidprodIncidents.edges.find((incident) => {
                 return incident.node.reports.find((report) => {
-                  return report === report_number;
+                  return report.report_number === report_number;
```
cesarvarela, line 140: This should use the new nested reports query instead of matching like this.

site/gatsby-site/src/components/landing/RandomIncidentsCarousel.js
```diff
@@ -15,7 +15,9 @@ const RandomIncidentsCarousel = () => {
           allMongodbAiidprodIncidents {
             nodes {
               incident_id
-              reports
+              reports {
```
cesarvarela, line 18: Ditto.

cesarvarela (commented):
Mostly comments about reaping the benefits of this change a little more. I'm ok with merging as it is if it is too much work since the improvements are not visible anyway.

clari182:
@cesarvarela if you agree, let's leave it as it is and I'll start applying the changes in a new PR. (#2108)

PR #2103, opened 2023-06-19
Deploy to Production

cesarvarela:


cesarvarela:
@smcgregor ready for review

clari182:
> Things we talked about on the call (blocking):
> 
> * Setting the latest incident box to not auto-scroll
Fixed here #2109 
> * Having more than 2 incidents in the carousel
Fixed here #2102
> 
> New things (needs fixing):
> 
> * The carousel already took up a big portion of the landing page and the new card that wraps the carousel introduces more whitespace surrounding it. It is too much space on the landing page.
Fixed here #2109 
> * The carousel does not resize particularly well. This was a previously existing problem.
Will fix later to not hold this up

cesarvarela:
@smcgregor ready for another round

smcgregor:
Merging because we need to, but I am not a fan of the state of the landing page with this change set. I think you plan on addressing the sizing, but the inconsistency between the presentation of different content blocks on the landing page looks sloppy to me. The carousel changes exacerbated the issue when the block title moved out of the card.

Issue #2117, opened 2023-06-22
Inaccurate Counter for Incident Responses


smcgregor:
https://incidentdatabase.ai/entities/

Sort by "Incident Responses" and you will see the top rows as below. The problem is that it is a single incident that has multiple responses and each response from different parties is increasing the response count. [Roberto Mata](https://incidentdatabase.ai/entities/roberto-mata/) for instance is not an entity in more than a single incident, but he has apparently responded to several incidents.

<img width="1192" alt="Screenshot 2023-06-23 at 12 06 48 AM" src="https://github.com/responsible-ai-collaborative/aiid/assets/64780/71a381bf-3402-4cc8-9a98-4df3e865cb40">

smcgregor:
+1

PR #2119, opened 2023-06-23
Fix entity response counter

lmcnulty:
Resolves https://github.com/responsible-ai-collaborative/aiid/issues/2117.

### DIFF
--- a/site/gatsby-site/src/components/entities/EntitiesTable.js
+++ b/site/gatsby-site/src/components/entities/EntitiesTable.js
@@ -280,5 +280,5 @@ export default function EntitiesTable({ data, className = '', ...props }) {
   return <Table table={table} className={className} {...props} />;
 }
 
-var getResponseCount = (responses) =>
+const getResponseCount = (responses) =>
   new Set(responses.map((response) => response.incident_id)).size;
--- a/site/gatsby-site/src/components/entities/EntitiesTable.js
+++ b/site/gatsby-site/src/components/entities/EntitiesTable.js
@@ -92,7 +92,7 @@ function ResponseCell({ cell }) {
       )
     : cell.value;
 
-  const responseCount = new Set(filtered.map((response) => response.incident_id)).size;
+  const responseCount = getResponseCount(filtered);
 
   return (
     <div data-cy={`cell-${column.id}`}>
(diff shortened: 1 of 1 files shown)


site/gatsby-site/src/components/entities/EntitiesTable.js
```diff
@@ -276,3 +279,6 @@ export default function EntitiesTable({ data, className = '', ...props }) {
 
   return <Table table={table} className={className} {...props} />;
 }
+
+var getResponseCount = (responses) =>
```
cesarvarela, line 283: Nit: can you use `const` instead? It is probably the only function in the entire app using `var` :D

cesarvarela (changes requested):

site/gatsby-site/src/components/entities/EntitiesTable.js
lmcnulty: Done. Mostly for my own future reference, here's what I was thinking:

I like hoisting for functions, since it can allow you to put the most important information at the top. E.g.

```javascript
const importantThing = boringEscapeLogic(importantLogic(parameter));

function importantLogic(e) { ... }

function boringEscapeLogic(e) { ... }
```

I was using `var` because I thought it allowed me to get the same hoisting behavior as `function` but with arrow syntax, but I read more about it and realize that's not true. `var` allows use before declaration without a syntax error, but the value is undefined.

```javascript
// This does not work
const importantThing = boringEscapeLogic(importantLogic(parameter));

var importantLogic = (e) => { }

var boringEscapeLogic = (e) => { }
```

It worked because I didn't actually need hoisting here since the call is contained in another function.

```javascript
// This does work
const getImportantThing = () => boringEscapeLogic(importantLogic(parameter));

var importantLogic = (e) => { }

var boringEscapeLogic = (e) => { }

getImportantThing(parameter);
```

cesarvarela: Interesting! And that's why I don't like it, it leads to essays about JS quirks 😆 

And I think that's what ECMA tried to get away from when they introduced `let` and `const`, but that's for another essay too 🤣

cesarvarela (approved):

PR #2129, opened 2023-06-27
Add editors relationship to submissions

cesarvarela:


cesarvarela:
I think that should work because user ids are stored as an array of ids 🤔

pdcp1:
@cesarvarela Yes, you're right. It's an array of strings because it get the value from:
`const { _id: undefined, ...submission } = await submissions.findOne({ _id: input.submission_id });`

Good to merge!

PR #2130, opened 2023-06-28
Feature UI for removing duplicates

lmcnulty:
https://github.com/responsible-ai-collaborative/aiid/assets/25443411/d7bd4190-6578-49e9-89b2-110a620d5a30


smcgregor:
@lmcnulty how are you planning on solving this? Manipulating the duplicate table?

lmcnulty:
@smcgregor It adds an entry to the duplicates collection, and transfers reports and classifications to the selected incident.

site/gatsby-site/src/components/cite/Tools.js
```diff
@@ -162,4 +198,123 @@ function Tools({
...
+          try {
+            updateIncidentResponse = await updateIncident({
+              variables: {
+                query: { incident_id: duplicateIncidentId },
+                set: { reports: { link: reportIds } },
+              },
+            });
+          } catch (e) {
+            console.error(updateIncidentResponse);
+            alert(`Could not transfer reports to incident ${duplicateIncidentId}.`);
```
cesarvarela, line 280: Let's use toasts instead of alerts for consistency.

Also, using toasts and passing the exception will log it to Rollbar and give us visibility on when this fails.

site/gatsby-site/src/components/cite/Tools.js
```diff
@@ -162,4 +198,123 @@ function Tools({
...
+          window.location.pathname = '/';
+        }}
+      >
+        Remove Duplicate
+      </Button>
+    </>
+  );
+}
+
+const removeTypename = (obj) => {
```
cesarvarela, line 314: This triggered some alarms, but I guess it works 🤷 

And after **four** years, it looks like the Apollo team found a solution:
https://github.com/apollographql/apollo-feature-requests/issues/6#issuecomment-1507687565

site/gatsby-site/src/components/cite/Tools.js
```diff
@@ -162,4 +198,123 @@ function Tools({
...
+    FIND_INCIDENT,
+    { variables: { query: { incident_id: duplicateIncidentId } } }
+  );
+
+  const duplicateIncident = duplicateIncidentData?.incident;
+
+  return (
+    <>
+      <Trans>Transfer reports to incident</Trans>{' '}
+      <input
```
cesarvarela, line 222: What about using the existing `IncidentsField` component? It has auto-complete and will be more consistent with other forms, although I get that it is more geared toward multiple selections, we could add an option to limit selection to one item.

cesarvarela (commented):
And please add tests!

pdcp1:
@lmcnulty As part of this PR we should delete all User Subscriptions to the Incident (`type=incident`) which is marked as duplicated since that Incident won't exist anymore in the `incidents` collection.

site/gatsby-site/src/components/cite/RemoveDuplicateModal.js
```diff
@@ -0,0 +1,194 @@
...
+
+              setSubmitting(false);
+              onClose();
+              addToast({
+                severity: SEVERITY.success,
+                message: [
+                  `Incident ${incident.incident_id} marked`,
+                  `as duplicate of ${values.duplicateIncidentId[0]}.`,
+                  `Its page will updated within 24 hours.`,
+                ].join(' '),
```
pdcp1, line 157: I guess the reason for using `join` here is to avoid a long line. Personally, I prefer to have a simple string in these cases. Specially if we want to add translations with `<Trans>` or `t()`

site/gatsby-site/src/components/cite/RemoveDuplicateModal.js
```diff
@@ -0,0 +1,194 @@
...
+              });
+              setTimeout(() => (window.location.pathname = '/'), 5000);
```
pdcp1, line 159: Why we need a timeout here?

site/gatsby-site/src/components/cite/RemoveDuplicateModal.js
```diff
@@ -0,0 +1,194 @@
...
+              values.duplicateIncidentId.length == 0 ||
+              classificationsLoading ||
+              !classificationsData ||
+              !subscriptionsData ||
+              subscriptionsLoading ||
+              isSubmitting;
+
+            const submit = async () => {
+              setSubmitting(true);
+              const reportIds = removeTypename(
```
pdcp1, line 60: We don't need to call `removeTypename` function here since we are using just the `report_number`

site/gatsby-site/src/components/cite/RemoveDuplicateModal.js
```diff
@@ -0,0 +1,194 @@
...
+      </Modal.Body>
+    </Modal>
+  );
+}
+
+const removeTypename = (obj) => {
+  const replaced = JSON.stringify(obj).replace(/"__typename":"[A-Za-z]*",/g, '');
+
+  return JSON.parse(replaced);
+};
```
pdcp1, line 194: Please try to use this [removeTypenameFromVariables](https://www.apollographql.com/docs/react/api/link/apollo-link-remove-typename/)  Apollo function

pdcp1 (changes requested):

site/gatsby-site/src/components/cite/RemoveDuplicateModal.js
lmcnulty: So that there's time to read the status message in the toast.

site/gatsby-site/src/components/cite/RemoveDuplicateModal.js
lmcnulty: I naïvely thought that a function called `removeTypenameFromVariables` would take variables as input and remove the typenames from them. But actually it's a builder for an `ApolloLink` object that you're supposed to pass to the `ApolloClient` constructor. However, if I understand correctly, we have one client that we use everywhere through hooks. So this requires either instantiating a new client just for this, which would be annoying, or removing the typenames everywhere, which I think is ok but I would want to check with everyone. The change is made [here](https://github.com/responsible-ai-collaborative/aiid/pull/2130/files#diff-669c71f154e5d21df106aeb41c82230771bf55c23c69c1a84e175fff5c5fa8eb).

pdcp1: That's great. We don't use those `_typenames` fields anywhere, so if that approach using ApolloLink works, we can clean up the code where we delete them manually.

pdcp1:
@lmcnulty Correct me if I'm wrong, but it looks like you are done with all my comments/suggestions from my previous CR.
I see some comments from @cesarvarela that I'm not sure if you also take a look at.

cesarvarela (commented):
Please merge staging!

kepae:
@pdcp1 @cesarvarela @lmcnulty Let's please coordinate on whether this PR is ready to merge. Staging was merged, but I don't know what was communicated after that.

This feature is a good win for editors.

pdcp1 (approved):

cesarvarela:
:shipit:

lmcnulty:
@pdcp1, @cesarvarela Nope, I'm good to merge this.

kepae:
Thanks all! The taxes of coordination and cooperation.

PR #2135, opened 2023-07-01
Update twitter tests

cesarvarela:
@smcgregor, this should fix the current production build


smcgregor:
I thought it was a flaky test and retried it. I didn't think Twitter randomly changing user account policy could be the problem.

PR #2136, opened 2023-07-03
Fix share twitter test

pdcp1:
Twitter change the URL when the user shares an incident


cesarvarela:
@pdcp1, please fix conflicts!

clari182 (approved):

PR #2142, opened 2023-07-07
Add tSNE csv export script

lmcnulty:
A collaborator wanted a CSV with the data from the tSNE visualization. I wrote this script to generate it. Committing this won't have any effect on the site, but I figure we might as well add it in case this comes up again.


lmcnulty:
Closing as stale and not necessary to the codebase.

PR #2159, opened 2023-07-14
Deploy to Production

cesarvarela:


cesarvarela:
Added a fix to the current production build

cesarvarela:
@clari182 I think we might not need the `customTypesApiToken` since we are copy-pasting custom types definitions instead of hitting the API directly, but please confirm.

cesarvarela:
@smcgregor this should fix the current production build

smcgregor:
@Janetbananet you were wanting a final update to the blog content. Do you have an ETA on that? Can we deploy, content wise?

Janetbananet:
I updated the verbiage in Prismic Staging yesterday (saved, not published), but I believe @clari182 was going to help move it across the finish line today since she knows more about your current deploy processes and how that integrates with Prismic.

clari182:
@cesarvarela the `customTypesApiToken` is required for the `gatsby-source-prismic` plugin to fetch the custom types. Otherwise we'd need to define them inside the code as well (as a schema).
@Janetbananet I triggered the deploy to staging with your prismic updates. It looked good to me!

Issue #2047, opened 2023-05-26
Issue, Hazard, Risk, Vulnerability, Audits, and Controversy Support


smcgregor:
*This is a working document with some elements that are ready for development*

While there is convergence on what constitutes an "AI incident", there are still considerable differences between definitions of how the concept of "incident in waiting" is defined. We call these "issues," the OECD is calling them "hazards," Robust Intelligence calls them "risks," ARVA calls them "vulnerabilities," AIAAIC calls them "controversies," and various algorithmic assessment organizations call them audits (or at least, audits will always produce one or more incident in waiting). All these things vary subtly in definition, application, and use between organizations.

The role of the Responsible AI Collaborative going back to the [original research publication](https://arxiv.org/abs/2011.08512) has always been to act as the union of multiple perspectives and provide tools to support sharing across those perspectives. This is a challenging proposition. Pretty much any multi-stakeholder ontological project I am aware of has inevitably degenerated into never ending discussions over the most difficult elements to define. For something that has no underlying, singular "right" answer, it is best to find ways of moving forward that don't require universal agreement. **The purpose of this GitHub issue is to detail how to proceed technologically without needing to resolve the definitional question of "incident in waiting".** 

The AIID's current entrant into this space is the "[AI Issue](https://incidentdatabase.ai/apps/discover/?display=details&is_incident_report=false&page=1&sortBy=instant_search-en-featured)." We chose this [term](https://arxiv.org/abs/2211.10384) intentionally to cover multiple aspects of "incident in waiting." It is meant to be specific enough to capture elements of risk, while general enough to cover the field. The "issue" term also means we can index concepts covered by other communities and link out to those communities if/when they operate their own processes. While we would prefer such organizations join the Responsible AI Collaborative and integrate from the beginning, that will not be possible universally (e.g., when a database is operated by a sovereign state). Therefore, we need to maintain flexibility.

This also plugs into the drive for federating the AI Incident Database -- something that we will soon have a test case for with an index of deepfakes. Incident databases for things like deepfakes require different editing processes and metadata. How federation works with incidents is fairly clear. Incidents have a natural scope that will support federating responsibilities among multiple nodes. However, this does not work for incidents in waiting. Often there is no concrete definition of what specific system can produce the incident. Worse, all systems will produce a great many incidents when placed into the wrong context. Behind every system is an infinity of risks. This is why the [ForHumanity](https://forhumanity.center/) audit criteria centers on these four elements,

[...]

PR #2170, opened 2023-07-19
Update schema to handle reports linked to classifications

cesarvarela:
This tweaks the classifications schema to allow linking a classification to an incident or a report, or something else.

Before, `classifications` had `incident_id`.

Now `classifications` have two fields:

- `issue_id`, the incident  ID, report number, or something else.
- `issue_type` specifies the target type (`incident`, `report`, etc.)

part of #2047 

Tests passed here: https://github.com/cesarvarela/aiid/pull/54

Successful deploy: https://deploy-preview-54--cesarvarela-staging.netlify.app/

### DIFF
--- a/site/gatsby-site/src/templates/citeDynamicTemplate.js
+++ b/site/gatsby-site/src/templates/citeDynamicTemplate.js
@@ -54,7 +54,7 @@ function CiteDynamicTemplate({
   });
 
   const { data: classificationsData } = useQuery(FIND_CLASSIFICATION, {
-    variables: { query: { incident_id, publish: true } },
+    variables: { query: { issue_id: incident_id, issue_type: 'incident', publish: true } },
   });
 
   useEffect(() => {
--- a/site/gatsby-site/src/templates/citeDynamicTemplate.js
+++ b/site/gatsby-site/src/templates/citeDynamicTemplate.js
@@ -89,7 +89,7 @@ function CiteDynamicTemplate({
         locale,
       });
 
-      const sortedIncidentReports = sortIncidentsByDatePublished(incidentReports);
+      const sortedIncidentReports = sortIncidentsByDatePublished([...incidentReports]);
 
       const sortedReports = sortedIncidentReports.filter((report) => isCompleteReport(report));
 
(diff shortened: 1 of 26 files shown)


cesarvarela:
ping @smcgregor and @lmcnulty  for comments

lmcnulty:
Did you consider keeping `incident_id` but adding `report_number` as well? That would enforce querying by the correct type. As it is, it's easy to imagine someone writing this by mistake:

```javascript
const incident100 = classifications.findOne({ issue_id: 100 });
//                                                         ^
//                             Missing `issue_type: "incident"`
//                       Might return classification for report 100!
```

Whereas if someone has to pick between querying by `incident_id` or `report_number`, then it would be obvious if they used the wrong one:

```javascript
const incident100 = classifications.findOne({ report_number: 100 });
```

site/gatsby-site/src/templates/citeDynamicTemplate.js
```diff
@@ -129,6 +129,9 @@ function CiteDynamicTemplate({
           <Spinner />
         </div>
       )}
+      <div>
+        {String(loading)} : {String(loadingIncident)}
+      </div>
```
pdcp1, line 134: I don't understand this code. It looks like it's a debugging code since `loading` and `loadingIncident` are boolean variables

pdcp1 (changes requested):

site/gatsby-site/src/templates/citeDynamicTemplate.js
cesarvarela: Woor that should have never been pushed

pdcp1 (approved):

smcgregor:
Following up on Luna's comment... Would it make sense to make both the report and incident numbers lists so that taxonomies could apply to many different records of those types? Please feel free to reject this if it breaks all the things, but it occurs to me that such functionality would confer the ability to apply classifications to multiple records across those two types.

cesarvarela:
My idea was to allow **anything** to be linked to classification, but if that won't ever happen, I'll switch to adding `report_number` instead. And yes, using lists will break "all the things," but this is the moment to do that.

smcgregor:
I see your point, but I am also trying to think through whether we configure classifications to be associated with multiple records. I see some benefits, but also some costs. Maybe a discussion on this design?

cesarvarela:
We can talk about it on tomorrow's call if that works for you.

PR #2072, opened 2023-06-07
Create new `history` DB for version history

pdcp1:
This close https://github.com/responsible-ai-collaborative/aiid/issues/2041

## Features added

- [ ] new field `epoch_date_modified` on `submissions`
- [ ] All `epoch_date_modified` fields on `reports` and `submissions` now includes hh:mm:ss
- [ ] new `logReportHistory` function that creates a new entry into the `reports` collection in the `history` DB
- [ ] new `logIncidentHistory` function that creates a new entry into the `incidents` collection in the `history` DB

## Deployment environment
https://app.netlify.com/sites/pablocosta-staging/deploys

## Testing environment
https://deploy-preview-21--pablocosta-staging.netlify.app/
Admin user: pablo@botsfactory.io
pass: testUser123

## Places where we log a history change

Reports and Incidents |  
-- | --
**URL** | **Action**
/cite/10#26 | Show more on a report and Flag report
/cite/edit/?report_number=10 | Edit report
/apps/discover | Flag report
  |  
/cite/9 | Flag incident as nor related
/apps/incidents | Edit incident
/incidents/edit?incident_id=10 | Edit incident
/cite/10 | Clone incident
/incidents/new | New incident
N/A | Promote submission to incident/report

PR #2173, opened 2023-07-21
Move `transformIncidentData` and `transformReportData` into a custom hook

pdcp1:
On top of #2072


pdcp1:
Closing this in favor of #2185

Issue #2048, opened 2023-05-29
Move Page Content Out of Source Code


smcgregor:
At present many of the pages are source controlled rather than authored in the DB and statically rendered. This is not maintainable long term and it makes it more difficult for non-incidentdatabase.ai websites to stand up versions of the codebase.

Options to consider include,

1. Moving everything into the mongoDB database and build those pages at server side render time
2. Move some of the content into a config file and stand up the rest of the content in a new subdomain that uses a CMS
3. Something else?

smcgregor:
@cesarvarela thoughts on this?

cesarvarela:
The "something else" for me is to integrate a headless CMS. There are a few out there, some cheap, some very expensive, I have previous experience integrating with Prismic, and their free tier is probably enough.

https://prismic.io/ 

https://prismic.io/docs/gatsby

The workflow would be as follows:

- user adds a new post on Prismic, using the rich text editor (no more markdown)
- it may or may not set the post as published
- we can choose to fire the Netlify webhook from Prismic once a post is published or let the cron build pick it up

smcgregor:
Free plan is enough or the $1,200 plan? Absence of markdown is annoying.

cesarvarela:
Unless I'm missing something the free or at most ~20usd plans are enough.

There are other options that do support markdown like: https://tina.io/ and https://www.sanity.io/, I don't have experience with them but it is the same concept.

smcgregor:
I didn't hover the pricing cards. Without hovering them, you don't see the lower priced plans. Thanks for the links!

cesarvarela:
This seems related: https://www.contentlayer.dev/blog/working-with-content-is-hard-for-developers

PR #2177, opened 2023-07-24
Feature/prismic sidebar

clari182:
- Part of #2048 

Uses sidebar content from Prismic: https://deploy-preview-2177--staging-aiid.netlify.app/
Uses config sidebar content (falls back to config if no prismic content is available): https://deploy-preview-21--clarayoudale-staging.netlify.app/


site/gatsby-site/src/components/sidebar/index.js
```diff
@@ -9,8 +9,65 @@ import { faChevronLeft } from '@fortawesome/free-solid-svg-icons';
...
+        }
+      }
+    }
+  `);
+
+  let sidebarItems = [];
+
+  if (sidebar.edges.length > 0) {
+    sidebarItems = sidebar.edges.map((item) => {
+      if (item.node?.data) {
```
cesarvarela, line 49: Is there an instance where this is not present? if this happens, `map` will add an `undefined` item.

clari182: No, it shouldn't happen because this is part of the prismic structure so if there's an item, it has node.data
I removed the if

cesarvarela (commented):
Can you add the schema definitions, too? So it doesn't break if it is missing like it did before with the footer.

cesarvarela (changes requested):
Something is going on with the sidebar:

image

cesarvarela (approved):

PR #2181, opened 2023-07-25
Deploy to Production

cesarvarela:


cesarvarela:
@smcgregor is ready for review. We noticed that it is impossible to roll back to the first version because we store the result of the edition rather than the pre-edited version. Should we do something like a migration that saves the first version of everything?

smcgregor:
The migration seems a reasonable approach, although it will immediately multiply the size of the DB. Where should I look, UI wise, when reviewing this functionality? What is the expected behavior at present? We need user-centered documentation for things like this.

smcgregor:
* I am getting a bunch of Cloudinary errors in the console, but the images are still displayed.
* `epoch_date_modified: epoch_date_modified` the schema documentation needs updating?

I am good with merging these to clear staging for the CSET changes if you are also 👍

cesarvarela:
@smcgregor Yes please merge to clear staging

PR #2185, opened 2023-07-25
Add `LogHistoryContext` and refactor log history functions

### LINKED ISSUE #2138: Move `transformIncidentData` into a custom hook
Could this logic (and the `transformIncidentData` call)  be done in one function? Maybe create a custom hook `useLogHistory` that returns a function and then we have to only add one line everywhere we want to log a change:

```
const logStuff = useLogHistory();

... 

await logStuff({ ...everything needed here...)
```

_Originally posted by @cesarvarela in https://github.com/responsible-ai-collaborative/aiid/pull/2072#discussion_r1248178065_

pdcp1:
This closes #2138


site/gatsby-site/src/contexts/LogHistoryContext.js
```diff
@@ -0,0 +1,64 @@
...
+  logReportHistory(report, updates, user) {
+    report;
+    updates;
+    user;
+  },
+});
+
+export default LogHistoryContext;
+
+export function LogHistoryContextProvider({ children }) {
```
cesarvarela, line 23: I think you don't need a context/provider, and just a `useLogHistory` function will do 🤔

pdcp1: Done

PR #2192, opened 2023-07-27
Feature/prismic footer

clari182:
- part of #2048 

This preview uses Prismic footer: https://deploy-preview-23--clarayoudale-staging.netlify.app/
This preview uses current footer (fallbacks to current config footer if no prismic content is available): https://deploy-preview-2192--staging-aiid.netlify.app/


pdcp1:
@clari182 Is it OK if both footers look exactly the same?
- https://deploy-preview-23--clarayoudale-staging.netlify.app
- https://deploy-preview-2192--staging-aiid.netlify.app

The difference is that the first one is pulling the footer info from Prismic and the second one from the code as we currently do, correct?

clari182:
> @clari182 Is it OK if both footers look exactly the same?
> 
> * https://deploy-preview-23--clarayoudale-staging.netlify.app
> * https://deploy-preview-2192--staging-aiid.netlify.app
> 
> The difference is that the first one is pulling the footer info from Prismic and the second one from the code as we currently do, correct?

Correct!

pdcp1 (approved):

PR #2193, opened 2023-07-27
Deploy to Production

cesarvarela:


cesarvarela:
@smcgregor ready for review. You'll have to fix conflicts, accepting incoming should be enough.

smcgregor:
Not blocking but needs addressing:

* The filters on the discover app still have the issue I messaged about on Slack
* When entering the incident number in the search endpoint, the search for title hangs.
* Clicking on the pie chart in the GMF taxonomy does not apply the filter that you clicked on

Still need to test

* Subscription and user cloning elements

smcgregor:
@cesarvarela can you resolve the conflicts then I'll merge? I am on my phone for the next 16 hours (or flying).

smcgregor:
There are differences between master and staging that are difficult to understand on my phone.

smcgregor:
@cesarvarela updated per above.

smcgregor:
I created a new incident on staging and added it. I'll look at the email when it comes through and merge after if it is good to go.

PR #2205, opened 2023-07-31
Hide hidden filters

### LINKED ISSUE #2204: Strange Discover Filter Behavior on Mobile
Narrow the screen to mobile widths and the `1` value appears on the top level of the filterable list.

Screenshot_20230728-184215

clari182:
- closes #2204


cesarvarela (commented):
Looks like this addresses the issue but there is some weirdness still (iPhone SE preset):


image

PR #2207, opened 2023-07-31
Fix/published taxonomies chart

clari182:
Filters unpublished taxonomies from chart page


cesarvarela (commented):
@clari182 where can I look at what difference this makes?

lmcnulty (approved):

PR #2215, opened 2023-08-02
Update schema to handle classifications linking to multiple incidents and/or reports

cesarvarela:
part of #2047 
Working deploy preview: https://deploy-preview-55--cesarvarela-staging.netlify.app/
PR with passing tests: https://github.com/cesarvarela/aiid/pull/55


site/gatsby-site/page-creators/createTsneVisualizationPage.js
```diff
@@ -41,9 +41,11 @@ const createTsneVisualizationPage = async (graphql, createPage) => {
 
   const classificationsQuery = await graphql(`
     query Classifications {
-      allMongodbAiidprodClassifications(filter: { incident_id: { in: [${incidentIds}] } }, limit: 9999999) {
+      allMongodbAiidprodClassifications(filter: { incidents: { elemMatch: { incident_id: { in: [${incidentIds}] } } } }) {
```
pdcp1, line 44: Why the limit is not needed anymore?

site/gatsby-site/src/pages/apps/classifications.js
```diff
@@ -260,7 +260,8 @@ export default function ClassificationsDbView(props) {
       const classificationsToRowsMap = (cObj) => {
         const row = {};
 
-        row['IncidentId'] = cObj.incident_id;
+        row['IncidentId'] = cObj.incidents[0].incident_id;
```
pdcp1, line 263: Is this what we wanted for the Incident Id column? Should we display just the first incident from the Classification?

pdcp1 (changes requested):

site/gatsby-site/src/pages/apps/classifications.js
cesarvarela: Yeah, we'll have to live with this code smell; at the moment, a CSET classification only applies to one and only one  incident.

site/gatsby-site/page-creators/createTsneVisualizationPage.js
cesarvarela: Whoops! Added back!

site/gatsby-site/src/contexts/userContext/UserContextProvider.js
```diff
@@ -41,9 +41,6 @@ const getApolloCLient = (getValidAccessToken) =>
         User: {
           keyFields: ['userId'],
         },
-        Classification: {
```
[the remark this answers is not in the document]
cesarvarela: I couldn't find a way to express the new way we'll be handling classifications:  https://www.apollographql.com/docs/react/caching/cache-configuration/.

(With this change, it reverts to the default behavior of using the `_id` field, so we should be good )

site/gatsby-site/src/pages/apps/classifications.js
cesarvarela: @kepae 
> should we throw a static build error since (a) all classifications are (currently) expected to have at least one incident, and (b) there's currently no plan for handling classifications without incidents? (per chatting yesterday)
Could also be an ERR log of some kind, for data-only problems. Build error might be too aggressive, but earlier better than later to do that.

I like that. We have a test file for this kind of check: `integrity.cy.js`; I'll add a new test.

pdcp1 (approved):

pdcp1:
@cesarvarela I just finished my review. It looks good 👍

kepae:
Ditto, thanks for adding and pointing me to the integrity tests, going to look more into those.

kepae (approved):

PR #2216, opened 2023-08-03
Add Incident Version History page

### LINKED ISSUE #2213: Create new Incident Version History page
This should follow the design approved from #2191

Note: this issue only includes the read-only version of this page, without the `View full version` and `Restore version` buttons.

image

pdcp1:
This closes #2213

### DIFF
--- a/site/gatsby-site/cypress/e2e/integration/incidents/history.cy.js
+++ b/site/gatsby-site/cypress/e2e/integration/incidents/history.cy.js
@@ -123,4 +123,16 @@ describe('Incidents', () => {
 
     cy.contains('There are no version history records for this Incident').should('exist');
   });
+
+  it('Should display an error message if no Incident ID is provided', () => {
+    cy.visit('/incidents/history?incident_id=');
+
+    cy.contains('Invalid Incident ID').should('exist');
+  });
+
+  it('Should display an error message if an invalid Incident ID is provided', () => {
+    cy.visit('/incidents/history?incident_id=xxx');
+
+    cy.contains('Invalid Incident ID').should('exist');
+  });
 });
--- a/site/gatsby-site/cypress/e2e/integration/incidents/history.cy.js
+++ b/site/gatsby-site/cypress/e2e/integration/incidents/history.cy.js
@@ -135,4 +135,14 @@ describe('Incidents', () => {
 
     cy.contains('Invalid Incident ID').should('exist');
   });
+
+  it('Should go back to the Incident', () => {
+    cy.visit(url);
+
+    cy.waitForStableDOM();
+
+    cy.contains(`Back to Incident 10`).click();
+
+    cy.url().should('include', '/cite/10');
+  });
 });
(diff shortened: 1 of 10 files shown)


cesarvarela (commented):
Can you tweak the conditional rendering so the empty message doesn't flash before showing the incident history? I you reload the page you'll see the "no version history" message before the actual data gets loaded.

Why this doesn't show any specific changes?

image

pdcp1:
@cesarvarela 

> Why this doesn't show any specific changes?

That's weird. But it looks like we have inconsistent data on staging on the `history` database, probably because of dev testing in the past.
I confirmed that those entries are all the same, so the row changes are empty.
I added a filtering logic to avoid this, just in case.
It's ready for review again

cesarvarela (commented):
It looks like there is something wrong with the user that gets set as the one making changes. After making a few tests it is showing Sean's user:

image

But I'm logged in as me:

image


Not sure how far we want to go with this, but it would be better if the diff for texts showed the specific text diff, right now it is showing in it as this:

image

But there are some packages around that do a better job, like this one:

image

Or maybe this only for string fields"

image



 PS:

Not sure if it is worth fixing, but there s something weird going on with id parameter:

This link with a random id has the following results:
https://deploy-preview-2216--staging-aiid.netlify.app/incidents/history/?incident_id=5rf

Empty id shows results too:
https://deploy-preview-2216--staging-aiid.netlify.app/incidents/history/?incident_id=

cesarvarela (commented):
Will we run a migration so every incident has its initial version?

image

pdcp1:
> Will we run a migration so every incident has its initial version?

@cesarvarela This is something that we didn't define. We can run a migration to generate an initial version to all incidents in the database, or start logging the versions from now on. The second approach doesn't take into consideration the first edit.

pdcp1:
@cesarvarela All fixed. Please review again

pdcp1:
@cesarvarela I made some visual improvements and also I added a "Back to Incident X" button on the top in the same way as the edit page.
The last deploy passed, so the preview link is updated.

pdcp1:
A new migration to initialize all first Incident and Report versions was added to this PR.

cesarvarela (commented):
The diff only compares to the previous point? In this example, I didn't change the editor notes field, but it still shows as if I was the one that added the text:

image

Also, there seems to be a caching issue:

- make a change and save
- Hit the got back to incident button
- Hit view history
- your recently made change won't be there. You have to refresh to see them

On another note, if you hit save without making any changes, it still creates a history point. Should we keep that?


(I've been testing on incident 9: https://deploy-preview-2216--staging-aiid.netlify.app/incidents/history/?incident_id=9)

pdcp1:
@cesarvarela 
> The diff only compares to the previous point?

Yes

>In this example, I didn't change the editor notes field, but it still shows as if I was the one that added the text:

I couldn't reproduce this issue. Can you test again?

>Also, there seems to be a caching issue

Fixed

> if you hit save without making any changes, it still creates a history point. Should we keep that?

Yes, every time that the user hit the Save button it will create a history version point. If we want to avoid this, we need to compare the updated version to the old one and if both are the same, skip the new version. I don't know if it is worth implementing it, since I think users don't save without making any changes too often.

Issue #2219, opened 2023-08-04
List of all incidents not rendering reports: repeated "Missing Report [object Object]"


kepae:
https://incidentdatabase.ai/summaries/incidents/

<img width="953" alt="image" src="https://github.com/responsible-ai-collaborative/aiid/assets/4238598/082cc4d3-c976-437b-91c0-458200290c63">

kepae:
Assigning to myself to triage, will call in backup if needed :-)
This page may have been live for a bit like this.

kepae:
* I've reproduced the issue locally, and found a quick fix. It has something to do with the report's `report_number` representation being nested in the schema/query that I haven't found yet:
\
broke:
`data.allMongodbAiidprodReports.nodes.find((r) => r.report_number == report_number)`
woke:
`data.allMongodbAiidprodReports.nodes.find((r) => r.report_number == report_number.report_number)`
\
Looking to understand this because the query shouldn't be problematic like this.

* *In addition, less important:* Incident 563 is a buggy dup of 564? It has an empty report list.

```
{
  first_report: undefined,
  incident: {
    incident_id: 563,
    title: 'Premières plaintes auprès de la Cnil contre ChatGPT',
    date: '2023-04-05',
    reports: []
  }
}
{
  first_report: {
    id: '9776902b-f3e0-56f0-b1fb-8c971d6210d4',
    report_number: 3205,
    title: 'Premières plaintes auprès de la Cnil contre ChatGPT',
    url: 'https://www.frenchweb.fr/premieres-plaintes-aupres-de-la-cnil-contre-chatgpt/441833',
    inputs_outputs: null
  },
  incident: {
    incident_id: 564,
    title: 'Premières plaintes auprès de la Cnil contre ChatGPT',
    date: '2023-04-05',
    reports: [ [Object] ]
  }
}
```

See working screenshot:

<img width="1103" alt="image" src="https://github.com/responsible-ai-collaborative/aiid/assets/4238598/a5c40dbe-5795-4ae2-b624-3d4223bf6da1">

PR #2220, opened 2023-08-04
Fix Incident list missing reports

kepae:
Triage fix https://github.com/responsible-ai-collaborative/aiid/issues/2219.
hat tip to @cesarvarela for knowing the schema change history.

I want to push to staging before the weekend, but next steps are also:
* would be reviewing other places where `report_number` appears in queries or reports are joined with incidents redundantly, as was here.
* adding preventative test.

fix working on my instance:
image


kepae:
@cesarvarela Would you mind "approving" this PR so that the netlify staging deploy can proceed?

kepae:
Ah, copy. I see now, staging@ rather than the feature branch name.

PR #2233, opened 2023-08-14
Add `Restore Version` button on Incident History page

### LINKED ISSUE #2231: Restore old Incident version feature
Add the `Restore Version` link on the Incident History page that pops up a confirmation message and restores the version selected.
Only editors should be able to restore a past version

image

pdcp1:
This closes #2231


pdcp1:
> I think this needs a loading state somewhere

@cesarvarela added

cesarvarela (changes requested):
It seems there is a bug when adding new entities on an edit:

image


Refreshing the page seems to work, so it is probably an Apollo cache thing:

image

cesarvarela (approved):

PR #2235, opened 2023-08-15
Add News Digest to sidebar

lmcnulty:
image

Also:

- Makes cards wider
- Puts them in a gray inset container – I think white cards against a white background look bad in general and we should move away from using them.
- Moves actions into a dropdown


kepae:
@lmcnulty can you detail how this data will stay fresh/updated? You mentioned a github workflow before.

lmcnulty:
@kepae It runs [this workflow](https://github.com/responsible-ai-collaborative/nlp-monitoring/actions/workflows/fetch-news.yml) on a daily basis.

Issue #2237, opened 2023-08-16
Error on Report edition


pdcp1:
On staging (and probably on Production) there is an error when we try to edit a Report.
It can be reproduced with this Report:

https://staging-aiid.netlify.app/cite/edit/?report_number=3193&incident_id=559

## Steps to reproduce

1. Open https://staging-aiid.netlify.app/cite/edit/?report_number=3193&incident_id=559
2. Edit any field
3. click `Save`
4. An error occurs

<img width="1142" alt="image" src="https://github.com/responsible-ai-collaborative/aiid/assets/6564809/0347cfd6-3da4-40d2-a0c8-095b5d7c431a">

pdcp1:
I found the issue here. It's a data issue on the DB. Some old submissions doesn't have the `user` field set, so when this submission is promoted to a report, this report results in a `report.user == undefined`

pdcp1:
Closing this since it was already merged into Staging.

PR #2239, opened 2023-08-17
Fix reports with `user == undefined`

pdcp1:
This PR fixes https://github.com/responsible-ai-collaborative/aiid/issues/2237

- Fix the current reports with `user == undefied`
- Set `null` `user` value for submissions that are submitted anonymously

### DIFF
--- a/site/gatsby-site/migrations/2023.08.17T15.56.33.fix-report-users.js
+++ b/site/gatsby-site/migrations/2023.08.17T15.56.33.fix-report-users.js
@@ -1,0 +1,21 @@
+const config = require('../config');
+
+/**
+ *
+ * @param {{context: {client: import('mongodb').MongoClient}}} context
+ */
+
+exports.up = async ({ context: { client } }) => {
+  await client.connect();
+
+  const reportsCollection = client.db(config.realm.production_db.db_name).collection('reports');
+
+  const updateResult = await reportsCollection.updateMany(
+    {
+      user: undefined,
+    },
+    { $unset: { user: '' } }
+  );
+
+  console.log('Update result:', updateResult);
+};
--- a/site/gatsby-site/migrations/2023.08.17T15.56.33.fix-report-users.js
+++ b/site/gatsby-site/migrations/2023.08.17T15.56.33.fix-report-users.js
@@ -19,3 +19,6 @@ exports.up = async ({ context: { client } }) => {
 
   console.log('Update result:', updateResult);
 };
+
+/** @type {import('umzug').MigrationFn<any>} */
+exports.down = async () => {};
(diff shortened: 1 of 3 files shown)


cesarvarela (commented):
Looks good, but I'm unsure about setting Sean up as de default submitter.

@smcgregor @kepae ?

site/gatsby-site/migrations/2023.08.17T15.56.33.fix-report-users.js
```diff
@@ -0,0 +1,21 @@
...
+
+  const updateResult = await reportsCollection.updateMany(
+    {
+      user: undefined,
+    },
+    { $unset: { user: '' } }
+  );
+
+  console.log('Update result:', updateResult);
+};
```
kepae, line 21: Please consider adding the empty `down` function even though there's no logical way to revert -- this way the node migrator tool won't throw an error on down migrations.

kepae:
Agree. It may be better to create a default editor user (with no real permissions) that is only used for backfilling submission metadata.

Would it be possible to create a default editor user?

site/gatsby-site/migrations/2023.08.17T15.56.33.fix-report-users.js
pdcp1: added

pdcp1:
We are currently usign Sean as a default Editor when promoting a submission (see `promoteSubmissionToReport.js` code below), so I keep consistency for the `user` as well.

```
const editors = (!submission.incident_editors || !submission.incident_editors.length)
        ? [SEAN_USER_ID]
        : submission.incident_editors;
```

kepae:
Ah sounds good, thanks for explaining -- it's the "existing" logic. We shouldn't block fixing the bug on it, so LGTM.

smcgregor:
+1 with having a system default user.

kepae:
Opened https://github.com/responsible-ai-collaborative/aiid/issues/2240 to track that. Since this bugfix doesn't change existing behavior, I don't believe we should block it in this PR -- unless it would explode the number of submissions attached to your user, but I don't think it should if this only affects old submissions that pre-dated user logic.

kepae:
Ah, wait. editor != user.

Will submissions from anonymous users now be attributed to Sean?

pdcp1:
After further analysis, we decided to set the `report.user = null` (instead of Sean user) for submissions that are submitted by an anonymous user.
Currently, we have more than 2500 reports in this situation and the report edition works fine.

In the future, when we have a Default user, we can assign this Default user when an anonymous user submit a submission.

cesarvarela (commented):
LGTM I'll wait for the deploy to finish before merging

kepae (approved):

Issue #2242, opened 2023-08-18
View Incident history version details


pdcp1:
1.  Add `View full version` button on Incident History page

<img width="944" alt="image" src="https://github.com/responsible-ai-collaborative/aiid/assets/6564809/87071c12-e997-4bde-89d9-8d64e8dcca73">

2. Add View History version modal

<img width="769" alt="image" src="https://github.com/responsible-ai-collaborative/aiid/assets/6564809/546b48a5-9449-4518-b50f-677f841d5279">

pdcp1:
Closing this since it was merged into Staging

PR #2248, opened 2023-08-22
Add view full incident version modal

pdcp1:
This closes https://github.com/responsible-ai-collaborative/aiid/issues/2242


cesarvarela (commented):
Can you try to get the deploy passing?

cesarvarela (commented):
Sometimes it is not using the modal full width:

image


Not related, but this should show a user name instead.

image

cesarvarela (commented):
Seems like editor names are missing:

image


Maybe it should also render the timeline?

cesarvarela:
I think this should render the Cite template in its fullest to be really full :P

cesarvarela (commented):
Should we hide the right sidebar? It looks weird otherwise:

image

It is missing user first/last names when using anonymous users, maybe show a placeholder, like `Redacted` or something like that 🤔

cesarvarela (commented):
We should hide the editor since you are not able to edit old version classifications:
image

cesarvarela (changes requested):
What do these empty squares mean?

https://deploy-preview-2248--staging-aiid.netlify.app/incidents/history/?incident_id=559

image

pdcp1:
> What do these empty squares mean?

@cesarvarela It's because the entities `test` on that history record doesn't exist on the `entities` collection.

image

Probably because of a previous DB restore from Production using the Github action.
I just realized that the `Restore Prod DB into Staging` Github action doesn't restore the `history` database.

cesarvarela:
> I just realized that the Restore Prod DB into Staging Github action doesn't restore the history database.

I wonder if it should though 🤔

cesarvarela (approved):

PR #2253, opened 2023-08-23
Fix layout overflow issue on firefox

### LINKED ISSUE #2252: Content overflows horizontally at medium widths
This may be related to recent changes to the layout.

image

cesarvarela:
It's been a while since having to deal with browser rendering inconsistencies but here we are.

closes #2252


cesarvarela:
@lmcnulty unless you have a better solution, I'll merge this, and we can fix It without hacks later

Issue #2254, opened 2023-08-23
Cleanup Layout CSS soup and leftover CSS files


cesarvarela:
There is a ton of stuff going on with the layout component CSS that could use some cleanup:

<img width="702" alt="image" src="https://github.com/responsible-ai-collaborative/aiid/assets/1172479/4f552940-4d13-4726-92b7-895959cb97f7">

- instead of using the `tw-layout` class, we should be using tailwind classes only as it is the recommended practice
- `#content` doesn't seem to need  the `flex-*` classes
- many nested `w-full` divs aren't needed

We should also remove the `global.css` file (and use tailwind instead) and probably `custom.css`.

Tech-debty stuff.

PR #2255, opened 2023-08-23
Deploy to Production

cesarvarela:
Quick fix for #2252 

The reason to use `overflow:clip` is to be able to use `position:sticky` and not have to resort to Intersection Observers and the like. It is not working correctly on Firefox, as it does on Chrome and Safari.

related: #2254 

@kepae


kepae:
I'll confirm and merge this once it hits staging in https://github.com/responsible-ai-collaborative/aiid/pull/2253. 👍

kepae (approved):

PR #2258, opened 2023-08-23
Deploy to Production

cesarvarela:

PR #2261, opened 2023-08-24
Don't restore the `user` report field

pdcp1:
This fix an issue on this PR https://github.com/responsible-ai-collaborative/aiid/pull/2251 

- When you try to restore a Report's version with a `user` field set (edge case), it throws an error.
ie: https://deploy-preview-2251--staging-aiid.netlify.app/cite/history?report_number=3035&incident_id=543

image

- Another error occurs when the Report version entry has an `embedding` value. In this case, we should remove the `__typename` field from the `embedding` object

### DIFF
--- a/site/gatsby-site/cypress/e2e/integration/reportHistory.cy.js
+++ b/site/gatsby-site/cypress/e2e/integration/reportHistory.cy.js
@@ -252,6 +252,7 @@ describe('Report History', () => {
 
     delete updatedReport._id;
     delete updatedReport.modifiedBy;
+    delete updatedReport.user;
 
     cy.wait('@UpdateReport').then((xhr) => {
       expect(xhr.request.body.operationName).to.eq('UpdateReport');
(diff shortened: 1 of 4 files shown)


kepae:
Could you explain a little bit more? Sounds like this should go into production (https://github.com/responsible-ai-collaborative/aiid/pull/2258) at the same time as #2251

pdcp1:
> Could you explain a little bit more? Sounds like this should go into production (#2258) at the same time as #2251

@kepae Yes, it should be merged into Production along with https://github.com/responsible-ai-collaborative/aiid/pull/2251

With this PR we are fixing the following issue:

- When you try to restore a Report's version with a `user` field set (edge case), it throws an error.
ie: https://deploy-preview-2251--staging-aiid.netlify.app/cite/history?report_number=3035&incident_id=543
image
 

- The same Report's version restore feature on this PR preview link is working:

https://deploy-preview-2261--staging-aiid.netlify.app/cite/history?report_number=3035&incident_id=543

Issue #2272, opened 2023-09-06
`admin.cy.js` test is failing due to a slow performance query


pdcp1:
Staging deploys are failing due to a bad performance on the `FindUsers` graphql query. It takes ~1 minute to complete and the test fails before that.

```
query FindUsers {
    users {
      roles
      userId
      first_name
      last_name
      adminData {
        email
        disabled
        creationDate
        lastAuthenticationDate
      }
    }
  }
```

A temporary solution to unblock the deployment process is to skip this test. In the meantime, we must find a better solution to that [Users table](https://staging-aiid.netlify.app/admin/). A pagination approach could help.

kepae:
Let's skip the test temporarily to unblock out deploys and keep this issue open to track it.

cesarvarela:
The reason for this is that the atlas admin API is extremely slow:

https://github.com/responsible-ai-collaborative/aiid/blob/master/site/realm/functions/apiRequest.js#L28-L35

One possible solution is to remove the `adminData` tree from the main query and fetch each user's `adminData` individually.

pdcp1:
Closing this since it was already merged into Staging

PR #2274, opened 2023-09-07
Refactor Admin users table to load the Custom Data on demand

pdcp1:
This fixes https://github.com/responsible-ai-collaborative/aiid/issues/2272

Following this approach: https://github.com/responsible-ai-collaborative/aiid/issues/2272#issuecomment-1708751868

Now we load the table with the Users DB data and in parallel, we load the Realm User data (email, creation date, last authentication date)

## Preview link
https://deploy-preview-2274--staging-aiid.netlify.app/admin/


site/gatsby-site/src/components/users/UserCreationDateCell.js
kepae, line 1: Any reason to use separate files with default exports rather than one file with named exports for these components? They use the same imports and grouping them could make it easier to maintain. Perhaps `UserInfoCells.js`?

I don't feel strongly, just wanted to understand.

site/gatsby-site/src/components/users/UserCreationDateCell.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { format } from 'date-fns';
+import { FIND_USER } from '../../graphql/users';
+import { useQuery } from '@apollo/client/react';
+import { Spinner } from 'flowbite-react';
+
+const UserCreationDateCell = ({ userId }) => {
+  const { data, loading } = useQuery(FIND_USER, {
```
kepae, line 8: IIUC this will create separate parallel requests for the same user object for each of the user's creation date, email, auth date, etc fields. 

Is there a way to only request the user once and render the row information only on return?

I'm concerned about creating a large number of requests overall to Atlas, too, in case the traffic is "bursty" and gets limited, or for cause of billing.

What do you think? Can you measure/count the outgoing requests?

site/gatsby-site/src/components/users/UserCreationDateCell.js
pdcp1: I use separate files to have individual components in case we want to reuse them, but in this case, they won't be reused since all are part of the Users table. So, I refactored it to be all in the same file `UserInfoCells.js` as you suggested.

site/gatsby-site/src/components/users/UserCreationDateCell.js
kepae: (If it improves our performance today and doesn't cause issues, I would be happy to leave this as a future optimization. We shouldn't block fixing the test on this further optimization.)

pdcp1: The `useQuery` hook has an [in-memory cache by default](https://www.apollographql.com/docs/react/data/queries/#cache-first) (`cache-first`). So we only call the Atlas graphql API once for each user.
This also improves the pagination performance since it won't refetch from the server if you go back to a previous page.

Screenshot 2023-09-08 at 11 59 02

kepae (approved):
stellar

PR #2287, opened 2023-09-14
Feature/prismic docs

clari182:
- Part of https://github.com/responsible-ai-collaborative/aiid/issues/2048

Allows docs to be created via Prismic.
Docs are essentially pages, like about, contact & follow etc

This fetches docs from prismic and available doc mdx files present. It will only create MDX pages with available slugs.
For example if the contact page is in Prismic with slug `/contact` and an MDX file exists with slug `/contact` it will only use the prismic one.


pdcp1 (changes requested):
The code looks good. The only thing that I think is missing is the test of this new feature.

pdcp1 (approved):

cesarvarela (approved):

clari182:
@cesarvarela can we merge this?

Issue #2240, opened 2023-08-17
Create and use default user for promoting submissions


kepae:
Currently, the logic that promotes submissions defaults to using Sean's user ID for the **editor** field if it's unset (it's been like this for a while; in the past it was just his name). To my knowledge/code search, no other place uses his user ID like this.

https://github.com/responsible-ai-collaborative/aiid/blob/418a34d7e3461a02e6f7d6af4c7ca8373f1f806a/site/realm/functions/promoteSubmissionToReport.js#L29

If we need to fallback to satisfy the expectation that every submission promoted to a report has an editor ID, we should create a default user with no permissions and substitute it there.

pdcp1:
@kepae I need your help on this. In order to change this default user, we should create a new user manually from the **Production** Atlas UI: `App Services` > `App Users` > `Add New User`.
Fill in any email that you consider appropriate for a default user and a password.
After creating the user, filter the list by email and copy the user Id here.

Then, I will change the code to use that user.

kepae:
Thanks for your patience, just confirmed the new account:

Default user:
email: `defaultuser at incidentdatabase dot ai`
_id: `65031f49ec066d7c64380f5c`

I'll add this to the wiki, perhaps in a "special public values" section.

pdcp1:
Closing this since it was merged into Staging

PR #2289, opened 2023-09-14
Change Sean user to a new default user

pdcp1:
This closes https://github.com/responsible-ai-collaborative/aiid/issues/2240

- Change Sean user  to `65031f49ec066d7c64380f5c` default user
- Update default user name to `Anonymous User`

### DIFF
--- a/site/gatsby-site/migrations/2023.09.14T18.44.03.create-default-user.js
+++ b/site/gatsby-site/migrations/2023.09.14T18.44.03.create-default-user.js
@@ -1,5 +1,3 @@
-const { ObjectId } = require('bson');
-
 const config = require('../config');
 
 const userId = '65031f49ec066d7c64380f5c'; // Default user. For more information refer to the wiki page: https://github.com/responsible-ai-collaborative/aiid/wiki/Special-non%E2%80%90secret-values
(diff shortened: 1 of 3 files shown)


cesarvarela (commented):
This will only affect old `submissions` that do not have the `editors` field set, right?

pdcp1:
Yes, this only affects when an editor promotes any `submissions` that do not have the `incident_editors` field set (submitted anonymously).
The migration is just to set the default user's first and last name.

site/gatsby-site/migrations/2023.09.14T18.44.03.create-default-user.js
```diff
@@ -0,0 +1,79 @@
...
+
+  const last_name = 'User';
+
+  await client.connect();
+
+  const usersCollection = client.db(config.realm.production_db.db_custom_data).collection('users');
+
+  const defaultUser = await usersCollection.findOne({ userId: userId });
+
+  if (defaultUser) {
```
cesarvarela, line 23: Is this because it already exists in staging and not production? It is a bit weird to have a condition like this in a migration 🤔

cesarvarela (commented):
Last question before merging!

site/gatsby-site/migrations/2023.09.14T18.44.03.create-default-user.js
pdcp1: Actually the opposite, this user already exists on Prod (because Kevin created it manually to get the user ID, which is correct), but not in Staging or any local env.

site/gatsby-site/migrations/2023.09.14T18.44.03.create-default-user.js
```diff
@@ -0,0 +1,79 @@
...
+      { _id: defaultUser._id },
+      {
+        $unset: {
+          first_name: '',
+          last_name: '',
+        },
+      }
+    );
+
+    console.log(`Rollback userId: ${defaultUser.userId}, ${updateResult.modifiedCount}`);
```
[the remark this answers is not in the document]
pdcp1: The `await usersCollection.updateOne` call above always returns a non-undefined value of type [UpdateResult](https://www.mongodb.com/docs/manual/reference/method/db.collection.updateOne/#returns). There is no need to add an object check.

site/gatsby-site/migrations/2023.09.14T18.44.03.create-default-user.js
cesarvarela: What about restoring staging to prodction before merging this one so we are sure it works?

Otherwise will be testing it live.

site/realm/functions/promoteSubmissionToReport.js
```diff
@@ -26,7 +26,7 @@ exports = async (input) => {
       const lastIncident = await incidents.find({}).sort({ incident_id: -1 }).limit(1).next();
 
       const editors = (!submission.incident_editors || !submission.incident_editors.length)
-        ? ['619b47ea5eed5334edfa3bbc']
+        ? ['65031f49ec066d7c64380f5c']
```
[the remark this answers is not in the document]
pdcp1: Done 👍

site/gatsby-site/migrations/2023.09.14T18.44.03.create-default-user.js
pdcp1: I took that approach into consideration, but the Github Action for restoring the database doesn't include the users' collection (`customData` db). We should restore the users' collection manually.
I'm OK doing it that way.

cesarvarela: I prefer to do that manually and get the peace of mind of testing this first in staging

site/gatsby-site/migrations/2023.09.14T18.44.03.create-default-user.js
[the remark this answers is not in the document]
kepae: (Oops, you're absolutely right -- I didn't see the indentation properly, thought it was outside the `if()`.)

site/gatsby-site/migrations/2023.09.14T18.44.03.create-default-user.js
kepae: +1

pdcp1: Done! The `users` collection was restored from Production to Staging.

cesarvarela (approved):

kepae (approved):

Issue #2279, opened 2023-09-11
Generate new Rollbar token and update ENV variables


pdcp1:
I received this email from Rollbar, so I assume this will affect Sean Rollbar account as well.
We should regenerate a new token and update all ENV variables (local, staging, and Production).
I don't have access to Rollbar to do it, so please someone assign this to yourself if you have it.

```
Dear pdcp1 team,

On September 6, 2023 at 8:28AM PDT, we identified an irregularity in our data warehouse query logs. We immediately performed an initial forensic analysis, determining what had occurred and conducting initial assessment of the impact.

Our preliminary analysis established that a cloud platform service account which only had access to our data warehouse was used by an unauthorized party. When we became aware of this access we disabled the service account and began analyzing what actions had been taken by the unauthorized party. The party first tried to launch compute resources, and after that failed for lack of permission, they accessed the data warehouse and ran searches that suggested they were interested in Bitcoin wallets or other cloud credentials. Our investigation indicates this access occurred from August 9, 2023 to August 11, 2023. We will also engage a third-party forensic consultant to assist us in verifying these findings, and that work is ongoing.

Our analysis of the incident continues, but we are contacting you now because our initial forensic analysis indicates the unauthorized party accessed data about your account, including:

- Usernames and user email addresses
- Account name
- Project and environment names
- Project access tokens
- Project service link configuration


We are taking the following actions:
- Project access tokens with 'read' or 'write' scope: these tokens could allow access to your Rollbar project data (occurrences, items, etc). These tokens have been expired. You can refresh each token using the Rollbar UI, API, or Terraform provider.
- Project access tokens with 'post_server_item' scope: these tokens could allow access to send data into your project (but not read data). **These tokens will be expired in 30 days.**


Although our investigation is ongoing, we hold the security of our customers data paramount and are therefore writing to promptly notify you of the discovery and the steps we have taken.

Please feel free to contact us using the information below if you have any additional questions or if there is further information we can provide.

Sincerely,
Brian Rue, Co-founder and CEO

brian@rollbar.com

Rollbar Security
security@rollbar.com

Rollbar Support
support@rollbar.com

```

smcgregor:
@kepae I added you to the rollbar account as an owner. It looks like we need to rotate the tokens and Rollbar will have taken care of notifying the developers that have accounts there?

kepae:
[...]

PR #2295, opened 2023-09-18
Deploy to Production

kepae:


smcgregor:
🫡 (yay to no longer being the default user)

cesarvarela:
I've been playing around with the new Submission UI and got an error when assigning an editor, but it isn't showing here:

https://rollbar.com/seanbmcgregor/all/items/?sort=%5Bobject%20Object%5D&status=active&date_from=&date_to=&environments=staging&activated_to=&framework=&levels=40&levels=50&activated_from=&offset=0&timezone=America%2FMontevideo&assigned_user=&date_filtering=seen&projects=500413&query=&enc_query=

It would be nice to be sure we are properly tracking errors before merging since this is a new feature.

kepae:
@cesarvarela @clari182 Agree, we would especially like to know if editors hit problems. Why wouldn't errors get logged by rollbar?

I also don't see any errors for staging at all in rollbar. I rotated the token in https://github.com/responsible-ai-collaborative/aiid/issues/2279 but I wasn't using rollbar prior, and don't know if ifs new/old that there are no errors there.

cesarvarela:
This line sends the error to Rollbar, so my guess is that there is a token problem yup.

https://github.com/responsible-ai-collaborative/aiid/pull/2295/files#diff-4e040be5c7c9fd9f759d67e5c1f0aa4d435c8596a52c25ec4b62e9133d7ceb7cR115

@pdcp1 might know exactly why it isn't being logged.

pdcp1:
@cesarvarela You're right. Rollbar logs are not working on Staging. I'm checking what's happening.

pdcp1:
@kepae Just to confirm, which Rollbar project did you get the token from?
Is it https://app.rollbar.com/a/seanbmcgregor ?

kepae:
@pdcp1 Yes, exactly. However, I'm thinking I used the wrong "scope" for token -- they work differently between projects and accounts. I used `write`, not `post_client_item`. I'll update accordingly.

kepae:
I switched from `write` to `post_client_item` in staging, prod, and the empty netlify environments. I'll kick off another staging build to deploy that in staging.

pdcp1:
Update from Rollbar issue.
I found that we have a console error on Staging (the same happens on my localhost) related to CORS.

```
Access to script at 'https://cdn.rollbar.com/rollbarjs/refs/tags/v2.26.1/rollbar.min.js' from origin 'http://localhost:8000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```
image

kepae:
@cesarvarela Want to try and reproduce errors again in staging, now that rollbar is back to life?

cesarvarela:
Couldn't reproduce it.

But I'm noticing now that invalid fields are not getting highlighted in red. In this case, the Incident Title field should have the red outline:

image

Like it does in the submission form:


image

kepae:
Hmm. I think it's best to fix that before releasing, since it's core to the editor checks.
On the other hand, we can provide an FYI to editors and go ahead -- provided we fix it later this week.

@clari182 your thoughts?

clari182:
@kepae I'll provide a fix shortly

PR #2297, opened 2023-09-19
Reduce width of CSET charts, add lines, tweaks

lmcnulty:
After:

image

Before:

image

---

Most of the diff is from moving the GroupBarChart component to its own file. Right now it appears on `/taxonomy/CSETv1/` as well as `/summaries/cset-chart/`. I made the latter page while CSET was asking for graphs but hadn't specified where they wanted them to appear. I think we can probably remove this page altogether, but we should check with them first.


kepae:
@lmcnulty is this ready for review? please feel free to roll the dice/target a victim to review. :-)

clari182 (approved):

PR #2306, opened 2023-09-22
Fix classifications test

cesarvarela:


site/gatsby-site/netlify.toml
```diff
@@ -11,7 +11,7 @@ TERM = "xterm"
     enable = false
   [plugins.inputs.postBuild]
     enable = true
-    record = false
+    record = true
```
kepae, line 14: Do you expect this to slow down our tests? If so, is it worthwhile to enable for this week and then disable soon, or?

cesarvarela: I've disabled it for $now would like to get everything passing first

PR #2316, opened 2023-09-27
Stay on current page when claiming/unclaiming

### LINKED ISSUE #2309: Submission Queue - clicking "claim" changes the sort
From @smcgregor:
> Clicking claim changes the sort, which tends to make the thing you just claimed disappear

### LINKED ISSUE #2310: Submission Queue - Easy "unclaim" button
Nice to have: after an editor "claims" a submission, adding a live "unclaim" button immediately in its place.

(h/t @smcgregor)

clari182:
- Stays on current page when claiming or unclaiming
- Adds unclaim functionality

- closes #2309 
- closes #2310


site/gatsby-site/react-table-config.d.ts
```diff
@@ -0,0 +1,120 @@
+import {
```
pdcp1, line 1: Is this file needed? I'm curious about the file path `site/gatsby-site`. It looks inappropriate

pdcp1 (changes requested):

site/gatsby-site/react-table-config.d.ts
clari182: Yes, it's to avoid [this](https://stackoverflow.com/questions/60006656/react-table-not-assignable-to-type-partialtablestateobject) issue, following the solution provided by the package [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-table/Readme.md).

pdcp1: OK, I got it! I wonder if we can watch that issue in the `react-table` for a possible official fix.

pdcp1 (approved):

clari182:
@cesarvarela can you review this PR please? Thanks!

site/gatsby-site/react-table-config.d.ts
cesarvarela: I agree with the weird location comment, I think the convention is to have them inside a `/types` folder.

cesarvarela (commented):
I think this change is big enough to update or add a new test.

pdcp1:
@clari182 I think we should update this branch with the Staging code to see if the tests pass.

Issue #2312, opened 2023-09-26
Add test that checks Rollbar is logging errors


pdcp1:
In order to prevent [this](https://github.com/responsible-ai-collaborative/aiid/pull/2295#issuecomment-1736356902) situation in the future, we should add a test that check if Rollbar is working correctly.

pdcp1:
Closing this it was already merged into Staging

PR #2320, opened 2023-09-28
Add test for Rollbar logging on the website

pdcp1:
This closes https://github.com/responsible-ai-collaborative/aiid/issues/2312


cesarvarela (commented):
I thought of using the `unit` folder for unit tests, and this is an integration test.

Other than that 🚢 

(I know that it is inside an `e2e` folder, which makes everything more confusing, but Cypress doesn't make it easy to change.)

PR #2323, opened 2023-09-29
Fetch path from social media items

### LINKED ISSUE #2321: Fix RSS footer icon link
If the footer is getting from Prismic, the RSS icon link is not working.

image

I assume that this is because of the way that we store it on Prismic:

image

We should change that item to use the `url` field pointing to the Production RSS: https://incidentdatabase.ai/rss.xml

clari182:
- closes #2321

### DIFF
--- a/site/gatsby-site/src/components/layout/Footer.js
+++ b/site/gatsby-site/src/components/layout/Footer.js
@@ -48,6 +48,7 @@ export default function Footer() {
                 url {
                   url
                 }
+                path
               }
             }
           }


site/gatsby-site/src/components/layout/Footer.js
```diff
@@ -48,6 +48,7 @@ export default function Footer() {
                 url {
                   url
                 }
+                path
```
pdcp1, line 51: Why do we need to add the `path` field? The `RSS` item is using the `path` field or the `url` field?

pdcp1 (changes requested):

site/gatsby-site/src/components/layout/Footer.js
clari182: RSS is using the path because it's not an external URL

pdcp1 (approved):

PR #2324, opened 2023-09-29
Create a page for every report, whether it is linked to an incident or not.

### LINKED ISSUE #2229: Improve report pages: Show page for each `/report/*`, regardless of incident associations
Currently, simple raw report pages are only generated for reports that are **not** associated with incidents.
For example, report 2973 generates a page at https://incidentdatabase.ai/reports/2973/ as it does not support a minted incidents (but rather exists as an *issue report*), but report 1 does not generate such a page(https://incidentdatabase.ai/reports/1/) because it is linked to an incident (https://incidentdatabase.ai/cite/1/).

This behavior may simply be because the report pages weren't prioritized, or because this is a built-in structure to avoid linking a report when an incident page should instead be the focus. However, if a report later becomes part of an incident, the report URL is now a dead end. (should it be that way? do protest.)

Creating a page for every report is also useful in the case where future discovery/explore features may surface reports independently of incidents (e.g. #2047). A useful page would also contain pointers to incidents that are based on the report.

Goals:
- [ ] every report URL endpoint results in a report page
- [ ] report pages surface connection/membership to the incident(s) they support

cesarvarela:
closes #2229


site/gatsby-site/cypress/e2e/integration/report.cy.js
```diff
@@ -7,6 +7,24 @@ describe('Report pages', () => {
...
+  });
+
+  it('Should not display associated Incidents', () => {
+    cy.visit(`/reports/2974/`);
+
+    cy.contains('Associated Incidents').should('not.exist');
+  });
+
   it('Should always be expanded', () => {
     cy.visit(url);
```
kepae, line 29: reading the test linearly is a little confusing because the `url` identifier is bound to a test constant above.

Could you use and re-use the same identifiers?

Maybe: `report_with_incidents` + `report_no_incidents` ?

site/gatsby-site/src/templates/report.js
```diff
@@ -84,7 +85,35 @@ function ReportPage(props) {
...
+
+            <Card.Body>
+              {incidents.nodes.map((incident) => (
+                <Link to={`/cite/${incident.incident_id}`} key={incident.incident_id}>
+                  <h4 className="mb-2 text-md font-bold tracking-tight text-gray-900 dark:text-white">
+                    <span className="text-sm">
+                      <Trans>Incident {{ id: incident.incident_id }}</Trans>
+                    </span>
+                    <span className="ml-2 bg-red-100 text-red-800 text-xs font-semibold mr-2 px-2.5 py-0.5 rounded dark:bg-green-200 dark:text-green-900">
+                      <Trans ns="entities" count={incident.reports.length}>
```
kepae, line 104: `ns="reports"`? or am I about to learn something?

kepae (changes requested):
on text hover, there's a blue underline. I like underlines on hover but seems inconsistent for our style at the moment.

image

site/gatsby-site/src/templates/report.js
kepae: (but this info tag was a nice touch, user was delighted)

kepae:
> on text hover, there's a blue underline. I like underlines on hover but seems inconsistent for our style at the moment.
> image

Please ignore. Now when I visit other pages linking to incidents I notice it.
e.g.  links to incidents in entity pages.

site/gatsby-site/src/templates/report.js
cesarvarela: Those namespaces refer to the files in this folder: https://github.com/responsible-ai-collaborative/aiid/tree/master/site/gatsby-site/i18n/locales

Only the `translation.json` file is loaded by default, and all others are requested on demand. If a text string is used in multiple places, we place it there. I made some rearrangements here to reflect this.

Issue #2108, opened 2023-06-20
Use new report linked query 


clari182:
Following #2102 improvement, the new logic needs to be used in the following places:

- [x] CreateCitationPages - https://github.com/responsible-ai-collaborative/aiid/pull/2111
- [x] Incident Form - https://github.com/responsible-ai-collaborative/aiid/pull/2338
- [x] Random Incident Carousel - https://github.com/responsible-ai-collaborative/aiid/pull/2339
- [x] Leaderboards (Original/Unique) - https://github.com/responsible-ai-collaborative/aiid/pull/2327
- [x] summaries/flagged - https://github.com/responsible-ai-collaborative/aiid/pull/2338
- [x] summaries/incidents
- [x] incidentsOverTime -https://github.com/responsible-ai-collaborative/aiid/pull/2338
- [x] entities - only uses length of reports

PR #2327, opened 2023-10-02
Use new reports query link to fetch leaderboards

clari182:
- Part of #2108


cesarvarela (commented):
Something is going on here if you compare the numbers to production:

image

PR #2333, opened 2023-10-04
Checklisting – add authentication

lmcnulty:
Resolves #2300.

PR #2336, opened 2023-10-05
Checklists – Add owner to UI

lmcnulty:
Building on #2333


lmcnulty:
I don't think this is worth keeping separate from https://github.com/responsible-ai-collaborative/aiid/pull/2333.

PR #2339, opened 2023-10-06
Update reports query for random incidents carousel

clari182:
- part of #2108


pdcp1 (changes requested):
We should choose one name for this component since we use `Random Incidents` and `Random Reports`. The same happens with the translations for those texts.
For example: `src/components/landing/RandomReports.js` uses `RandomIncidentsCarousel` component.

pdcp1:
@clari182 Could you please add a test for this component on landing.cy.js test file?

pdcp1 (changes requested):
@clari182 Could you please add a test for this component on landing.cy.js test file?

pdcp1 (approved):

PR #2348, opened 2023-10-13
Fetch each user data on load to enable sort

### LINKED ISSUE #2318: Admin – Email filter broken
Trying to search/filter by email in the Admin interface yields no results. Entering a single character or any substring present in the emails returns zero hits.
Also seems that the date fields are not searchable, but those are less a priority.

However, searching on other fields, such as First or Last name, works as expected; substrings return results.

image

clari182:
- closes #2318


pdcp1 (changes requested):
We should add a test to verify that the email filter is working. The same for the rest of the filters.

pdcp1 (approved):

PR #2358, opened 2023-10-17
Upgrade sidebar zindex

### LINKED ISSUE #2345: Collapsed sidebar tooltip appearing under page content
image

image

image

clari182:
- closes #2345


lmcnulty (commented):
Unfortunately, this un-fixes #2164. I'm not sure there's a way to solve both of these without rethinking the whole app's use of stacking contexts and z-indices.

clari182:
Oof, good catch! I'll invest some time to think how to fix both.

PR #2369, opened 2023-10-23
Checklisting minor UI fixes

lmcnulty:
It would be good to get these in before the demo at the MTF demo tomorrow if that's possible.


kepae:
As discussed on Slack, the deploy preview isn't whitelisted for user data access (which the checklists app requires to fully function) which is alright, but it fails to render entirely in the deploy environment. We should fix that.

@lmcnulty Would you please track an issue for that?

PR #2372, opened 2023-10-23
Remove duplicated query and limit users to 9999

### LINKED ISSUE #2353: Edit Incident – Blank page error
Specifically on Incident 586, per Danny:

> When I click "Edit Incident" on that page, something similar to the "Your Account" glitch occurs; the page attempts to load and then suddenly goes blank.

I can reproduce this in production currently:
https://incidentdatabase.ai/incidents/edit/?incident_id=568
Has failed requests with error:
`TypeError: p.users.find(...) is undefined`

clari182:
- closes #2353 

On staging this can be tested on this incident: https://deploy-preview-2372--staging-aiid.netlify.app/incidents/edit/?incident_id=16 since it was failing for the same reason


site/gatsby-site/src/components/users/UsersField.js
```diff
@@ -31,11 +31,16 @@ export default function UsersField({ id, name, placeHolder = '' }) {
...
       setSelected((selected) =>
-        selected.map(({ id }) => {
-          const { userId, first_name, last_name } = data.users.find((user) => user.userId == id);
+        selected
+          .map(({ id }) => {
+            const user = data.users.find((user) => user.userId == id);
 
-          return { id: userId, first_name, last_name };
-        })
+            if (!user) return null;
```
cesarvarela, line 38: This will map `null` objects in the array. I think it should filter then instead.

cesarvarela (commented):
Maybe unrelated but got an error when saving the edited incident:

image

clari182:
@cesarvarela can you check if this is still happening? I couldn't reproduce it.

cesarvarela:
@clari182 try adding another editor

clari182:
@cesarvarela ready for another review. Issue was related to user being unhautenticated

site/gatsby-site/src/components/users/UsersField.js
```diff
@@ -31,11 +31,13 @@ export default function UsersField({ id, name, placeHolder = '' }) {
   useEffect(() => {
     if (data?.users) {
       setSelected((selected) =>
-        selected.map(({ id }) => {
-          const { userId, first_name, last_name } = data.users.find((user) => user.userId == id);
+        selected
+          .filter(({ id }) => data.users.some((user) => user.userId == id))
```
cesarvarela, line 35: This may not be necessary. On other than dev environments, it would never happen 🤔

Issue #2386, opened 2023-10-30
Checklisting – Main Dashboard


kepae:
Currently, all checklists are publicly readable from the main `/apps/checklists/` endpoint. While this works today for demo purposes, we should come up with the UI and user expectations of the entry point/dashboard.

Ideally, the dashboard of the checklisting application

- [ ] shows a user only their own checklists
- [ ] has sorting functions -- creation, last edit, alphabetical by system name
- [ ] indicates the number of risks in each checklist (and maybe a notion of progress?)

There is more to be said about making individual user-owned checklists private or public. But for initial launch, we should default to closed.

PR #2394, opened 2023-11-04
Hide unowned checklists, sort index, convert to using GraphQL API

lmcnulty:
- Now queries only the user-owned checklists on the index page.
- Now shows generated risks on the checklist cards, not just manual risks, which shows progress as the risk icons get colored in. There's also now a message if no risks are found.
- Fixes inconsistency between GraphQL and JSON API (fields returned, failure condition) and removes REST API entirely.
  - In this PR because we're using the GraphQL API on the index page and it needed to be changed to work.
- Allows sorting index page by newest, oldest, last-edited, and alphabetical.
- Adds `date_created` and `date_updated` to implement the above.

Per #2386.


kepae:
Drive-by thought -- I think the important thing is that un-owned checklists are never returned to the client. I should have clarified this in https://github.com/responsible-ai-collaborative/aiid/issues/2386: 

"shows a user only their own checklists" -> "application only retrieves and shows checklists owned by the user"

What do you think? This means work on the resolver ("server") side.

lmcnulty:
@kepae Is this a privacy concern or an efficiency concern? If it's about privacy, then we also need to store the checklists in the `customData` database with the user data, since dumps of `aiidProd` are publicly accessible. Sean seemed ok with not treating checklists as private data, but that may be different at this phase.

kepae:
My concern is about efficiency, as each time this page loads it is potentially requesting O(1000) checklists.
It's a dynamic page and ultimately slower than other large exhaustive but static pages such as the [full incident list](https://incidentdatabase.ai/summaries/incidents/).

Then again, this could just be premature optimization -- payload size likely isn't a problem for a while.
If it means migrating and storing checklists in `customData`, then let's hold off.

I'm not concerned about privacy because yes, all user-generated risk checklists are intended to be public, and that hasn't changed. Though this does create a separate reminder -- we should make that transparent to users/remind them that checklists are public artifacts: https://github.com/responsible-ai-collaborative/aiid/issues/2396

pdcp1:
@lmcnulty Do you have a preview link for a working environment? I try to test create a new checklist using https://deploy-preview-2394--staging-aiid.netlify.app but I got this error:

```
{
    "data": null,
    "errors": [
        {
            "message": "Cannot query field \"date_created\" on type \"Checklist\".",
            "locations": [
                {
                    "line": 5,
                    "column": 5
                }
            ]
        },
        {
            "message": "Cannot query field \"date_updated\" on type \"Checklist\".",
            "locations": [
                {
                    "line": 6,
                    "column": 5
                }
            ]
        }
    ]
}
```

site/gatsby-site/src/components/checklists/ChecklistsIndex.js
```diff
@@ -47,43 +122,56 @@ const ChecklistsIndex = ({ users }) => {
...
-                  });
-                  return;
-                }
-                window.location = '/apps/checklists?id=' + newChecklist.id;
-              }}
-            >
-              <Trans>New</Trans>
-            </Button>
-          )}
+          <div className="flex gap-2 items-center">
```
pdcp1, line 125: We should use the `<ListSkeleton />` component instead of `<Spinner />` component in the loading state. We should also improve the empty list layout, maybe with an indicating message of the empty state.

image

site/gatsby-site/src/components/checklists/ChecklistsIndex.js
```diff
@@ -47,43 +122,56 @@ const ChecklistsIndex = ({ users }) => {
...
+            <label htmlFor="sort-by">Sort</label>
+            <Select id="sort-by" onChange={(evt) => setSortBy(evt.target.value)}>
+              <option value="alphabetical">Alphabetical</option>
+              <option value="oldest-first">Oldest First</option>
+              <option value="newest-first">Newest First</option>
+              <option value="last-updated">Last Updated</option>
+            </Select>
```
pdcp1, line 132: The text of this Select component overlaps the selector arrow
image

site/realm/graphql/custom_resolvers/query_risks.json
```diff
@@ -140,6 +146,15 @@
                 },
                 "tag": {
                     "bsonType": "string"
+                },
+                "tags": {
+                    "bsonType": "array",
+                    "items": {
+                        "bsonType": "string"
+                    }
+                },
```
pdcp1, line 155: Is it OK to have both `tag` and `tags` fields?

pdcp1 (changes requested):

site/gatsby-site/src/components/checklists/CheckListForm.js
```diff
@@ -433,6 +443,7 @@ const searchRisks = async ({
   setAllPrecedents,
```
kepae, line 443: Please update the (very nice) function documentation above reflecting the use of the GraphQL API.

site/gatsby-site/src/components/checklists/ChecklistsIndex.js
```diff
@@ -14,28 +15,102 @@ import {
...
+  //     title
+  //   }
+  //   GMFKnownAITechnicalFailureUnderspecification:
+  //   risks(input: {tags: ["GMF:Known AI Technical Failure:Underspecification"]}) {
+  //     tags
+  //     title
+  //   }
+  // }
+  const riskQuery = skipRisksQuery
+    ? 'query { thisWontRunBut { useQueryMakesUseProvideValidGraphQL } }'
```
kepae, line 73:
```suggestion
    ? 'query { thisWontRunBut { useQueryMakesUsProvideValidGraphQL } }'
```
Please consider an inline comment above this.

Help me understand: why use a conditional here and create a mock/dud query, and not conditionally skip the `useQuery` hook below, or rely on `{ skip: skipRisksQuery }` ?

kepae (commented):
First pass. There's a good amount of stuff here.

site/gatsby-site/src/components/checklists/ChecklistsIndex.js
lmcnulty: React hooks can't run conditionally. We are relying on `{ skip: skipRisksQuery }`, but it will still throw an exception if first parameter isn't graphql.

site/realm/graphql/custom_resolvers/query_risks.json
lmcnulty: `tag` is legacy from the old JSON api. I don't think its necessary anymore, but I talked with Kevin and the plan is to remove it until we have more tests. #2442

site/gatsby-site/src/components/checklists/ChecklistsIndex.js
lmcnulty: Fixed.

image

site/gatsby-site/src/components/checklists/ChecklistsIndex.js
lmcnulty: Skeletons:

image


Empty list:

image

kepae (approved):
LGTM! I'll wait for @pdcp1 to review the other feedback, and @lmcnulty can tell me when it's good to merge.
We can add this to the current production release and unblock other checklisting PRs waiting.

pdcp1:
The code and the changes requested LGTM 👍 
We should wait for a successful build before merging. I just launched a new build with the latest code: https://app.netlify.com/teams/responsible-ai-collaborative/builds/65721d9f9850ae0333a91344

pdcp1 (approved):

kepae:
@lmcnulty you may have to merge staging into this PR to update tests that are failing in the current build:
https://app.netlify.com/teams/responsible-ai-collaborative/builds/65721d9f9850ae0333a91344

```
3:01:33 PM:   1) Entities page
3:01:33 PM:        Should display Entity responses:
3:01:33 PM:       Timed out retrying after 4000ms
3:01:33 PM:       + expected - actual
3:01:33 PM:       -"3 Incident responses"
3:01:33 PM:       +"2 Incident responses"
3:01:33 PM:       
3:01:33 PM:       at Context.eval (webpack://gatsby-gitbook-aiid/./cypress/e2e/integration/entities.cy.js:63:49)
```

I'm noting this as a developer pain. An unrelated test failure is stopping the deploy and testing. This should be mitigated when we use a github runner for building, then separate testing to stay on GH and push deploy preview to Netlify regardless.

lmcnulty:
@kepae I think this is ready to merge, although I can't test that much functionality since I get the origin forbidden error and I don't have the ability to add allowed origins on staging.

pdcp1:
> @kepae I think this is ready to merge, although I can't test that much functionality since I get the origin forbidden error and I don't have the ability to add allowed origins on staging.

@lmcnulty I just added the URL to the allowed origins URL list on Staging, now you should be able to test

lmcnulty:
@pdcp1 For `https://deploy-preview-2394--staging-aiid.netlify.app`? I'm still getting origin forbidden.

pdcp1:
> @pdcp1 For `https://deploy-preview-2394--staging-aiid.netlify.app`? I'm still getting origin forbidden.

@lmcnulty Yes. For some reason (maybe a recent Staging deployment) that URL was deleted from the allowed origins on Staging. Now is set again. Please try

lmcnulty:
Ok, I still can't create checklists because the `date_created` and `date_updated` fields aren't in the checklists schema on staging. I'll create a deploy preview from my dev environment https://github.com/lmcnulty/aiid/pull/30.

Issue #2335, opened 2023-10-04
User Accounts – Implement flow for re-sending verification emails


kepae:
Currently, when a user tried to sign-in without having verified their account, they only receive an error toast stating "Confirmation required".

At this point, there should be a way for a user to request to re-send the verification email, in the event that the original email never reached them or can't find it. (In practice, this is likely a new verification email if the token has expired.)

pdcp1:
Closing this since it was already merged into Staging

PR #2397, opened 2023-11-06
Add onResendConfirmationClick function on the toast message

pdcp1:
This closes https://github.com/responsible-ai-collaborative/aiid/issues/2335

## Steps to test
- Signup with a new user
- You should receive the account verification email
- Try to login with that user without confirm the account via email
- A toast message appears with the option to "Resend Verification email", click it
- You should receive another account verification email


cesarvarela:
We should add captchas and take some abuse prevention measures but it is another issue.

PR #2398, opened 2023-11-07
Feature discover tree

### LINKED ISSUE #2330: Discover – Improve Classification filtering UX
Currently, taxonomy classifications are present as filter facets in the discover app. However, there are a few UX improvements to be made.

* The `Classifications` filter currently displays classifications in ascending alphabetical order of the taxonomy namespace, and within that, descending number of results by stringified field.
    * The user does not learn what taxonomies are available, up front.
    * Reasonably, the filter view does not show _all_ fields -- but without entering any data, it will only show a limited number of fields from `CSETv0`, potentially misleading the user to think these are the only classifications to filter by.

Potential improvements:
* At the top of the filter card, display a condensed multi-option selection list of taxonomy namespaces (e.g. CSETv0, CSETv1, GMF).
* When there is no search term, sort fields first by number of results for each field, _then_ sort by alphabetical string. Depends on if this yields a nice result in practice.)

This overlaps with goals from #2047. It might be better to pause #2047 until more expressive data is in the DB, and until then, beef up/improve the Discover interface (which could also benefit from https://github.com/responsible-ai-collaborative/aiid/issues/2292).
 
________

The current view:

image

cesarvarela:
closes #2330

https://deploy-preview-59--cesarvarela-staging.netlify.app/apps/discover/


cesarvarela:
@kepae not finished yet, but some early feedback wouldn't hurt

kepae:
Looking good 🔥 . Things that could be addressed beyond what isn't clear from [your original mock](https://github.com/responsible-ai-collaborative/aiid/issues/2330#issuecomment-1749498903) (such as borders and string search for attributes):

* I think the selected refinement value should also be highlighted in the hierarchical view. The selected value gets moved to the first position (+) and also highlighted above (+), but should be highlighted easier to find for the user after they've selected it here, too.
image

And re-including here (from https://github.com/responsible-ai-collaborative/aiid/issues/2330#issuecomment-1749710673):

> * If selected facets are pulled out above to be readable (+1), how about including an "x" to dismiss on the left-hand side?
> 
> * What do you think about being able to add `NOT`-type query refinements, and pulling them out as red-colored facets? It would help with finding differences in taxonomy coverage. Then users could perform: `"CSETv0 : Near Miss : *" && NOT "CSETv1 : Harm Level : Near Miss"`

(Depending on how refinements are applied, it may not be worth trying to do `NOT` refinements in this PR.)

Issue #2404, opened 2023-11-10
Submissions - too many emails being sent


clari182:
Editors are receiving emails from submissions they didn't submit

smcgregor:
I am also receiving duplicate email notifications for the new incidents. I first received incidents 600 through 603 `Nov 10, 2023, 10:57 PM`, then again today at `10:52 AM`

I am going to stop builds from running until this is diagnosed.

Adding the critical label.

smcgregor:
I checked the builds. Last night failed on the `Deploy Site` step: https://app.netlify.com/sites/aiid/deploys/654f1883958de3000880cf4f

```
10:57:28 PM: /plugins/netlify-plugin-process-notifications (onSuccess event)  
10:57:28 PM: ───────────────────────────────────────────────────────────────────
10:57:28 PM: ​
10:57:28 PM: Processing pending notifications...
10:58:46 PM: Plugin error: since "onSuccess" happens after deploy, the build has already succeeded and cannot fail anymore. This plugin should either:
10:58:46 PM: - use utils.build.failPlugin() instead of utils.build.failBuild() to clarify that the plugin failed, but not the build.
10:58:46 PM: - use "onPostBuild" instead of "onSuccess" if the plugin failure should make the build fail too. Please note that "onPostBuild" (unlike "onSuccess") happens before deploy.
10:58:46 PM: ​
10:58:46 PM: Plugin "/plugins/netlify-plugin-process-notifications" failed  
10:58:46 PM: ─────────────────────────────────────────────────────────────────
10:58:46 PM: ​
10:58:46 PM:   Error message
10:58:46 PM:   ApolloError: Error message
10:58:46 PM:   Response not successful: Received status code 400
10:58:46 PM: ​
10:58:46 PM:   Plugin details
10:58:46 PM:   Package:        /plugins/netlify-plugin-process-notifications
10:58:46 PM:   Version:        0.0.1
10:58:46 PM: ​
10:58:46 PM:   Error location
10:58:46 PM:   In "onSuccess" event in "/plugins/netlify-plugin-process-notifications" from netlify.toml
```

The subsequent build made it through this step and displayed,
```
10:51:41 AM: Processing pending notifications...
10:53:03 AM: 58 notifications were processed!
```

[...]

PR #2411, opened 2023-11-14
Remove submission notification/subscription when parent incidents

clari182:
- part of #2404 

Only creates a subscription and notification when a new issue is created. Not when there are parent issues.
This was another scenario where there might be duplicates.


cesarvarela (changes requested):
I think this should require a new test/update.

cesarvarela (approved):

PR #2422, opened 2023-11-20
Fetch articles from sites using cookies

### LINKED ISSUE #2421: Fetching info from Washington Post articles
Low priority here, but it could expedite incident reporting if "fetch info" worked for Washington Post articles (which does quite a bit of incident/issue reporting). I currently have to manually input all information from WaPo articles, which isn't a big deal, but it does add more time to the process. No rush, though -- thank you!
-d

lmcnulty:
Resolves #2421. Fetching articles from https://www.washingtonpost.com was failing because the initial request sets some cookies and then redirects to the same page. Without those cookies, the page content is not served. As far as I can tell, the parser we're using [is supposed to accept cookies](https://github.com/postlight/parser/blob/main/src/resource/utils/fetch-resource.js#L95) but in fact does not, so this handles fetching the HTML and setting the cookies manually when fetching from The Washington Post. Probably other sites will do a similar thing, but I'm not certain that this won't break parsing other sites if we enable it across the board.

### DIFF
--- a/site/gatsby-site/cypress/e2e/integration/submit.cy.js
+++ b/site/gatsby-site/cypress/e2e/integration/submit.cy.js
@@ -1798,4 +1798,44 @@ describe('The Submit form', () => {
 
     cy.contains('Please review. Some data is missing.').should('exist');
   });
+
+  it('Should fetch article', () => {
+    cy.visit(url);
+
+    cy.intercept('GET', parserURL).as('parseNews');
+
+    cy.get('input[name="url"]').type(
+      `https://www.arstechnica.com/gadgets/2017/11/youtube-to-crack-down-on-inappropriate-content-masked-as-kids-cartoons/`
+    );
+
+    cy.get('button').contains('Fetch info').click();
+
+    cy.wait('@parseNews');
+
+    cy.get('.tw-toast')
+      .contains('Please verify all information programmatically pulled from the report')
+      .should('exist');
+
+    cy.get('.tw-toast').contains('Error fetching news.').should('not.exist');
+  });
+
+  it('Should fetch article from site using cookies', () => {
+    cy.visit(url);
+
+    cy.intercept('GET', parserURL).as('parseNews');
+
+    cy.get('input[name="url"]').type(
+      'https://www.washingtonpost.com/technology/2023/02/16/microsoft-bing-ai-chatbot-sydney/'
+    );
+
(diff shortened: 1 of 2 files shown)


smcgregor:
The solution looks good to me in general, but let's not have a different fetch routine for different websites.

cesarvarela:
Agree. I would add/update some tests too.

lmcnulty:
New solution: we start by running the native fetch routine, and if it doesn't work, then we try the custom one. This should prevent any breakage for sites that already work, and it could enable fetching from any site that uses cookies in a similar way.

site/gatsby-site/cypress/e2e/integration/submit.cy.js
```diff
@@ -1798,4 +1798,44 @@ describe('The Submit form', () => {
...
+    cy.wait('@parseNews');
+
+    cy.get('.tw-toast')
+      .contains('Please verify all information programmatically pulled from the report')
+      .should('exist');
+
+    cy.get('.tw-toast').contains('Error fetching news.').should('not.exist');
+  });
+
+  it('Should fetch article from site using cookies', () => {
```
kepae, line 1822:
```suggestion
  it('Should fetch article from site using cookies as fallback', () => {
```

site/gatsby-site/src/api/parseNews.js
```diff
@@ -21,3 +22,51 @@ export default async function handler(req, res) {
...
+    if (config.cookies) {
+      parserConfig.html = await getHtmlWithCookies(url);
+    }
+
+    const article = await Parser.parse(url, parserConfig);
+
+    return article;
+  } catch (error) {
+    if (config.cookies) {
+      throw error;
```
kepae, line 39: Since this error is facing editors, is this the best way to surface the error? would a toast be more helpful to a non-dev user? And would rollbar catch this?

site/gatsby-site/src/api/parseNews.js
```diff
@@ -21,3 +22,51 @@ export default async function handler(req, res) {
 
   res.status(200).json(response);
 }
+
+const getArticle = async (url, config) => {
```
kepae, line 26: Please comment the recursive nature of this function. :-)

kepae (changes requested):

kepae:
I'll merge this now. @lmcnulty would you please click test this in staging after it deploys?

PR #2459, opened 2023-11-30
Filter unique pending notifications to avoid duplicate emails

pdcp1:
This PR solved the problem of sending duplicated emails to the same user. Issue: https://github.com/responsible-ai-collaborative/aiid/issues/2404

- We add a filter to the recipient list to avoid duplicated recipients for the same email.

## Testing notes
Preview link https://deploy-preview-32--pablocosta-staging.netlify.app/
Note that in order to trigger the `process notification` function it should be triggered manually from Pablo's Atlas account.

## Note
We still have an issue in the origin of the actual bug that inserts multiple pending notifications and subscriptions for the same user. But this can be treated in another PR.

## Duplicated pending notifications in Production:

image

## Duplicated subscriptions in Production:

image


cesarvarela (commented):
Can you add a test that has duplicated subscriptions/notifications and makes sure the emails are not sent multiple times?

pdcp1:
> Can you add a test that has duplicated subscriptions/notifications and makes sure the emails are not sent multiple times?

@cesarvarela done! please review

kepae (approved):

Issue #2473, opened 2023-12-06
Improve submission-promoted code


clari182:
We are currently using a subscription and a notification for `submission-promoted` emails.
To simplify (and avoid future issues), we should simplify this and only create a notification. Since this is only triggered once and the subscription is no longer needed.

The notification would need to have a new field called userId only used for this kind of scenario where a notification doesn't need a subscription

PR #2441, opened 2023-11-29
Remove subscriptions duplicate - migration

clari182:
Migration that removes duplicates in subscriptions collection

PR #2503, opened 2023-12-18
Re-enable notifications plug-in

### LINKED ISSUE #2404: Submissions - too many emails being sent

kepae:
Following #2476 and #2441, we should be able to re-enable notifications. It might also be a good idea to wait for #2459 and #2473.

This should close #2404.


pdcp1 (approved):
LGTM

clari182 (approved):

Issue #2468, opened 2023-12-05
Checklisting – Typing tag manually should resolve to fully-qualified tag if one matches


lmcnulty:
Right now if you select, e.g. "Code Generation" from the dropdown, it resolves to `GMF:Known AI Goal:Code Generation`. If you type "Code Generation" and press enter, it just gets stored as `"Code Generation"` which won't match anything.

PR #2505, opened 2023-12-19
Checklists – Resolve tag entry to fully-qualified when typed directly

lmcnulty:
Resolves https://github.com/responsible-ai-collaborative/aiid/issues/2468

### DIFF
--- a/site/gatsby-site/src/components/checklists/CheckListForm.js
+++ b/site/gatsby-site/src/components/checklists/CheckListForm.js
@@ -274,14 +274,7 @@ const QueryTagInput = ({
           const selectedTags = [];
 
           for (const v of value) {
-            let tag;
-
-            if (v.includes(':')) {
-              tag = v;
-            } else {
-              tag = tags.find((t) => t.replace(/.*:/g, '') == v);
-            }
-            selectedTags.push(tag);
+            selectedTags.push(v.includes(':') ? v : tags.find((t) => t.replace(/.*:/g, '') == v));
           }
           setFieldValue(id, selectedTags);
         } else {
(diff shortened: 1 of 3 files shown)


site/gatsby-site/src/components/checklists/CheckListForm.js
```diff
@@ -270,7 +270,16 @@ const QueryTagInput = ({
       value={idValue}
       options={tags.filter((tag) => include(tag.split(':')))}
       onChange={(value) => {
-        setFieldValue(id, value);
+        if (labelKey == abbreviatedTag) {
```
kepae, line 273: It's difficult for me to understand what is going on with this logic -- I had to rely on the PR title. Maybe the intention should be commented here.

In addition,`abbreviatedTag()` is a function and this branch will never be activated, IIUC:
https://github.com/responsible-ai-collaborative/aiid/blob/master/site/gatsby-site/src/utils/checklists.js#L42
And if I read the regex correctly, it removes every character in a string up until the last ":" if there is one.

So what's the condition we want?

Then, this function should also be used below, right?
```javascript
selectedTags.push(v.includes(':') ? v : tags.find((t) => abbreviatedTag(t) == v));
```

I could also use readability help on understanding what `value` is expected to be. A list of tags?

lmcnulty: I was writing this just before out meeting, in which we discussed it.

---

I rewrote for readability and added docs.

> In addition,abbreviatedTag() is a function and this branch will never be activated, IIUC:

`labelKey` is also a function – the name is misleading, but it comes from the typeahead library. It takes the underlying value for a menu option and returns what will be displayed. Resolving to fully-qualified tags is only necessary when the tags are abbreviated, so I was checking whether `abbreviatedTag` was being used. However, I now realize that this is confusing, so I've moved this functionality within the component and put it behind a `trimTaxonomy` flag.

kepae (approved):
Thank you for such thorough commenting. Future maintainers (you, me, the president) will appreciate it.

PR #2515, opened 2023-12-27
Checklists – Add dates to cards

lmcnulty:
Resolves https://github.com/responsible-ai-collaborative/aiid/issues/2447

### DIFF
--- a/site/gatsby-site/src/components/checklists/ChecklistsIndex.js
+++ b/site/gatsby-site/src/components/checklists/ChecklistsIndex.js
@@ -1,5 +1,5 @@
 import React, { useState, useEffect } from 'react';
-import { Button } from 'flowbite-react';
+import { Button, Tooltip } from 'flowbite-react';
 import Card from 'elements/Card';
 import Select from 'elements/Select';
 import { Trans, useTranslation } from 'react-i18next';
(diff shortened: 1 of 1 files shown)


site/gatsby-site/src/components/checklists/ChecklistsIndex.js
```diff
@@ -297,11 +313,43 @@ const CheckListCard = ({ checklist, setChecklists, owner }) => {
...
+                      Updated {dateUpdatedYmdHm}
+                    </>
+                  }
+                >
+                  <time dateTime={checklist.date_created}>{dateCreatedYmd}</time>
+                </Tooltip>
+              </>
+            ),
+          ]
+            .filter((e) => e)
```
kepae, line 340: why filter falsy values? are we expecting empty fragments at this point? it would hide any errors with the above logic, which should never render an empty div, IIUC.

lmcnulty: We want to display the owner's name and the creation date, joined by a bullet. But sometimes the owner hasn't entered a name in the system. I'm not sure if there will be a way for the date to be undefined in production, but it sometimes happens in development. In either case, if only one of those fields is defined, we want to display it without a bullet. The `.reduce` here is the React equivalent of `.join(' • ')`. So if we have a name and a date, we get "Alice Bobson • 2024-01-02", but if we just have a date then we get just "2024-01-02". If we didn't filter out falsies, then we would get " • 2024-01-02".

kepae (approved):

Issue #2522, opened 2024-01-03
Checklisting – Page Crashing on load for signed-in user


kepae:
While signed in to my account and after briefly flashing the checklist skeletons, the page crashes entirely. From the console:

```javascript
Uncaught TypeError: e is undefined
    V checklists.js:43
    Ze ChecklistsIndex.js:317
    Ze ChecklistsIndex.js:310
    React 7
    S scheduler.production.min.js:13
    T scheduler.production.min.js:14
    60053 scheduler.production.min.js:14
```

I can still access an empty page from anonymous users.

kepae:
I can still access and create checklists in staging, which I don't believe has too much of a relevant feature drift from production. This might be a data handling issue.

kepae:
Coming back to this with a dump of the graphQL request and response under my account that leads to this error.

```json
"errors": [
    {
      "message": "cannot compare to undefined",
      "path": [
        "CSETvInfrastructureSectorsTransportationGMFKnownAIGoalContentSearchGMFKnownAIGoalTranslationGMFKnownAITechnologyLanguageModelingGMFPotentialAITechnologyTransformerTraversal"
      ]
    },
    {
      "message": "cannot compare to undefined",
      "path": [
        "AggregationGMFKnownAIGoalContentSearchGMFKnownAIGoalTranslationGMFPotentialAITechnologyClassificationGMFPotentialAITechnologyTransformer"
      ]
    }
  ],
```

---

Request:
```
{
GMFKnownAIGoalContentSearchGMFKnownAIGoalTranslationGMFPotentialAITechnologyClassificationGMFPotentialAITechnologyTransformer: risks(
    input: {tags: ["GMF:Known AI Goal:Content Search", "GMF:Known AI Goal:Translation", "GMF:Potential AI Technology:Classification", "GMF:Potential AI Technology:Transformer"]}
  ) {
    tags
    title
    __typename
  }
  AggregationGMFKnownAIGoalContentSearchGMFKnownAIGoalTranslationGMFPotentialAITechnologyClassificationGMFPotentialAITechnologyTransformer: risks(
    input: {tags: ["GMF:Known AI Goal:Content Search", "GMF:Known AI Goal:Translation", "Aggregation", "GMF:Potential AI Technology:Classification", "GMF:Potential AI Technology:Transformer"]}
  ) {
    tags
    title
    __typename
  }
[...]

Issue #2370, opened 2023-10-23
Checklisting – Blanking on failure to load data


lmcnulty:
See https://github.com/responsible-ai-collaborative/aiid/pull/2369#issuecomment-1775702195.

kepae:
This is similar behavior to the crash seen in #2522; we should try to prevent a total page crash at all costs, and keep the failure to the inner components.

Issue #2315, opened 2023-09-27
Checklisting – Improve error handling


lmcnulty:

kepae:
This would help in #2522.

PR #2524, opened 2024-01-04
Checklists – Improve error handling, add null checks

lmcnulty:
I wasn't able to determine the cause of https://github.com/responsible-ai-collaborative/aiid/issues/2522, but I tried to make it such that missing data in general won't cause page crashes.

Also addresses https://github.com/responsible-ai-collaborative/aiid/issues/2315

I don't think https://github.com/responsible-ai-collaborative/aiid/issues/2370 is happening anymore in the first place but if it is this will probably stop it.

### DIFF
--- a/site/gatsby-site/src/components/checklists/CheckListForm.js
+++ b/site/gatsby-site/src/components/checklists/CheckListForm.js
@@ -577,7 +577,7 @@ const searchRisks = async ({
       if (notDuplicate) {
         risksToAdd.push(newRisk);
       }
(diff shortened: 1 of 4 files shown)


cesarvarela (changes requested):
I'm not a fan of relying too much on fallback assignments (`||`) because it implies that an error is being masked or the way code is structured is not the best, but if it is too much work to make it nicer, let's keep it.

Please remove the `console.log`s, and check that no test needs to be added/updated.

lmcnulty:
Stale

PR #2547, opened 2024-01-09
Document new Github workflows in README

cesarvarela:
It is probably time for a complete restructuring of the readme, but in the meantime...

PR #2548, opened 2024-01-09
Add staging branch deploy workflow

cesarvarela:
## Instructions before merging into `master`

1. Disable continuous integration from Netlify:

image

2. Delete all environment variables.


.github/workflows/deploy.yml
```diff
@@ -3,8 +3,22 @@ on:
   workflow_call:
     inputs:
       environment:
+        description: 'The Github environment to load secrets from'
         type: string
         required: true
+      context:
```
kepae, line 9: suggestion: a more descriptive variable, since it's actually very netlify specific? e.g. `netlify-context` (or `netlify_context`, whichever is the proper style).

same for the `alias` input below.

.github/workflows/staging.yml
```diff
@@ -0,0 +1,34 @@
+name: Deploy Staging Branch
+
+on:
+  push:
+    branches:
+      - staging
+jobs:
```
kepae, line 7: nit: fix whitespace? 

```
on:
  push:
    branches:
      - staging

jobs:
  call-test-build:
    ...
```

kepae: otherwise... very satisfying workflow to read

.github/workflows/deploy.yml
```diff
@@ -84,7 +98,11 @@ jobs:
         working-directory: site/gatsby-site
         run: |
           set -e
-          OUTPUT=$(bash -c "netlify deploy --json --alias=pr-${{ github.event.pull_request.number }}" | tr '\n' ' ')
+          if [[ -z "${{ inputs.alias }}" ]]; then
```
kepae, line 101: This logic is worth inline comments of the expected outputs. If I read this correctly...

"If the netlify alias is not set, target the "production" environment of the site. Otherwise, deploy to the given alias."

.github/workflows/staging.yml
cesarvarela: Which whitespace are you referring to?

kepae: sorry. A space between the `on:` block scope and the `jobs:` block scope, therefore putting a space after l6.

currently:
image

suggested:
```yaml
on:
  push:
    branches:
      - staging

jobs:
  call-test-build:
    uses: ./.github/workflows/test-build.yml
    secrets: inherit
    with:
      sha: ${{ github.sha }}
      environment: staging
```

kepae: (re-activating review comment ^)

kepae:
@cesarvarela I'm happy to merge this with https://github.com/responsible-ai-collaborative/aiid/pull/2547 . Let me know when you're happy with it? We should triple-check that the rest of the team is comfortable with the new workflows.

kepae (approved):

PR #2564, opened 2024-01-15
Add production deploy workflow

### LINKED ISSUE #2542: Add production github workflows

cesarvarela:
A new `production` GitHub environment needs to be created before merging this, and CI on Netlify needs to be disabled. 

closes #2542 and #2587


.github/workflows/production.yml
```diff
@@ -0,0 +1,46 @@
...
+    - cron:  '0 6,18 * * *'
+
+jobs:
+
+  call-atlas:
+    uses: ./.github/workflows/realm.yml
+    secrets: inherit
+    with:
+      sha: ${{ github.sha }}
+      environment: staging
```
kepae, line 17:
```suggestion
      environment: production
```

I understand we need the environment on GitHub allocated first, but want to catch it here too.

kepae (commented):
Let's do this after we fix our build and roll out staging -> prod again.

kepae (approved):

PR #2565, opened 2024-01-16
Blog – Update language

kepae:


kepae:
@clari182 Would you please double-check the translations of the introductory paragraphs in Spanish in French? In Spanish, I changed the acronym to "MASI", but I'm unaware of the French equivalent. Is "CSAM" recognized?

kepae:
One problem with putting the figure in the front of the article is that the card renders of the blog post no longer preview the text – only "...".

I will look into how this can be overridden or fixed.

Front page:
image

Blog page:
image

kepae:
Going to merge this for the production push, but would still appreciate a quick glance when you get the chance. :-)

clari182:
Hey @kepae ! 
Regarding the CSAM acronym: 
Before publishing the french translations I actually researched the CSAM equivalent a bit and it doesn't really exist. In France it looks like there's not a direct acronym to define this. So that's why I left it like that. But if you want we can add the explanation like this `(Connu sous le nom de CSAM (Child Sexual Abuse Material) aux États-Unis)`

Regarding the `...` issue, I'm researching it!

PR #2571, opened 2024-01-18
Display deployed commit sha on footer

### LINKED ISSUE #2569: Display latest commit from build in site footer
In the spirit of maturing our releases, we can indicate what commit the site is built from in the footer.
In the future, if we tag our releases and maintain artifacts, we can use a different version labeling scheme – but let's use commits for now.

cesarvarela:
closes #2569


cesarvarela:
Merging this so we have something to test the staging deploy

PR #2572, opened 2024-01-19
If excerpt doesn't exist use body as preview removing markdown

clari182:
Resolves https://github.com/responsible-ai-collaborative/aiid/pull/2565#issuecomment-1893992366

If excerpt is available it's used to preview the description, otherwise it falls back first to the metaDescription and then to the markdown, removing html and markdown tags.
I also added some metaDescription for the latest blog post. Please check if it's ok.

Test URL: https://pr-42--clarayoudale-staging.netlify.app/

### DIFF
--- a/site/gatsby-site/blog/deepfakes-child-safety/index.mdx
+++ b/site/gatsby-site/blog/deepfakes-child-safety/index.mdx
@@ -1,7 +1,7 @@
 ---
 title: 'Deepfakes and Child Safety: A Survey and Analysis of 2023 Incidents and Responses'
 metaTitle: 'Deepfakes and Child Safety: A Survey and Analysis of 2023 Incidents and Responses'
-metaDescription: ""
+metaDescription: "Rising concerns and legal challenges regarding the creation and distribution of AI-generated child sexual abuse materials in 2023."
 date: '2024-01-09'
 image: './images/image_doj.jpg'
 author: 'Daniel Atherton'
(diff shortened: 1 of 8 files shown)


site/gatsby-site/blog/deepfakes-child-safety/index.mdx
```diff
@@ -1,7 +1,7 @@
 ---
 title: 'Deepfakes and Child Safety: A Survey and Analysis of 2023 Incidents and Responses'
 metaTitle: 'Deepfakes and Child Safety: A Survey and Analysis of 2023 Incidents and Responses'
-metaDescription: ""
+metaDescription: "Rising concerns and legal challenges regarding the creation and distribution of AI-generated child sexual abuse materials in 2023."
```
kepae, line 4: I'm concerned about using the meta description as a literal preview. It's intended for summarizing the page, not being the content exactly.

What do you think about adding and using instead a specific dedicated `previewText`? That might also simplify the logic: use `previewText` first before falling back to other content.

I understand this could be a pain with supporting both Prismic and MDX. We can try to simplify too if that's the case.

Lastly, a working preview would be really helpful here, but I see there's problems with the workflow to deploy. Let me know if you're blocked here.

clari182: `previewText` sounds good. Not a problem for prismic because we only use rich text there so the preview text is already markdown and html free. For mdx we can add the `previewText` and if doesn't exist falleback to the content, removing markdowns.

clari182:
@kepae ready for another review. 

I added a previewText on my test environment here: https://pr-42--clarayoudale-staging.netlify.app/

kepae (approved):
noice

kepae:
I'll merge this, and then we can test the preview with the image changes from https://github.com/responsible-ai-collaborative/aiid/pull/2565#issuecomment-1893992366 in a follow-up.

PR #2585, opened 2024-01-24
Add empty env workflow

cesarvarela:
The `empty` environment on GitHub needs to be set up before merging this.


cesarvarela:
Things have changed a lot since then

PR #2589, opened 2024-01-24
Use production context to make sure migrations and translations run first

### LINKED ISSUE #2588: Migrations and Translation process not being run
Because the "Build site for testing" job is run with deploy context as `deploy-preview`, we do not run either the migration or translation process and therefore the build fails if new incidents were created in between builds.

https://github.com/responsible-ai-collaborative/aiid/actions/runs/7632199565/job/20827625761#step:7:4385

image

cesarvarela:
Should fix current broken staging build

closes #2588

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -61,7 +61,7 @@ jobs:
         run: npm install netlify-cli -g
 
       - name: Build using Netlify
-        run: netlify build --context ${{ inputs.context }} --offline
+        run: netlify build --context ${{ inputs.netlify-context }} --offline
         env:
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         run: npm install netlify-cli -g
 
       - name: Build using Netlify
-        run: netlify build --context ${{ inputs.context }} --offline
+        run: netlify build --context ${{ inputs.netlify-context }} --offline
```
cesarvarela, line 64: Unrelated but this was wrong

kepae: if only we had type-checked configs ;-)

kepae (approved):

PR #2593, opened 2024-01-25
Run cron using new workflows

cesarvarela:
This runs the production build automatically every 12 hours and replaces the old cron.


cesarvarela:
closing as this is included in #2564

Issue #2583, opened 2024-01-23
Discover – making filters more prominent


kepae:
Per https://github.com/responsible-ai-collaborative/aiid/issues/1643#issue-1597175006:
> "I didn't know the discover app supports filters". I think doing something to make it slightly more prominent would be good. Maybe adding the filter icon would be enough.

kepae:
Closing in favor of the existing https://github.com/responsible-ai-collaborative/aiid/issues/1643.

PR #2602, opened 2024-01-30
Show single row of filter dropdowns

lmcnulty:
Addresses the filter visibility problem of #1643, #2583.

Current thoughts:

When space is available, put two filters on first row with options to expend

image

When there's not enough space for one row, show one extra row of filters, which expands to two.

image

When there's enough horizontal space, show in sidebar. I'll separate this into a different PR.

image

---

I've made the filter dropdown buttons white, because I expect blue buttons to mean "primary" and it's not obvious that filters should be considered the primary action. But I'm unsure on that point.


kepae:
On improving the current filter display:
This is great. A major improvement for me is, by default, leading with some filters available and showing "show more", rather than requiring the user to find the "filter search" drop down.

I'm neutral on the color changes of the filter buttons. The most important thing is that they show distinctly when they are selected -- even if being deeplinked from elsewhere.

On sidebar:
Definitely agree it's a great idea to do a **sidebar in a separate PR and issue**. I think that could be a bigger lift and I'm not sold on the idea yet, as we'll now have multiple columns and I'm not sure of placing filter/facet logic on the right-hand side. Plus, we just did a lot of work to make the classifications filter work like a tree, which works best as a drop-down. That said, this is part of maturing and growing the UX here, and we might like it better. But let's not prioritize this just yet.

@cesarvarela what do you think? You're most familiar with this UI and might also have some prior thoughts on it.

lmcnulty:
Current implemented behavior:

When space is available, first two filters appear inline with the other controls.

image

When expanded, it shows two rows below

image

When there's not enough room to show the filters inline, there's one row when not expanded

image

kepae:
@lmcnulty is this good for review? If so, would you please assign some reviewers?
I think you still might have to split this into two PRs. Feel free to open up separate PRs/branches if that's easier.

clari182:
This looks good! The only thing I noticed is the dropdown size changes on open/close:
image
image

Also @kepae is there a way to know on the Algolia's analytics which filter is most used? I think that would be a good opportunity to put the 2 most used first.

clari182 (changes requested):
This looks good! The only thing I noticed is the dropdown size changes on open/close:
image
image

Also @kepae is there a way to know on the Algolia's analytics which filter is most used? I think that would be a good opportunity to put the 2 most used first.

lmcnulty:
@clari182 Fixed!

kepae:
> Also @kepae is there a way to know on the Algolia's analytics which filter is most used? I think that would be a good opportunity to put the 2 most used first.

@clari182 As far as I know, no, there are no metrics from Algolia on which queries/searches are more popular. We would have to record that ourselves, which likely isn't a priority right now. But hopefully this UX change could drive more use of the filters overall!

site/gatsby-site/src/components/discover/REFINEMENT_LISTS.js
```diff
@@ -128,4 +128,14 @@ const REFINEMENT_LISTS = [
...
-export default REFINEMENT_LISTS;
+const FIRST_ROW = REFINEMENT_LISTS.filter((refinement) =>
+  [
+    'classifications',
+    'epoch_incident_date',
+    'epoch_date_published',
+    'source_domain',
+    'language',
+  ].includes(refinement.attribute)
+);
```
clari182, line 139: I think we should slice the first 5 instead, in case we end up updating the refinements or change their order

clari182 (changes requested):
One minor change and can you add a test for this new functionality? Thanks!

site/gatsby-site/src/components/discover/REFINEMENT_LISTS.js
lmcnulty: Good point. I also changed the order so that the first row would show the ones I would guess are most interesting, in the absence of data.

image

I also remembered that I changed the wording on the expand/collapse filters button, so those need translations. I went with "More filters" = "Ver más filtros" = "Plus de filtres" // "Fewer filters" = "Ver menos filtros" = "Moins de filtres". Let me know if those are incorrect or not idiomatic.

Test coming up...

site/gatsby-site/src/components/discover/Filters.js
```diff
@@ -1,12 +1,10 @@
 import React from 'react';
+
```
clari182, line 2: Remove new empty line

clari182 (changes requested):
Looks good, just one tiny thing, there's an extra empty line in Filters.js

After that fix we're good to merge!

clari182 (approved):

PR #2603, opened 2024-01-30
Update actions/cache to v4

kepae:
Eliminates a Node 16 deprecation warning annotation in our workflows:

> `call-test-build / Build site for testing`
> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/cache/restore@v3, actions/cache/save@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

actions/cache v4: https://github.com/actions/cache?tab=readme-ov-file#v4
> * Updated to node 20
> * Added a save-always flag to save the cache even if a prior step fails

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -37,7 +37,7 @@ jobs:
 
       - name: Read node modules from cache
         id: cache-nodemodules
-        uses: actions/cache/restore@v3
+        uses: actions/cache/restore@v4
         env:
           cache-name: cache-install-folder
         with:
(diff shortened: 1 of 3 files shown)


pdcp1 (approved):
LGTM. I still see some warnings on this PR's Checks https://github.com/responsible-ai-collaborative/aiid/actions/runs/7716111743?pr=2603
But we can merge this a see what else is missing

kepae:
regarding this PR's warnings – that's because the repository isn't using the workflow files defined in this PR, but rather those that are already checked into the `master` branch. When we merge this to `staging` and then `master`, we should be good.

PR #2613, opened 2024-02-02
Customizable runners on production and staging workflows

cesarvarela:
Defines two new variables that let us set specific runner labels for production and staging deployments:

- PRODUCTION_RUNNER_LABEL 
- STAGING_RUNNER_LABEL

By default, `ubuntu-latest` is used, but it can be set to a label (the runner name)

Setting an 8 core box per job shaves some build and deploy minutes but seems to have no effect when running  tests:

Standard 4 core runners (~42 minutes):

image

8 cores (~28 minutes):

image

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -21,13 +21,12 @@ on:
       runner-label:
         description: The label of the runner to use
         type: string
-        default: ubuntu-latest
 
 jobs:
   netlify-deploy:
     environment: ${{ inputs.environment }}
     runs-on:
-      labels: ${{ inputs.runner-label }}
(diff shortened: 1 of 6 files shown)


kepae:
Cool, this is great to have. Why have separate `PRODUCTION_` and `STAGING_` vars and not use environment-level vars, i.e. just `RUNNER_LABEL`?

kepae:
Ah, right, because the call.happens at the repo level, not before the GitHub env is selected and.passed down.

kepae (approved):

cesarvarela:
> Ah, right, because the call.happens at the repo level, not before the GitHub env is selected and.passed down.

Exactly!

Another finding (at least for me) is that you can't have empty variables :/

PR #2618, opened 2024-02-06
Display largest count on classifications filter

### LINKED ISSUE #2577: Discover – Classifications filter facets counting field values, not search results
The field facet currently counts the number of unique _values_ in the given property, rather than the number of search results (incidents and/or reports) attributed with each value.

For example: CSETv1 > Physical Objects should be 28 rather than 2 in the below example (if I count correctly).

image

See https://github.com/responsible-ai-collaborative/aiid/issues/2330#issuecomment-1749498903 for an example mock with proper facet counts.

@cesarvarela -- could be a good opportunity to share Algolia responsibilities with the team so it's not just you!

cesarvarela:
closes #2577

### DIFF
--- a/site/gatsby-site/src/components/discover/filterTypes/Classifications.js
+++ b/site/gatsby-site/src/components/discover/filterTypes/Classifications.js
@@ -29,7 +29,8 @@ function Attribute({ name, refinement, searchResults }) {
         tabIndex={0}
         onClick={() => setCollapsed((c) => !c)}
       >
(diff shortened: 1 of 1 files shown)


kepae:
Wanted to investigate this more, @cesarvarela ? Or should we ship this more accurate-but-always-approximate fix? :-)

cesarvarela:
@kepae let's ship this as it is

PR #2619, opened 2024-02-07
On demand workflow triggers, allow build caches to be skipped, allow force deploy, allow custom runners from env

### LINKED ISSUE #2617: On demand re-deploys to production and staging without hitting cached builds
We need a way to use our new continuous deployment workflows to push the exact same code but pulling in new data from Mongo as well as Prismic. This is difficult when re-running an existing workflow because it will look to use a cached build from a prior run, requiring that the caches be manually deleted before the re-run.

We should add the ability to initiate a _new_ workflow run/deployment via a `workflow_dispatch` configuration, as well as adding an explicit parameter to skip the build cache that the user can configure when initiating the workflow:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#providing-inputs

cesarvarela:
closes #2617, #2631

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -5,13 +5,30 @@ on:
     branches:
       - master
   schedule:
-    - cron:  '0 6,18 * * *'
-
+    - cron: "0 6,18 * * *"
+  workflow_dispatch:
+    inputs:
+      skip-cache:
+        description: "Skip Cache"
+        required: true
+        default: true
+        type: boolean
+      force-deploy:
+        description: "Deploy even if tests fail"
+        required: true
+        default: false
+        type: boolean
 jobs:
+
+  cache-modifier:
+    uses: ./.github/workflows/cache-modifier.yml
+    with:
+      skip-cache: ${{ github.event.inputs.skip-cache }}
 
   call-atlas:
     uses: ./.github/workflows/realm.yml
     secrets: inherit
+    needs: cache-modifier
     with:
       sha: ${{ github.sha }}
       environment: production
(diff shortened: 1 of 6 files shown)


kepae:
Per our current workflow issues, I think we should also add a `force_push` input to our workflows that executes the deploy step in staging/prod workflows even if the test-build step fails. This sounds ugly, but I think we need it anyway in case we have more pressing problems like this in the future, e.g. broken production. I believe we can accomplish this with the conditional on a new input: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-not-requiring-successful-dependent-jobs. 

 I'm thinking we should add this to your PR here, and target master branch with a hotfix since it's "just" a workflow change. I would prefer not to release the other staging changes until we get this addressed.

.github/workflows/production.yml
```diff
@@ -6,11 +6,25 @@ on:
       - master
   schedule:
     - cron: "0 6,18 * * *"
+  workflow_dispatch:
+    inputs:
+      skip-cache:
+        description: "Skip Cache"
+        required: true
+        default: true
+        type: boolean
```
pdcp1, line 15: We should add the same input as `staging.yml`

```
force-deploy:
  description: "Deploy even if tests fail"
  required: true
  default: false
  type: boolean
```

pdcp1 (changes requested):

pdcp1 (approved):

kepae (approved):

kepae:
@cesarvarela please updating to point at `master` and merge when you are ready! We'll have to force push after anyways.

kepae (approved):

PR #2630, opened 2024-02-09
Skip flaky tests on `pages.cy.js`

pdcp1:
In order to unblock the Production build we should skip the flaky tests on `pages.cy.js`


kepae:
Thanks Pablo. In the interest of time, I'll merge this to staging and we'll let that run complete.

Issue #2640, opened 2024-02-14
File path parsing logic for `.mdx` files fails if working path contains additional period (`.`) characters


kepae:
When trying to build the site, file parsing logic for `.mdx` files will fail if the entire path contains more than one period (`.`) character that doesn't conform to the filename + language code expectation.
The logic seems to fail here:
https://github.com/responsible-ai-collaborative/aiid/blob/4f925d9877319ffc67535b626347ac20ed862e67/site/gatsby-site/plugins/gatsby-theme-i18n/gatsby-node.js#L227

This splitting logic seems to handle the optional language suffixes for files, e.g. `post.mdx` and `post.es.mdx`.

My machine user name, `k.p`, contains a `.`, so this build fails. With some print debugging:

```js
//split the filename in three parts split by the dot.
let [thePath /* lang */, , ext] = mdxFile.split(`.`);
console.log({mdxFile, thePath, ext})
```
```
{
  mdxFile: '/Users/k.p/github/kepae/aiid/site/gatsby-site/blog/improv-ai/index.mdx',
  thePath: '/Users/k',
  ext: 'mdx'
}

 ERROR #11321  API.NODE.EXECUTION

"gatsby-plugin-mdx" threw an error while running the onCreatePage lifecycle:

Could not locate File node for /Users/k.mdx

  324 |   const fileNode = getNodesByType(`File`).find(node => node.absolutePath === mdxPath);
  325 |   if (!fileNode) {
> 326 |     throw new Error(`Could not locate File node for ${mdxPath}`);
      |           ^
  327 |   }
  328 |
  329 |   // Avoid loops

File: node_modules/gatsby-plugin-mdx/dist/gatsby-node.js:326:11
```

PR #2645, opened 2024-02-16
Fix mdx file path parsing of dots

kepae:
Addresses #2640.


site/gatsby-site/plugins/gatsby-theme-i18n/gatsby-node.js
```diff
@@ -220,27 +220,45 @@ exports.onCreatePage = ({ page, actions }, themeOptions) => {
...
-        //if there's data in the third part, just keep the first and last part, removing the language
-        theFilePath = `${thePath}.${ext}`;
-      } else {
-        //if there's no content in the third part, it means that there's no language part. No need to remove the language
+      let fileName = path.basename(mdxFile);
+
+      let ext = path.extname(mdxFile);
+
+      if (ext !== '.mdx') {
+        console.error(`Unexpected file format in mdx path parsing: ${mdxFile}`);
```
cesarvarela, line 233: Should we crash the build instead? Logs like this probably get ignored.

(or should be logged as warnings)

site/gatsby-site/plugins/gatsby-theme-i18n/gatsby-node.js
```diff
@@ -220,27 +220,45 @@ exports.onCreatePage = ({ page, actions }, themeOptions) => {
...
+
+      if (fileNamePieces.length == 2) {
+        // ex: index.mdx
         theFilePath = mdxFile;
+      } else if (fileNamePieces.length == 3) {
+        // ex: index.es.mdx
+        // Keep everything except the language code.
+        theFilePath = `${path.join(thePath, fileNamePieces.at(0))}${ext}`;
+      } else {
+        console.error(`Unexpected file format in mdx path parsing: ${mdxFile}`);
```
cesarvarela, line 247: ditto

site/gatsby-site/plugins/gatsby-theme-i18n/gatsby-node.js
kepae: Hard agree. What's the best way to crash here?

kepae: Chose `throw` since there's not a specific gatsby hook (like I was expecting).

Issue #2590, opened 2024-01-24
Update common entities


jphall663:
I suggest we update the common entities feature on the incidentdatabase.ai landing page. It would be ideal for the common entities to reflect systems instead of/in addition to companies/organizations. 

I understand this could require improved query mechanisms, and we may consider minimizing the feature until more incident response occurs or query approaches can be updated to reflect systems.

kepae:
I agree. The entities we currently show rarely change, which doesn't give a sense of the times (e.g. LLM chatbot chaos).

There's a few things we can do.

* In the short term, minimize the common entities feature. Move it down the page, or remove entirely?
    * Or, calculate the most common entities in the recent N months (3? 6?) to provide a more recent view. We can calculate what this would be and see.
* In the longer term, support indexing AI systems as entities in our entity collection. We currently only attempt to name developers, deployers, and victims – that is, organizations and people – and we don't characterize them beyond name. Given the challenges inherent to _stably naming and separating AI systems (or components)_ from one another, we would probably also prioritize:
    * #2465 
    * #2536 
    * As well as -- generalizing our entity relationships so that "links" between data on the platform are not restricted to particular documents (e.g. reports only point to incidents, but can also be linked to systems and other entities in a semantic way).

@jphall663 @datherton09 @Janetbananet -- this change and another one suggested (#2563) have editorial and voice impacts on the landing page. I would suggest we discuss these in the next product meeting.

smcgregor:
Idea to throw in the mix: list the most recent incident responses.

jphall663:
In the short term, I am in favor of:

> Move it down the page, or remove entirely

On #2563, my feeling is I like the current setup. It draws me in sometimes, however, it also has click-baity vibes. I am concerned that referencing just the incident will look boring. I wonder if we could keep an image in the display, as it is now, whether we link to incident or news report.

Janet-ResponsibleAI:
If we move "Common Entities" down the page, then it bumps up the next item in the horizontal view which is the "Quick Add New Report" link. Not to say there is anything inherently wrong with that, but it points to a broader question about the contents and order of contents on the landing page.

I'm going to create a mock-up of some changes to the landing page and open a new GitHub issue to generate discussion regarding the overall content and order, with an eye towards what's above the fold.

[...]

PR #2653, opened 2024-02-20
Remove "Common Entities" component

pdcp1:
This PR:
- removes the Home page Common Entities component

Reference https://github.com/responsible-ai-collaborative/aiid/issues/2590


kepae (commented):
drive-by comment – I don't think we should delete the component file and assets. The replacement component might use it as a base, and it can be deleted later/after.

kepae (approved):
Will merge on test pass :-)

pdcp1:
@kepae Tests passed 🚀

PR #2662, opened 2024-02-23
Add lookupByUrl  REST endpoint

cesarvarela:
example request:

https://pr-2662--staging-aiid.netlify.app/api/lookupbyurl?urls=https%3A%2F%2Fwww.goodmorningamerica.com%2Ffamily%2Fstory%2Fparents-inappropriate-content-youtube-54993637%2Chttps%3A%2F%2Fnoresults.com

swagger ui:
https://pr-2662--staging-aiid.netlify.app/api-spec/


cesarvarela:
closing in favor of #2662

site/gatsby-site/src/api/lookupbyurl.js
```diff
@@ -0,0 +1,109 @@
+import Rollbar from 'rollbar';
+import siteConfig from '../../config';
+
+import Cors from 'cors';
+
+const cors = Cors();
+
+const isValidURL = (string) => {
+  try {
+    new URL(string);
```
kepae, line 10: For a follow-up PR – I think we should consider matching on URLs regardless of whether they include protocol (only strings with `http(s)` are "valid" URLs) and regardless of including `www`. Yes, `www` is technically a sub-domain, but I think the user would expect that URLs are found all the same.

```js
a = new URL("https://nytimes.com") 
b = new URL("https://www.nytimes.com") 
console.log(a.host, b.host)
```
```
> nytimes.com www.nytimes.com
```

site/gatsby-site/src/api/lookupbyurl.js
```diff
@@ -0,0 +1,109 @@
...
+});
+
+export default async function (req, res) {
+  try {
+    await handler(req, res);
+  } catch (error) {
+    console.error(error);
+    rollbar.error(error);
+
+    res.status(500).send('An error occurred');
```
kepae, line 107: Any other error we should propagate? I guess it doesn't make sense to return 404s since we handle multiple URLs here. Trying to think of best practice for `not found` in an array.

kepae (commented):
I can get an error back if I add an empty url ( `,,` -> `%2C%2C`):
https://pr-2662--staging-aiid.netlify.app/api/lookupbyurl?urls=https%3A%2F%2Fwww.goodmorningamerica.com%2Ffamily%2Fstory%2Fparents-inappropriate-content-youtube-54993637%2C%2Chttps%3A%2F%2Fnoresults.com

Should we error on empty URLs?

kepae:
Ping? Let's include this in our next production push. 💪

site/gatsby-site/static/spec.json
```diff
@@ -0,0 +1,94 @@
...
+    "openapi": "3.0.0",
+    "info": {
+        "title": "REST API",
+        "version": "1.0.0"
+    },
+    "paths": {
+        "/api/lookupbyurl": {
+            "get": {
+                "summary": "Lookup reports and incidents by URL",
+                "parameters": [
```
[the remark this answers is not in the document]
cesarvarela: Yes, you are correct.

site/gatsby-site/src/api/lookupbyurl.js
cesarvarela: Maybe if none of them, but it is weird. It is also missing `400`, but I didn't want to make it too complicated initially.

cesarvarela:
I've updated the api to use `explode:false`, which after integrating the validator I've noticed is a better way to handle the `urls` parameter in array format.

site/gatsby-site/src/api/lookupbyurl.js
cesarvarela: Add a validator which covers most errors

kepae (approved):

PR #2562, opened 2024-01-12
Support JA (Japanese 🇯🇵 ) language

pdcp1:
This PR adds support for the Japanese language.

More information about adding a new language in our [wiki page](https://github.com/responsible-ai-collaborative/aiid/wiki/Adding-a-new-language-support)

## Important notes before merge it
- `ja` available language is temporary harcoded in `site/gatsby-site/config.js`

## Testing preview link 🇯🇵 
https://pr-2562--staging-aiid.netlify.app/ja/

PR #2670, opened 2024-02-28
Limit the number of reports to translate by filtering by submission date

pdcp1:
From this https://github.com/responsible-ai-collaborative/aiid/pull/2562#issuecomment-1969214657

This PR:

- Adds `TRANSLATE_SUBMISSION_DATE_START` optional env variable. Variable format `yyyy-mm-dd`
E.g.: `TRANSLATE_SUBMISSION_DATE_START=2024-02-15`
This variable sets the date from which the reports will be translated

### DIFF
--- a/site/gatsby-site/src/utils/Translator.js
+++ b/site/gatsby-site/src/utils/Translator.js
@@ -130,6 +130,14 @@ class Translator {
     let reportsQuery = {};
 
     if (this.submissionDateStart) {
+      // Check if the date is valid
+      if (isNaN(Date.parse(this.submissionDateStart))) {
+        const errorMessage = `Translation process error: Invalid date format for TRANSLATE_SUBMISSION_DATE_START env variable: [${this.submissionDateStart}]`;
+
+        this.reporter.error(errorMessage);
+        throw errorMessage;
+      }
+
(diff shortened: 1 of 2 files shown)


site/gatsby-site/src/utils/Translator.js
```diff
@@ -125,7 +127,24 @@ class Translator {
...
     await this.mongoClient.connect();
 
-    const reports = await this.mongoClient.db('aiidprod').collection(`reports`).find({}).toArray();
+    let reportsQuery = {};
+
+    if (this.submissionDateStart) {
+      this.reporter.log(
+        `Translating incident reports submitted after [${this.submissionDateStart}]`
+      );
+      reportsQuery = { date_submitted: { $gte: new Date(this.submissionDateStart) } };
```
kepae, line 136: We should probably crash/throw + return error if we don't parse the date correctly (not sure how `$gte` will like null :-) )

pdcp1: The `if (this.submissionDateStart)` statement avoids `null` values but I added a check for another type of invalid string value like `blah blah` just in case. Thanks for your input!

PR #2683, opened 2024-03-07
Only create an 'incident' notification if whitelistead field updated

### LINKED ISSUE #2676: Subscriptions – Notifications sent for "on update" triggers when any field in document changes
The recent migration to add `created_at` metadata to incident documents resulted in too many notifications being sent.

The proposed solution is to make explicit the fields we want to result in a trigger (a.k.a. a "whitelist"), rather than implicitly consider any chance to the raw document an update worth notifying for.

clari182:
- closes #2676 

Whitelisted fields: `title`, `description`, `date`, `Alleged deployer of AI system`, `Alleged developer of AI system`, `Alleged harmed or nearly harmed parties`, `reports`, `editors` and `editor_notes`

How to test:

- Fork this branch to your repo and update realm
- Manually update an incident from the DB - **Note:** the incident must have an `incident` type subscription
- Check the Realm logs for function `onIncidentUpdate`
- If you've updated a field that is **not** whitelisted you should see the trigger executed in the logs
- If you've update a field that is whitelisted you should **not** see the trigger executed in the logs

### DIFF
--- a/site/realm/functions/onIncidentUpdate.js
+++ b/site/realm/functions/onIncidentUpdate.js
@@ -15,7 +15,7 @@ exports = async function (changeEvent) {
   }
 
   const updatedFields = Object.keys(updateDescription.updatedFields);
-  
+
   const incidentId = fullDocument.incident_id;
 
   console.log(`Processing updates on incident ${incidentId}`);
--- a/site/realm/functions/onIncidentUpdate.js
+++ b/site/realm/functions/onIncidentUpdate.js
@@ -14,8 +14,11 @@ exports = async function (changeEvent) {
     return;
   }
 
+  const notificationWhitelistedFields = ['title', 'description', 'date', 'Alleged deployer of AI system', 'Alleged developer of AI system', 'Alleged harmed or nearly harmed parties', 'reports', 'editors', 'editor_notes'];
+
   const updatedFields = Object.keys(updateDescription.updatedFields);
 
+  const willNotifyIncidentUpdate = updatedFields.some(field => notificationWhitelistedFields.includes(field));
   const incidentId = fullDocument.incident_id;
 
   console.log(`Processing updates on incident ${incidentId}`);
@@ -25,8 +28,6 @@ exports = async function (changeEvent) {
   const subscriptionsCollection = context.services.get('mongodb-atlas').db('customData').collection("subscriptions");
   const reportsCollection = context.services.get('mongodb-atlas').db('aiidprod').collection("reports");
   const subscriptionsToIncident = await subscriptionsCollection.find({ type: 'incident', incident_id: incidentId }).toArray();
-
-  console.log(`There are ${subscriptionsToIncident.length} subscribers to Incident: ${incidentId}`);
 
   // Process subscriptions to Incident updates
   try {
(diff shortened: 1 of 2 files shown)


site/realm/functions/onIncidentUpdate.js
```diff
@@ -14,8 +14,11 @@ exports = async function (changeEvent) {
     return;
   }
 
+  const notificationWhitelistedFields = ['title', 'description', 'date', 'Alleged deployer of AI system', 'Alleged developer of AI system', 'Alleged harmed or nearly harmed parties', 'reports', 'editors', 'editor_notes'];
+
   const updatedFields = Object.keys(updateDescription.updatedFields);
 
+  const willNotifyIncidentUpdate = updatedFields.some(field => notificationWhitelistedFields.includes(field));
```
kepae, line 21: Why not just return here if no fields are worth notifying? (instead of letting the db queries and complicated logic below start running)
I note that we return early in the above code for "invalid" change events, so this seems fair and safe, but I could be missing a reason.

(drive-by comment 🚗 )

site/realm/functions/onIncidentUpdate.js
```diff
@@ -14,10 +14,18 @@ exports = async function (changeEvent) {
     return;
   }
 
+  const notificationWhitelistedFields = [`title`, `description`, `date`, `Alleged deployer of AI system`, `Alleged developer of AI system`, `Alleged harmed or nearly harmed parties`, `reports`, `editors`, `editor_notes`];
```
kepae, line 17: I took more time to look at the Realm configurations, and we actually already have a way of triggering in Realm based on the updated fields. We ignore the tSNE and embedding updates, for example:
https://github.com/responsible-ai-collaborative/aiid/blob/91ab98aba974da67b2aedf921e430d4f9d2caf73/site/realm/triggers/onIncidentUpdate.json#L11

Overall, though, I like the logic living in a function we can test, rather than some configuration file we can't test as easily.

What do you think?
Also curious for @cesarvarela.

kepae (changes requested):
Can we test this by adding a unit test here in `site/gatsby-site/cypress/e2e/unit/functions/onIncidentUpdate.cy.js` ?

clari182:
I think it's actually better to use the config, and it's also editable in the code so we should be fine. If I'm correct the logic would be 

```
"match": {
            "$or": [
                {
                    "updateDescription.updatedFields.title": {
                        "$exists": true
                    }
                },
                {
                    "updateDescription.updatedFields.description": {
                        "$exists": true
                    }
                },
                {
                    "updateDescription.updatedFields.date": {
                        "$exists": true
                    }
                },
                {
                    "updateDescription.updatedFields.\"Alleged deployer of AI system\"": {
                        "$exists": true
                    }
                },
                {
                    "updateDescription.updatedFields.\"Alleged developer of AI system\"": {
                        "$exists": true
                    }
                },
                {
                    "updateDescription.updatedFields.\"Alleged harmed or nearly harmed parties\"": {
                        "$exists": true
                    }
                },
                {
                    "updateDescription.updatedFields.reports": {
                        "$exists": true
                    }
                },
                {
                    "updateDescription.updatedFields.editors": {
                        "$exists": true
                    }
                },
                {
                    "updateDescription.updatedFields.editor_notes": {
                        "$exists": true
                    }
                }
            ]
        }
```

I'll be testing it and see if it works ok with the fields that have spaces in them

site/realm/functions/onIncidentUpdate.js
cesarvarela: Sadly, we have to use configurations for now because no matter how careful you are with the function, the trigger still crashes if there are too many updates, like in the case of migrations. Filtering through configs avoids function execution altogether.

kepae (approved):
LGTM, I'll merge this shortly, and this should be easy to test in staging, too.

PR #2691, opened 2024-03-13
Feature/reports table view

### LINKED ISSUE #2681: Issue Reports – Dedicated page for viewing table or list of issue reports
Editors could use a more lightweight way of digging through the history of issue reports than the [discover interface's filtering method](https://incidentdatabase.ai/apps/discover/?display=list&is_incident_report=false), like the ["all incidents in list form"](https://incidentdatabase.ai/summaries/incidents/) summary a.k.a. "power user mode." The main impetus is so that when/if the database begins to accept more varied types of reports, there is a way to easily review the existing reports and connect them to incidents.

Specifically, we should list **all** issue reports in some manner along with dates of submission and publication, especially while the collection size is manageable for a single page. As of now, there are 223 such reports.
EDIT: Additionally, all reports should be visible on _one page_.

For now, the only technical handle on what constitutes an issue report is whether the field on the report document `is_incident_report` is `false`. This might change if we open up to more types of reports:
- #2292

Options:
- Create a static page (perhaps `/summaries/issues/`) that lists reports by their title, also displaying their date submitted and date published.
- Alternatively, we could create a table view (like we do for incidents as well: https://incidentdatabase.ai/apps/incidents/) that displays all this info. This might be worth it for easier sorting implementation.

@datherton09 @Janet-ResponsibleAI what do you think? Which interface would be more helpful?

clari182:
- closes #2681 

Test URL: https://pr-2691--staging-aiid.netlify.app/
Changes: 
- Added `Issue Reports` and `Reports` in the table view page
- Removed `/apps/reports` page, now everything is in the table view page `apps/incidents`


clari182:
@kepae @datherton09 
this is ready for review, let me know if the new implementation seems alright to you.

datherton09:
@clari182 Clara, this is wonderful! Thank you so much! 

One quick thing I was wondering about with sorting the issue reports by date is that I'm not entirely sure if it's actually sorting them in either ascending/descending order? (The ordering seems semi-haphazard? Or maybe I'm missing something!)

clari182:
@datherton09 date filter and sort fixed! Ready for another round of testing

kepae:
Sorry, didn't finish my review here before merge conflicts came up! Will review shortly.

kepae (approved):

Issue #2695, opened 2024-03-15
Last "Next Incident" button redirects to a null incident


pdcp1:
As part of the testing of https://github.com/responsible-ai-collaborative/aiid/pull/2689 I found an edge-case test.

## Steps to reproduce

1. Go to the last incident (at this moment the last incident is https://incidentdatabase.ai/cite/652 )
2. Scroll down and check the `Next Incident` button. It should be disabled
3. Click that `Next Incident` button anyways
4. It redirects to https://incidentdatabase.ai/cite/null/

The same happens with the first incident https://incidentdatabase.ai/cite/1/ and the `Previous Incident` button.

## Notes
In addition to the disabled visual state, the button shouldn't be clickable (completely disabled)

kepae:
I believe this is now fixed.

PR #2696, opened 2024-03-15
Disable Previous and Next incident buttons on first and last incidents

pdcp1:
This closes https://github.com/responsible-ai-collaborative/aiid/issues/2695


pdcp1:
@clari182 Nice catch! I just opened a new issue to add these buttons on the duplicated Incident's page https://github.com/responsible-ai-collaborative/aiid/issues/2698

PR #2700, opened 2024-03-19
Forwared graphql requests to Netlify function

cesarvarela:


site/gatsby-site/src/graphql/schema.js
```diff
@@ -0,0 +1,4282 @@
+export const typeDefs = `
```
kepae, line 1: This file looks like it's going to be fun to maintain, but hopefully easier (with better errors for development) than the current setup in the long run if we can break it down and compose the pieces.

kepae (approved):
Ship it when you're good to go! Let's try it.

PR #2710, opened 2024-03-27
Include incident title for latest reports on landing page, add separate pageCreator

### LINKED ISSUE #2563: Suggestion: Change "Latest Incident Report" to Incident ID rather than individual news report?
What would you all think of changing the "Latest Incident Report" that appears on the main page from the latest news report within an incident ID to the incident ID itself? This might be one of the ways to perform some slight distancing from journalistic bias issues. For example, the latest report (whose title is "Microsoft says its AI is safe. So why does it keep slashing people’s throats?") copies the sensationalistic tone of the reporting to our own work, placing it front and center for visitors to our site. Small adjustments that place our own titling at the forefront may be a way to mitigate this.

lmcnulty:
Resolves #2563.

image


site/gatsby-site/gatsby-config.js
```diff
@@ -305,6 +305,6 @@ module.exports = {
   plugins: plugins,
   trailingSlash: `always`,
   flags: {
-    DEV_SSR: true,
+    DEV_SSR: false,
```
lmcnulty, line 308: Don't include this.

site/gatsby-site/page-creators/createLandingPage.js
```diff
@@ -0,0 +1,85 @@
+const path = require('path');
+
+const createLandingPage = async (graphql, createPage) => {
+  const result = await graphql(`
+    query WordCounts {
```
kepae, line 5: please rename the query?

kepae (changes requested):
only one comment!

site/gatsby-site/page-creators/createLandingPage.js
lmcnulty: Good catch, thanks.

kepae (approved):

PR #2727, opened 2024-04-10
Deploy to Production

cesarvarela:
Including:
* #2726 
* #2618 
* #2696 
* #2722 
* #2734
* #2729 
* #2737 
* #2738


cesarvarela:
@kepae checked again. This is going to result in translating every report into Japanese 🤔 

At 20 USD per million characters in translation (might be more when translating to Japanese), 4000 characters on average per report, and 3600 reports:

`(20*(4000*3600))/1000000= ~288 USD `

kepae:
@cesarvarela thanks for the check. But even with the changes from https://github.com/responsible-ai-collaborative/aiid/pull/2670? `TRANSLATE_SUBMISSION_DATE_START` is set to 2024-03-01.

PR #2752, opened 2024-04-25
Add checklists to sidebar

lmcnulty:
https://github.com/responsible-ai-collaborative/aiid/issues/2746

The sidebar contents is actually controlled from Prismic now, so the change would need to be made there also.


clari182 (changes requested):
Looks good! The only thing is missing is "Risk Checklists" translations. I've added the sidebar item to Prismic prod, once deployed we need to publish it.

lmcnulty:
I've added French and Spanish translations – I'm not sure what to do for the Japanese, though.

kepae:
Let's not block on Japanese. We can add tests later for tracking which languages have which strings completed for translations across files. Maybe we can actually add this as a leaderboard problem for contributors? :-) 

But in the meantime, let's ship risk checklists in staging and get ready for making in public!

clari182 (approved):

PR #2756, opened 2024-04-25
Fix table view report links

### LINKED ISSUE #2229: Improve report pages: Show page for each `/report/*`, regardless of incident associations

clari182:
- closes #2229 

The original idea was to have the report page for _all_ reports. But this is not possible due to the quantity of pages generated, the build fails because of insufficient space.

So, the fix here instead is when viewing reports on the table view (https://pr-2756--staging-aiid.netlify.app/apps/incidents/?view=reports) the report links to the cite page of that report.


pdcp1:
@clari182 During my testing of https://pr-2756--staging-aiid.netlify.app/apps/incidents/?view=reports I realized that we have some timeout errors `502 502 Bad Gateway` on each `/graphql` call to retrieve the Live version of `reports` and Live version of `incidents`.

```
{"errorType":"LambdaTimeout","errorMessage":"2024-05-22T19:10:14.718Z afdded9a-fa1b-47d9-b13c-03372618773c Task timed out after 10.01 seconds"}
```

@cesarvarela Is this something related to the API changes? Or is this only happeninginn the Staging environment?

site/gatsby-site/src/components/reports/ReportCard.js
```diff
@@ -143,23 +138,19 @@ const ReportCard = ({
...
           <div className="flex">
-            {reportTitle ? (
-              <>{reportTitle}</>
-            ) : (
-              <button
-                className="w-3/4 text-left"
-                onClick={toggleReadMore}
-                onKeyDown={toggleReadMoreKeyDown}
-              >
+            <Link to={`/reports/${item.report_number}`} className="hover:no-underline">
```
pdcp1, line 141: @clari182 Is this URL correct? When I navigate to https://pr-2756--staging-aiid.netlify.app/cite/658#r3803 and click on the report title, it redirects me to https://pr-2756--staging-aiid.netlify.app/reports/3803/ and it doesn't exist.

image

pdcp1 (changes requested):

pdcp1 (approved):

Issue #1577, opened 2023-01-31
Random Incident Button


lmcnulty:
I think we could use a button or link similar to the "Random Article" link on Wikipedia, especially since our scope is small enough that many users will potentially be interested in any incident in the database.

PR #2767, opened 2024-05-02
Add random incident option

lmcnulty:
Resolves https://github.com/responsible-ai-collaborative/aiid/issues/1577.

Adds a url, `/random/` which redirects to the citation page for a randomly-selected incident. I imagine this being linked in the sidebar – it needs to be set in Prismic.


site/gatsby-site/i18n/locales/ja/translation.json
```diff
@@ -528,5 +528,7 @@
   "Fewer filters": "フィルタを減らす",
   "Tags": "タグ",
   "Language": "言語",
-  "Read the Source": "情報源を読む"
+  "Read the Source": "情報源を読む",
+  "Redirecting...": "Redirecting...",
+  "Random Incident": "Random Incident"
```
clari182, line 533: Please add Japanese translations

clari182 (changes requested):
Looks good! Just 2 japanese translations missing, and can you add some tests for this?

kepae:
Does this PR build? Errors on the test build pipeline: https://github.com/responsible-ai-collaborative/aiid/actions/runs/8929044728/job/24525956231?pr=2767

site/gatsby-site/i18n/locales/ja/translation.json
lmcnulty: Attempted. For "redirecting..." I went with "リダイレクト中..." which got from Google translate and [seems to be used for this exact purpose](https://www.google.com/search?q=%22%E3%83%AA%E3%83%80%E3%82%A4%E3%83%AC%E3%82%AF%E3%83%88%E4%B8%AD...%22) in a lot of sites. Then for the random button I went with "おまかせ表示" which is taken verbatim from the random page link of Japanese Wikipedia.

clari182 (commented):
This looks good!! Works as expected.
I've added the new item to the sidebar on staging and prod, when this PR is merged it will be visible on the sidebar in staging (and then prod).
Do you think it would be much work to add a button in the cite page to load another random incident?
If you know you're coming from a random incident, maybe the button would say something like "Display another random incident" (or maybe something shorter and better), so it re-triggers the `/random/` path

lmcnulty:
@clari182 I think a good way to handle this is putting a random incident link between the next and previous links on the cite page, as is often done in webcomics. Then there's no need to determine if we came from the random page.

image

clari182 (approved):

Issue #2763, opened 2024-04-30
Checklists – Altering the query for a manual risk crashes the page


lmcnulty:
Error:

> One unhandled runtime error found in your files. See the list below to fix it:
> 
> Error in function updateRisk in `./src/components/checklists/CheckListForm.js:184`
> 
> "`generated`" is read-only
> 
> ```
> ./src/components/checklists/CheckListForm.js:184
> 
>   182 |       for (const attribute in attributeValueMap) {
>   183 |         if (attribute != 'precedents') {
> > 184 |           updatedRisk.generated = false;
>       |          ^
>   185 |         }
>   186 |         updatedRisk[attribute] = attributeValueMap[attribute];
>   187 |       }
> ```

PR #2768, opened 2024-05-02
Checklists fix modify query crash

lmcnulty:
Resolves https://github.com/responsible-ai-collaborative/aiid/issues/2763


site/gatsby-site/src/components/checklists/CheckListForm.js
```diff
@@ -174,17 +174,19 @@ export default function CheckListForm({
   };
 
   const updateRisk = (risk, attributeValueMap) => {
-    const updatedRisks = [...values.risks];
+    const updatedRisks = JSON.parse(JSON.stringify(values.risks));
```
cesarvarela, line 177: Be careful that this cloning method results in data loss, dates for example get turned into strings.

`lodash.cloneDeep` is better.

cesarvarela (approved):

site/gatsby-site/src/components/checklists/CheckListForm.js
lmcnulty: Thanks for the tip!

PR #2780, opened 2024-05-06
Fix/previous next tests

### LINKED ISSUE #2770: Previous/Next Incidents tests are broken
`Should disable Previous and Next incident buttons on first and last incidents` and `Should disable Previous and Next incident buttons in header on first and last incidents` are broken on staging

clari182:
- closes #2770


pdcp1 (approved):
LGTM 👍

Issue #2346, opened 2023-10-11
Global search bar feature


kepae:
We should consider adding a global search bar experience that is easier to access from all parts of the site.

Currently, there are two major ways to globally search AIID data: the homepage search feature component and the discover page.

However, if a user is currently on another page (incident citation, spatial visualization, scrolling through incident reports, etc.,) they user would have to navigate back to either of the above pages to search.

Notably, there is already the Discover button in the left-hand menu. It is an important function but is slightly redundant with the "Discover Incidents" link below.

<img width="442" alt="image" src="https://github.com/responsible-ai-collaborative/aiid/assets/4238598/77612297-cc4c-442a-88a6-409f501401a5">

Perhaps that button could instead expand to/be replace with a search prompt.
Alternatively, a minimalist search prompt could be embedded in the layout header when not on the Discover page.

(h/t @cesarvarela for the suggestion, as it relates to making use of the Algolia index we already maintain.)

PR #2792, opened 2024-05-20
Add global search bar

lmcnulty:
Resolves https://github.com/responsible-ai-collaborative/aiid/issues/2346.

I put it in the top-right corner, with an expanding icon. That's where I would generally expect to find such a feature. I agree that we might want to rethink the redundant quick action buttons, but I don't think a search bar would look good in their place.

https://github.com/responsible-ai-collaborative/aiid/assets/25443411/fa109361-fd39-4a0a-8df1-70cf54c184b7

### DIFF
--- a/site/gatsby-site/src/components/ui/Header.js
+++ b/site/gatsby-site/src/components/ui/Header.js
@@ -1,7 +1,7 @@
 import React, { useState } from 'react';
 import { StaticQuery, graphql } from 'gatsby';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { faBars, faRssSquare } from '@fortawesome/free-solid-svg-icons';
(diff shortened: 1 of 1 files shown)


kepae:
This is very cool. How's it look on mobile? I think I'm not getting the right responsiveness when I activate the menu button -- it doesn't activate under the hover as "clickable" (in the browser) and isn't responding to clicks.

lmcnulty:
@kepae Now works with or without hover.

If it's in a phone-sized viewport, we just hide it, because if it's already going to be hidden in the menu I figure at the point the user might as well just navigate to the home page or discover page to search.

https://github.com/responsible-ai-collaborative/aiid/assets/25443411/8c05e5b1-e970-4db4-836a-9c2672652ae6

kepae (approved):
Merge when ready! And consider taking this as an opportunity to write a "new" playright test?

PR #2796, opened 2024-05-21
Add default sidebar icon

lmcnulty:
We've had multiple situations where an icon was missing from a sidebar item because of the code and the Prismic data not being in sync. This adds a shapes icon as a fallback so it won't look as embarrassing if that happens again.

image


pdcp1 (approved):
LGTM

pdcp1:
@lmcnulty The code looks good! Before merging this, please resolve the conflicts on `treeNode.js`

PR #2797, opened 2024-05-23
Feature/similar reciprocity

### LINKED ISSUE #2793: Add/maintain symmetric relationship for "similar" incidents
Per @datherton09, a good suggestion on implementing links between related incidents such that they materialize on both incidents:

> Hi all -- I was doing some work on incidents 547 and 681, which are two now-older Ron DeSantis deepfake stories (as in his campaign made deepfakes). I marked 547 manually as being a related incident to 681. 681 just went live today -- what I noticed is that 547 was updated to reflect that it's related to 681, but 681 did not itself have any indication that it was linked to 547 as a result. I believe I had noticed this before but not really 'registered' it -- that we don't seem to have cross-linking among incident IDs we manually mark as related (the information seems? to remain just in the incident we mark, rather than automatically generating the same reflection in the related ID). Might be worth looking into, but not a burning priority.

That is -- similar incidents are not mutually recorded or visible from both incidents. An editor (and user) might expect this link to be mutual and apparent on the site.

Data details: The way we currently record "similar" incidents manually selected by editors for a given incident is by populating the `editor_similar_incidents` on incident:

For example, a source incident `A` being marked as similar to destination incident `B`:
```
{
  "incident_id": A,
  "nlp_similar_incidents": [...],
  "editor_similar_incidents": [ B ],
  "editor_dissimilar_incidents": [],
}
```

However -- this information is not available on incident `B`. What we have is an _edge between nodes in a graph_ that is not recoverable from one of the nodes.

If we treat similarity as _symmetric_, we should reasonably expect this information to show up on incident `B` as well. If we want to do _that_, then should we duplicate this information on incident `B`? How would we sync the information on `A` and `B` to point to each other or update if that changes?

Other systems specifically maintain tables or collection that exclusively track edges between documents of a certain type.

Let's use this as an opportunity to discuss perhaps a more general solution, if possible. We want links and backlinks between documents that have a relationship that are easy to query -- in the future considering entities, reports, incidents, and other datatypes we might wish to create.

clari182:
- closes #2793 

Test URL https://pr-2797--staging-aiid.netlify.app/

How to test
1. Login as an editor or admin
2. Go to an incident - Edit that incident
3. Scroll to `Similar Incident Id`
4. Assign a similar incident (search by incident ID and then select `Yes`)
5. Assign a dissimilar incident (search by Incident ID and then select `No`)
6. Save
7. Go to the similar and dissimilar incidents, click "Edit Incident" and check that they are linked the same way

### DIFF
--- a/site/gatsby-site/src/components/SimilaritySelector.js
+++ b/site/gatsby-site/src/components/SimilaritySelector.js
@@ -72,7 +72,7 @@ const SimilaritySelector = ({
               removeFromNotSureList(incident_id);
             },
           },
-        ].map((button) => {
+        ].map((button, index) => {
           let btnProps = {
             size: 'xs',
             'aria-pressed': button.show,
@@ -85,6 +85,10 @@ const SimilaritySelector = ({
             btnProps.color = button.variant;
           } else {
             btnProps.color = 'light';
+          }
+
+          if (index === 1) {
+            btnProps.className = 'rounded-none';
           }
 
           return (
--- a/site/gatsby-site/cypress/e2e/integration/cite.cy.js
+++ b/site/gatsby-site/cypress/e2e/integration/cite.cy.js
@@ -845,4 +845,79 @@ describe('Cite pages', () => {
 
     cy.waitForStableDOM();
   });
+
+  it('Should link similar incidents', () => {
+    cy.login(Cypress.env('e2eUsername'), Cypress.env('e2ePassword'));
+
+    cy.conditionalIntercept(
+      '**/graphql',
+      (req) =>
+        req.body.operationName == 'UpdateIncidents' &&
+        req.body.variables.set.editor_similar_incidents == 10,
+      'updateSimilarIncidents',
+      {
+        data: {
+          updateManyIncidents: {
+            matchedCount: 1,
+            modifiedCount: 1,
+          },
+        },
+      }
+    );
+
+    cy.conditionalIntercept(
+      '**/graphql',
+      (req) =>
+        req.body.operationName == 'UpdateIncidents' &&
+        req.body.variables.set.editor_dissimilar_incidents == 10,
+      'updateDissimilarIncidents',
+      {
+        data: {
+          updateManyIncidents: {
+            matchedCount: 1,
+            modifiedCount: 1,
+          },
+        },
+      }
+    );
+
+    cy.visit('/incidents/edit/?incident_id=10');
+
+    cy.waitForStableDOM();
+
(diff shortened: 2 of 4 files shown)


pdcp1:
@clari182 The logic works according to the "How to test" section. However, I'm having a hard time understanding the UX.

Both sections (`Manually-selected similar and dissimilar incidents` and `Reports`) disappear and reload on each change without a "loading" or a spinner.
In addition to that, every time I click on `Yes` or `No`, all reports change under the same incident at the same time. Is that the expected behavior?

If this is not part of the PR scope, let me know so I can create a separate issue.

I have attached a screen recording to visualize what I mean.

https://github.com/responsible-ai-collaborative/aiid/assets/6564809/05d603bf-7151-4e87-9bb0-47ec29e81c6a

site/gatsby-site/src/components/SimilaritySelector.js
```diff
@@ -87,6 +87,10 @@ const SimilaritySelector = ({
             btnProps.color = 'light';
           }
 
+          if (index === 1) {
+            btnProps.className = 'rounded-none';
+          }
+
```
pdcp1, line 93: What is the purpose of this special condition?

pdcp1 (changes requested):

site/gatsby-site/src/components/SimilaritySelector.js
clari182: There is an issue with group buttons from flowbite where if you have more than 2 buttons, the ones in the middle sill apply a rounded border when they shouldn't. So in order for it to look 'nice' I had to manually remove the rounded borders.
image

clari182:
Hey @pdcp1 ! Regarding how assigning a similar or dissimilar incident works, it's as expected but I agree the UI is a bit confusing and could be improved.
When you enter an incident ID, what you see `below` are the search results and what you see `above` is what you selected (if you select, `No`, `Not sure` or `Yes`).
The reason you're seeing it repeated, it's because it's displaying `all` the reports from that incident.

Also, when you're selecting more than one, in order to see them all, you need to click `Load more`.

Maybe as an improvement, we should only display one card per incident, instead of all the reports. But this is a separate issue.

site/gatsby-site/src/components/SimilaritySelector.js
pdcp1: Gotcha! thanks for the explanation

pdcp1:
@clari182 
> Maybe as an improvement, we should only display one card per incident, instead of all the reports. But this is a separate issue.

Indeed, that was something I was also considering: displaying just one card for each incident with the incident report links inside.

pdcp1 (approved):

PR #2801, opened 2024-05-27
Migrate fields to local API

cesarvarela:
It migrates the `incidents,` `reports,` `entities,` `submissions,` and `users` collections to our local API.

https://github.com/aiidtest/aiid/actions/runs/10172934601


kepae (commented):
first comments, vamos

PR #2802, opened 2024-05-28
Add intercept for updating incident

clari182:
Found an issue on my latest similar/dissimilar test, where an intercept was missing when updating incident


pdcp1 (approved):
LGTM 👍

PR #2812, opened 2024-05-31
Add date range selector histogram

lmcnulty:


clari182 (changes requested):
Looks good!
The only issue I found is that the left side of the range is not clearing when Clearing the filter, and neither when Clearing all filters. It stays selected.
image

PR #2837, opened 2024-06-12
Add playwright tests for TSNE visualization

lmcnulty:
These are ported versions of the currently-disabled cypress tests.


pdcp1 (changes requested):
The code looks good! Please remove the equivalent Cypress tests on `tsneVisualization.cy.js`. If all of them are migrated, please delete the whole`tsneVisualization.cy.js` file

pdcp1 (approved):

PR #2883, opened 2024-06-17
Fix live incident data not loading

### LINKED ISSUE #2880: Incident 704 not found (Live incident data not loading)
Searching for "We asked ChatGPT for legal advice – here are five reasons why you shouldn’t" will turn up incident 704:

image

However, clicking on the "Show Details on Incident #&#x2060;704" button results in a page indicating the incident cannot be found:

image

cesarvarela:
closes #2880 

image


kepae:
What would a test that catches this and other similar failures look like?

kepae (approved):

cesarvarela:
There are some already, but because we are mocking the API we are missing them.

PR #2946, opened 2024-07-03
Revise Citation Page Taxonomy Table Style

lmcnulty:
---

Now we show the taxonomy table in two columns so as to take up less space.

When the taxonomy is expanded, we switch to one column.

When there are longer field values, they look bad in multiple columns, so we show one column with fewer entries even without expanding.

The number of entries depends on their length. If they're very long we can have as few as one.

The taxonomy tables are now color-coded. We could turn this off, but I think it makes them easier to tell apart.

The default color is gray.


clari182 (changes requested):
Looks great! If I understand correctly, it's ok for now if they're all gray right?
It would be nice if when the classifications collapse, to scroll to the top of that classifications. Some have a lot of details and when I click "Show fewer classifications" I have to scroll back up a lot. Let me know what you think.

kepae:
I think that's a great suggestion. How about we do that in a follow-up PR? That way we can have smaller PRs and being incremental. :-)

Issue #2955, opened 2024-07-09
Empty "applied taxonomies" is confusing


smcgregor:
See image below. If I didn't know what a taxonomy is, I will be confused why the words "Applied Taxonomies" at the bottom. There should likely be a "None" or something similar saying nothing as been applied. The layout of this, effectively table data, could likely also be quickly improved to show it is a key/value table.

<img width="824" alt="Screenshot 2024-07-08 at 6 08 40 PM" src="https://github.com/responsible-ai-collaborative/aiid/assets/64780/52657738-f9ae-47b9-b68a-20c716b9cd4c">

PR #2962, opened 2024-07-09
Don't show applied taxonomies unless there is at least one

lmcnulty:
Per https://github.com/responsible-ai-collaborative/aiid/issues/2955.


pdcp1 (approved):
CR Note: In addition to not showing applied taxonomies unless there is at least one, this PR also hides the `_Annotator` items

pdcp1:
@lmcnulty The PR looks good! Please resolve the staging conflicts so we can merge it.

PR #2980, opened 2024-07-17
Update branch names to target `main` rather than `master`

kepae:
To be merged after following:
https://github.com/github/renaming


kepae:
Renamed `master` to `main`. The redirects are flowing. Local clients and copies must be updated, though they should never have been pushing directly to `master` in the first place.

pdcp1 (approved):
It looks good to me 👍 
Let's merge after following https://github.com/github/renaming

PR #2935, opened 2024-06-28
Feature/playwright locale risks api translator

clari182:
- closes #2932 
- closes #2933 
- closes #2934

PR #2826, opened 2024-06-07
Migrate citeEdit to playwright

clari182:
- closes #2839

PR #2989, opened 2024-07-22
Deploy to Production

pdcp1:
- #2935
- #2826
- #2993
- #2962


cesarvarela:
Now that I see that the `sinon` package was added, we should probably run those tests using `jest` 🤔

kepae (approved):

Issue #2990, opened 2024-07-23
Live `FindIncident` query failing; new incident emails arriving before Netlify deploy completes, citation page is 404


kepae:
I just received the email notifications for new incidents. However, at the time of writing, the citation pages for them are unavailable on incidentdatabase.ai. Further, the live incident fallback doesn't appear to retrieve from the DB dynamically, resulting in a 404/not found. I can reproduce this and find that it's an issue with the `FindIncident` query:

```graphql
{
  "errors": [
    {
      "message": "Variable \"$query\" got invalid value \"748\" at \"query.incident_id\"; Int cannot represent non-integer value: \"748\"",
      "locations": [
        {
          "line": 1,
          "column": 20
        }
      ]
    }
  ]
}
```

Hucking it over to you @cesarvarela for consistency, as this showed up with other API issues, e.g. https://github.com/responsible-ai-collaborative/aiid/issues/2987

Sounds like we need some more test coverage for API migrations. :-)
(Or we need to trust our test suite again when the playwright migration finishes 🤡 )

pdcp1:
I found myself in the same situation when I received the new incident emails yesterday evening.
FYI, the query error can be reproduced at any time on Staging or Production by navigating to a non-existing Incident (like https://incidentdatabase.ai/cite/800)

pdcp1:
After merging #2993 into Staging this issue was resolved.

I just created a new Incident on Staging and it loads fine ✅ 
https://staging-aiid.netlify.app/cite/756

Also, if I try to access a non-existing Incident (e.g: https://staging-aiid.netlify.app/cite/1000 ) the graphql call doesn't fail with an error, it just returns a null (as expected) and displays an `Incident 1000 not found` message ✅

PR #2993, opened 2024-07-29
parse incident_id as int for GraphQL API

kepae:
Fix:
* https://github.com/responsible-ai-collaborative/aiid/issues/2990


kepae:
@pdcp1 I haven't tested this locally yet as I haven't configured my dev environment to work with the local API accessing the Realm endpoints.

I know this file and other behavior are going to change with:
* https://github.com/responsible-ai-collaborative/aiid/pull/2801

Do you think it would be safe to merge this to staging for testing, and then pushing with the release?
As long as the incident pages don't actually break because of this minor change, it feels like an okay bandaid.

pdcp1:
@kepae Yes, I think it is safe to merge into Staging and test it there.

pdcp1 (approved):

PR #3002, opened 2024-08-01
Add full e2e workflows to production and staging

cesarvarela:
Should probably work on better names later, but this should be enough to get them running here: https://github.com/responsible-ai-collaborative/aiid/actions/workflows/staging.yml (without having to get them on `main`.)

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -77,8 +77,8 @@ jobs:
     needs: [cache-modifier]
     secrets: inherit
     with:
-      sha: ${{ github.event.pull_request.head.sha }}
-      environment: staging
+      sha: ${{ github.sha }}
+      environment: production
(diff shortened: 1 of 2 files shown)


.github/workflows/production.yml
```diff
@@ -71,6 +71,16 @@ jobs:
...
       runner-label: ${{ vars.PRODUCTION_RUNNER_LABEL }}
 
+  call-test-playwright-full:
+    if: ${{ !failure() && !cancelled() }}
+    uses: ./.github/workflows/test-playwright-full.yml
+    needs: [cache-modifier]
+    secrets: inherit
+    with:
+      sha: ${{ github.event.pull_request.head.sha }}
+      environment: staging
```
kepae, line 81: production?

kepae (changes requested):

kepae (approved):

Issue #3021, opened 2024-08-14
Migration Risk checklists to new API


kepae:
per @cesarvarela :
- Convert Atlas checklist `schema.json` file to a GraphqlObject class and save it to `server/types/checklist.ts` (gpt is excellent for this)
- Add a `checklists` field definition to `server/fields/cheklists.ts` (look at existing types for examples: `server/fields/incidents.ts`)
- Convert `rules.json` to `graphql-shield` permissions and add them to `server/fields/checklists.ts`
- Add a test fixture for checklists `server/tests/fixtures/checklists.ts` and add it to `server/tests/mutation-fields.spec.ts` and `server/tests/query-fields.spec.ts` 
- Migrate custom risks resolver to the checklists field file `server/fields/checklists.ts`
- Add a Custom test for this resolver `server/tests/checklists.ts`

To run tests, `npm run test:api`.

To start a node server with only the API `npm run start:api` (helpful for debugging)

Note that the Graphql fields `checklist` and `checklists` signature changes; instead of receiving a `query` argument, the argument's name is `filter,` and the properties differ, so the front end needs to be updated accordingly.

PR #3027, opened 2024-08-16
Fix/update tests

cesarvarela:


site/gatsby-site/playwright/e2e-full/apps/checklistIndex.spec.ts
```diff
@@ -0,0 +1,251 @@
+import { expect, request } from '@playwright/test';
```
cesarvarela, line 1: ditto: https://github.com/responsible-ai-collaborative/aiid/issues/3021

.github/workflows/test-playwright-full.yml
```diff
@@ -28,8 +28,8 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        shardIndex: [1, 2, 3, 4, 5, 6]
-        shardTotal: [6]
+        shardIndex: [1, 2, 3, 4, 5, 6, 7, 8]
+        shardTotal: [8]
```
cesarvarela, line 32: More shards reduce the chances of the web server crashing because less tests are run per worker

site/gatsby-site/playwright.config.ts
```diff
@@ -39,7 +39,7 @@ export default defineConfig({
   projects: [
     {
       name: 'chromium',
-      use: { ...devices['Desktop Chrome'] },
+      use: { ...devices['Desktop Chrome'], launchOptions: { args: ['--auto-open-devtools-for-tabs'] } },
```
cesarvarela, line 42: I found myself adding this line every time, so I think it is a useful default.

site/gatsby-site/playwright/e2e-full/cite.spec.ts
```diff
@@ -567,8 +559,6 @@ test.describe('Cite pages', () => {
 
     test('Should link similar incidents', async ({ page, login }) => {
 
-        test.slow();
```
cesarvarela, line 570: I've found that `tests.slow()` makes things worse by delaying the time to get tests results, and overall it is not necessary, we might want to delete them all later.

site/gatsby-site/server/rules.ts
```diff
@@ -22,4 +27,28 @@ export const isRole = (role: string) => rule()(
     },
 )
 
+export const isSelf = () => rule()(
```
cesarvarela, line 28: My idea here is that only `admin` users can query data about a user other than themselves.

We might need to tweak this in the future so `first_name` and `last_name` are public.

site/gatsby-site/server/remote.ts
```diff
@@ -47,7 +47,7 @@ const ignoreTypes = [
     'CreateVariantInput',
 
     // 'Incident',
-    'IncidentQueryInput',
+    // 'IncidentQueryInput',
```
pdcp1, line 50: Does the `TODO` on line 27 and these comments make sense? I'm not clear

pdcp1 (approved):

site/gatsby-site/server/remote.ts
cesarvarela: Yes, what's happening there is that the subscription -> user relation breaks if we ignore the `UserQueryInput` type.

Issue #3023, opened 2024-08-14
Fix CSET taxonomies snapshots and CSV exports


pdcp1:
As part of this issue, we should:

- [x] Fix the existing CSET database dump by using the correct namespace
  - https://github.com/responsible-ai-collaborative/aiid/pull/3030 
- [x] Convert the CSET taxonomy CSV output into a more user-friendly format
  - https://github.com/responsible-ai-collaborative/aiid/pull/3082
- [x] Export GMF taxonomy in a similar CSV format
  - https://github.com/responsible-ai-collaborative/aiid/pull/3082
- [x] Analyse the possibility of migrating the DB dump code ([mongodb-awesome-backup](https://github.com/responsible-ai-collaborative/mongodb-awesome-backup)) to this repo as a Github action 
  - https://github.com/responsible-ai-collaborative/aiid/pull/3030

kepae:
And after fixing + updating, also export the GMF taxonomy in a similar format such that the CSVs could be compared, if desired. :-)

pdcp1:
All items are done. We can close this issue

pdcp1:
@responsible-ai-collaborative/aiid-dev 
I just checked the [first automatic backup](https://github.com/responsible-ai-collaborative/aiid/actions/runs/10881621540) (it runs each Monday) and it works as expected 👍 

<img width="310" alt="image" src="https://github.com/user-attachments/assets/2892be4c-55f2-4029-9279-74f4836ca558">



This is the file generated: https://pub-a5fe3e44369c4cabb576fa0d2c09fdf6.r2.dev/backup-20240916100453.tar.bz2

PR #3030, opened 2024-08-16
Migrate all `mongodb-awesome-backup` repo code/actions to this repo

pdcp1:
Part of issue #3023
 
This PR moves the whole GitHub Actions code from [mongodb-awesome-backup](https://github.com/responsible-ai-collaborative/mongodb-awesome-backup) to this repo.
It also fixes the broken CSET CSV file export.

The following GitHub Action is added. 
- `.github/workflows/db-backup.yml`

This action is responsible for generating the DB backup and [DB snapshots](https://incidentdatabase.ai/research/snapshots/). 

### New GitHub Actions environment secrets needed
These Cloudflare credentials must have write permissions
- `secrets.CLOUDFLARE_R2_WRITE_ACCESS_KEY_ID`
- `secrets.CLOUDFLARE_R2_WRITE_SECRET_ACCESS_KEY`

### Testing

This code can be tested on Pablo's environment https://github.com/pdcp1/aiid/actions/workflows/db-backup.yml

### Release considerations

Both GitHub Actions (`Public backup to the cloud` and `Private backup to the cloud`) on the [original repo](https://github.com/responsible-ai-collaborative/mongodb-awesome-backup/actions) should be disabled as soon as we release this PR into Production.


kepae:
Two thoughts:

* I'm wondering about the explosion of environment variables in this pipeline. They add some complexity to our repo, and I forgot how many there were! Is there something we can do to isolate these variables from the rest of AIID? or make clear which ones are shared? I'm thinking more about this.
* We might not need a private backup anymore given the release of the CSET data!

Let's talk about these in our weekly dev meeting.

pdcp1:
> * I'm wondering about the explosion of environment variables in this pipeline. They add some complexity to our repo, and I forgot how many there were! Is there something we can do to isolate these variables from the rest of AIID? or make clear which ones are shared? I'm thinking more about this.

I have the same feeling about the complexity added. But I think we can reuse some of the env variables/secrets. For example the Cloudflare variables.
Considering the GitHub limitations on variables/secrets, one of the alternatives to keep them separate from the AIID site variables is creating a new GitHub environment (e.g.: "backups") and putting all backup-specific variables there.

> We might not need a private backup anymore given the release of the CSET data!

Also, if we can skip the private backup processing, it will reduce code complexity and the number of required env variables.

Let's discuss it tomorrow with the rest of the team!

pdcp1:
After some refactoring, we only need to add two more environment secrets. Both secrets are needed to enable write access to Cloudflare R2 bucket

.github/workflows/db-backup.yml
```diff
@@ -0,0 +1,55 @@
+name: Public backup to the cloud
+
+on:
+  schedule:
+    - cron: "0 10 * * 1" # At 10:00 on Monday.
+  workflow_dispatch:
+    inputs:
+      home:
+        description: "This can be triggered from the GH page"
```
cesarvarela, line 9: This is because to have a `workflow_dispatch`, you need at least 1 input? Otherwise, could it be deleted?

cesarvarela (commented):
If we consider this repository a "monorepo," it should be in `/site/db-backup/*`. What do you think?

pdcp1:
> If we consider this repository a "monorepo," it should be in `/site/db-backup/*`. What do you think?

@cesarvarela: When we discussed the folder structure with @kepae, we decided to have a `tools` folder where we can potentially add more tools in the future. However, considering the `site` folder with subfolders such as `gatsby-site` and `realm`, maybe it's better to move the `db-backup` there.

.github/workflows/db-backup.yml
pdcp1: The inputs in this case are not mandatory; they were only for documentation purposes. I just deleted them. 👍

cesarvarela (approved):

.github/workflows/db-backup.yml
```diff
@@ -0,0 +1,50 @@
...
+          ./bin/backup.sh
+          ./bin/prune.sh
+          ./bin/list.sh
+        working-directory: site/db-backup
+        env:
+          CLOUDFLARE_R2_ACCOUNT_ID: ${{ vars.CLOUDFLARE_R2_ACCOUNT_ID }}
+          CLOUDFLARE_R2_WRITE_ACCESS_KEY_ID: ${{ secrets.CLOUDFLARE_R2_WRITE_ACCESS_KEY_ID }}
+          CLOUDFLARE_R2_WRITE_SECRET_ACCESS_KEY: ${{ secrets.CLOUDFLARE_R2_WRITE_SECRET_ACCESS_KEY }}
+          CLOUDFLARE_R2_BUCKET_NAME: ${{ vars.CLOUDFLARE_R2_BUCKET_NAME }}
+          MONGODB_URI: ${{ secrets.MONGODB_CONNECTION_STRING }}
```
cesarvarela, line 50: we are not specifying which environment to get this secret from 🤔

cesarvarela (changes requested):
I'm taking my approval back because I'm noticing that we are not telling the workflow which environment it should get variables from (staging, production, etc.).

pdcp1:
> I'm taking my approval back because I'm noticing that we are not telling the workflow which environment it should get variables from (staging, production, etc.).

The environment input was added and it's ready for review

smcgregor:
@pdcp1 Please ping me if/when I can retire the old repo.

cesarvarela (approved):

kepae:
Secrets updated!
image

pdcp1:
> @pdcp1 Please ping me if/when I can retire the old repo.

@smcgregor According to our Slack conversation, @kepae will be in charge of this repo deactivation.

PR #3036, opened 2024-08-23
Feature/entity entity relationship

### LINKED ISSUE #2975: Support entity-entity relationships
Now that we have a base interface for editing entities, we can more easily curate their basic data and potential relationships between them. Following up @pdcp1's mock here, we can now start to implement basic relationships, starting with a catch-all `related` relationship while we figure out a more durable system to commit to editing.

The primary function this would serve is to **relate our soup of alleged developers, deployers, and victims implicated in AI incident reports.**

Additionally, an entity relationship system can allow us to **record and name AI systems and models and relate them to incidents, developers, and other AIID data.**

Per https://github.com/responsible-ai-collaborative/aiid/issues/2536#issuecomment-1960443680:
> An initial mockup for the Entity edit page is in our Figma workspace https://www.figma.com/file/KI28jWrOO3soKp9dTg7b0d/Entities-workflow?type=design&node-id=0%3A1&mode=design&t=a1M9nWWeHzoisF2d-1
>
> We should take this as a kick-off design. I added some fields that I consider appropriate but we can discuss which ones are more important to implement in an initial phase.
> Feel free to edit it. I'm open to any suggestions and discussions.
>
> image

### Option 1
Simply extend the entity object with an array of `related` entity IDs.

```js
{
  "entity_id": "amazon-warehouse-workers",
  "name": "Amazon warehouse workers"
  "related": [
      "amazon", // entity_id
       "warehouse-workers"
  ]
}
```

Pros:
* Simpler at first. Defers having to define further what "related" means.
* Finding related entities does not require a secondary lookup in linked data.
* Easy to migrate from later on.

Cons:
* `related` is a symmetric relationship; we would want both entities to have the property. Thus, we would have to synchronously update two entities that are currently stored separately in the `entities` collection – and ideally validate this continuously.
* Further validation: entities and their IDs in relationships must exist and be well-formed. We currently lack validation over entity data.
* Flexibility and hard-coding. It's relatively easy to expand/migrate from this implementation later on, but we are hard-coding the field "related" onto an entity. When we wish to include another entity-entity relationship (e.g. `developed_by`), will we just add another field to the mongo document? Or begin the migration to....

### Option 2
Create a collection specifically for entity-entity relationships, i.e. an edge table. We could store these as *basic* semantic triples (subject, object, predicate), with additional metadata about whether or not the property/relationship is symmetric.

We would be entering an early implementation of semantic triples: https://en.wikipedia.org/wiki/Semantic_triple

```js
{
  "pred": "related",
  "sub": "amazon-warehouse-workers",
  "obj": "amazon",
  "is_symmetric": true
}
```

Pros
* Start advancing towards semantic and linked data. :-)
* More flexible if we wish to start adding other property relationships already in our data, e.g. `developed_by`.
* Easier to start adding other relationships (e.g. `developed_by`, which would be complemented with a `develops` relationship).
* Easier to start adding metadata on the relationship (e.g. status, confidence, symmetry, provenance).

Cons
* Opening up the linked data storm!
* Lookups become more complicated, both in static and dynamic contexts. This would have implications for the GraphQL API we are developing.
* Validation: As with above, the ideal implementation has a way to validate that all edges/relationships are valid, i.e. connect two well-formed and existing entities with valid IDs.

### Other options
* MongoDB purports to have useful graph-like traversal qualities that we could explore using directly: https://www.mongodb.com/resources/basics/databases/mongodb-graph-database

### Nota Bene
We are casually entering further into the territory of linked/semantic data. Many implementations and standards for such data already exist, e.g. [RDFa](https://en.wikipedia.org/wiki/RDFa), [JSON-LD](https://en.wikipedia.org/wiki/JSON-LD).

We should consider how we can make existing standards of linked data work for us in the future. Incident data and reports of AI harms are inherently graph-like in the real world as well as on AIID.

### Related issues
* https://github.com/responsible-ai-collaborative/aiid/issues/2536
* https://github.com/responsible-ai-collaborative/aiid/issues/1437

clari182:
- closes #2975 
Test url: https://pr-51--clarayoudale-staging.netlify.app/
ping me for credentials

image
image
image


site/gatsby-site/server/fields/entity_relationships.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    ...generateMutationFields({ collectionName: 'entity_relationships', Type: Entity_relationshipType, generateFields: ['updateOne', 'upsertOne', 'deleteOne'] }),
+}
+
+export const permissions = {
+    Query: {
+        entity_relationships: allow,
+    },
+    Mutation: {
+        updateOneEntity_relationship: allow,
```
cesarvarela, line 24: One thing to note is that we don't have to stick to these mutations anymore, especially if we are performing more than one operation in a UI flow, i.e., it is better to have one `promoteSubmission` mutation that touches multiple collections than relying on separate small mutations like these. (This might not be the case just flagging this up)

cesarvarela (changes requested):
Please add a test fixture for the new collection. Look for inspiration in the `server/tests/fixtures` folder.

site/gatsby-site/playwright/e2e-full/entityEdit.spec.ts
```diff
@@ -103,4 +103,73 @@ test.describe('Edit Entity', () => {
...
     }
   );
+
+  test('Should successfully add Entity Relationship', async ({ page, login, skipOnEmptyEnvironment }) => {
+  
+    const userId = await login(process.env.E2E_ADMIN_USERNAME, process.env.E2E_ADMIN_PASSWORD);
+
+    await init();
+  
+    await init({ customData: { users: [{ userId, first_name: 'Test', last_name: 'User', roles: ['admin'] }] }, }, { drop: true });
```
cesarvarela, line 113: I updated the login fixture so it automatically adds the user to the db, you should only need to do 

```
init() // if necessary

login(...)

```

cesarvarela: (Look at existing examples once you merge latest staging)

site/gatsby-site/playwright/e2e-full/entityEdit.spec.ts
```diff
@@ -103,4 +103,73 @@ test.describe('Edit Entity', () => {
...
+    await expect(page.locator('.tw-toast')).toContainText('Entity updated successfully.');
+  });
+
+  test('Should successfully remove Entity Relationship', async ({ page, login, skipOnEmptyEnvironment }) => {
+  
+    const userId = await login(process.env.E2E_ADMIN_USERNAME, process.env.E2E_ADMIN_PASSWORD);
+
+    await init();
+  
+    await init({ customData: { users: [{ userId, first_name: 'Test', last_name: 'User', roles: ['admin'] }] }, }, { drop: true });
```
cesarvarela, line 148: ditto

site/gatsby-site/playwright/e2e/entities.spec.ts
```diff
@@ -11,50 +13,64 @@ test.describe('Entities page', () => {
...
-      expect(count).toBeGreaterThanOrEqual(10);
+      expect(count).toBeGreaterThanOrEqual(entities.length);
     });
   });
 
   test('Filter entities by name', async ({ page, skipOnEmptyEnvironment }) => {
+    await init();
     await page.goto(url);
-    await page.locator('[data-cy="input-filter-Entity"]').fill('Amazon');
+    await page.locator('[data-cy="input-filter-Entity"]').fill('Entity 1');
```
cesarvarela, line 36: The tests found in the `/e2e/` folder, which I think we'll have to rename to something like `/smoke,` run against production data. There should not be an entity called `Entity 1` there. This should probably be moved to the `e2e-full` folder or left as it is.

site/gatsby-site/playwright/e2e/entity.spec.ts
```diff
@@ -4,10 +4,11 @@ import emptySubscriptionsData from '../fixtures/subscriptions/empty-subscription
 import subscriptionsData from '../fixtures/subscriptions/subscriptions.json';
 import { SUBSCRIPTION_TYPE } from '../../src/utils/subscriptions.js';
 import config from '../config';
+import { init } from '../memory-mongo';
 
 const entity = {
-  entity_id: 'google',
-  name: 'Google',
+  entity_id: 'entity1',
```
cesarvarela, line 10: ditto

site/gatsby-site/playwright/e2e/entity.spec.ts
```diff
@@ -111,12 +115,14 @@ test.describe('Individual Entity page', () => {
...
 
   test('Should not display Edit button for unauthenticated users', async ({ page }) => {
+    await init();
     await page.goto(url);
     await expect(page.locator('[data-cy="edit-entity-btn"]')).not.toBeVisible();
   });
 
   test('Should display Edit button only for Admin users', async ({ page, login, skipOnEmptyEnvironment }) => {
-    await login(config.E2E_ADMIN_USERNAME, config.E2E_ADMIN_PASSWORD);
+    const userId = await login(config.E2E_ADMIN_USERNAME, config.E2E_ADMIN_PASSWORD);
```
cesarvarela, line 124: ditto

site/gatsby-site/server/fields/entities.ts
```diff
@@ -1,28 +1,107 @@
...
 
-    ...generateMutationFields({ collectionName: 'entities', Type: EntityType, generateFields: ['updateOne', 'upsertOne'] }),
-}
+  ...generateMutationFields({ collectionName: 'entities', Type: EntityType, generateFields: ['updateOne', 'upsertOne'] }),
 
-export const permissions = {
-    Query: {
-        entity: allow,
-        entities: allow,
+  updateOneEntity: {
```
cesarvarela, line 41: I would use names for custom resolvers that do not coincide with generated ones to avoid confusion, but it is not critical.

cesarvarela: Also, please add an API test for this custom resolver. It should be in `server/tests/entities.ts`, (look at `server/tests/reports.ts` for inspiration)

cesarvarela (changes requested):

cesarvarela:
@clari182 please merge staging

cesarvarela (commented):
One small comment: please get tests to pass.

cesarvarela (changes requested):
No reason for test to be failing on this PR since it does not depend on an Atlas push

cesarvarela (approved):

PR #3045, opened 2024-08-28
Allow local user account to be set via environment

cesarvarela:
This is part of the DX improvements of working against the in memory mongo server


cesarvarela:
This wont be necessary because of #3220

PR #3100, opened 2024-09-16
Make sure only valid entity names are used on initial form state

cesarvarela:
Fixes: https://raicollab.slack.com/archives/C03L4FLEL2Y/p1726433890359339

### DIFF
--- a/site/gatsby-site/src/components/submissions/SubmissionEdit.js
+++ b/site/gatsby-site/src/components/submissions/SubmissionEdit.js
@@ -166,15 +166,17 @@ const SubmissionEdit = ({ id }) => {
                   developers:
                     submission.developers === null
                       ? []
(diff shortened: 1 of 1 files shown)


site/gatsby-site/src/components/submissions/SubmissionEdit.js
```diff
@@ -166,15 +166,17 @@ const SubmissionEdit = ({ id }) => {
                   developers:
                     submission.developers === null
                       ? []
-                      : submission.developers.map((item) => item.name || item),
```
cesarvarela, line 169: The submission had a `harmed_party` with an empty name and was causing the Tags component to crash:

image

I'm not a fan of using entity names instead of ids, but this works for now.

kepae (approved):

Issue #3097, opened 2024-09-11
Email notifications are not working on Production


pdcp1:
From production Rollbar:

[[Process Pending Notifications: Submission Promoted]: read not permitted](https://app.rollbar.com/a/seanbmcgregor/fix/item/AIID/6857)

-----

Email notifications are working correctly on the GitHub action on Staging.

This is the Atlas MongoDB log on Staging:

<img width="1373" alt="image" src="https://github.com/user-attachments/assets/f2fd138b-700d-4647-b955-2c35b6d0d4e3">

It uses this user to call the `processNotification` mutation from the GitHub action:

**User name:** `staging-graphql`
**User id:** `624e7e038fd3c8e5c7770aa0`

<img width="1567" alt="image" src="https://github.com/user-attachments/assets/b4c3810f-b390-4b03-993e-287b3aff0053">

@kepae We should check the same on Production and see if the user has the permissions needed.

cesarvarela:
I will start migrating this collection to our API today or tomorrow, so the problems at least will be different 🥳

kepae:
I believe this issue is now fixed with the cleanup in #3104.

PR #3104, opened 2024-09-17
Update incidents with `editor_notes` or `flagged_dissimilar_incidents` fields null

pdcp1:
This fixes the issue that we have with the Atlas MongoDB incidents schema, which doesn't allow `null` values on fields:
- `editor_notes`
- `flagged_dissimilar_incidents`

Schema reference: `site/realm/data_sources/mongodb-atlas/aiidprod/incidents/schema.json`
```
"editor_notes": {
  "bsonType": "string"
}
```

```
"flagged_dissimilar_incidents": {
  "bsonType": "array",
    "items": {
      "bsonType": "int"
    }
}
```

The new API allows that, but some code still uses the Atlas MongoDB schema validation.

## Problems generated by this

- "Should retrieve risks with no tag provided" test failure
- It probably also fixes the Notifications process: https://github.com/responsible-ai-collaborative/aiid/issues/3097


cesarvarela (commented):
Should we make then required in the graphql data type?  Using `GraphqlNonNull` in the type definition.

pdcp1:
> Should we make them required in the graphql data type? `GraphqlNonNull` is used in the type definition.

@cesarvarela If we use `GraphQLNonNull`, are we forcing it to be not `null` every time the API is called?

pdcp1:
@cesarvarela this is ready for another review. The test is still failing in this preview because the migration will run once it is merged into Staging

cesarvarela (approved):

PR #3123, opened 2024-09-25
Add glowing aiid logo for emails

lmcnulty:
The [email template update PR](https://github.com/responsible-ai-collaborative/aiid/pull/3099) currently embeds the logo image with a data url. This is not supported everywhere, and the easiest solution is just to add it as a static asset to the website and then link it in the email.


lmcnulty:
We're using a cloudinary URL instead, so this is not necessary.

PR #3160, opened 2024-10-16
Migrate reports and incidents history to the new API

cesarvarela:
closes https://github.com/responsible-ai-collaborative/aiid/issues/3194

preview: https://pr-3160--staging-aiid.netlify.app

(might need to be whitelisted)


pdcp1:
@cesarvarela could you add some testing instructions and/or a preview link environment?

pdcp1 (approved):
LGTM 👍 
A side note for next PRs. I saw some changes here that are not strictly related to the incidents/reports history migration, and I wonder if they could be separate in another PR.

Issue #2886, opened 2024-06-18
Create a smoke test suite


cesarvarela:
It should be hosted in a `smoke` folder next to the `e2e` folder.

Playwright docs show an example implementation using projects:

https://playwright.dev/docs/test-projects#splitting-tests-into-projects

kepae:
I would propose we develop probes and data tests. These are slightly more descriptive than "smoke" tests for our purposes.

Probes would test the accessibility of features on the live production AIID site using Playwright. These include: landing page, review queue, risk checklists, news digest, and other dynamic features that rely on external APIs.

Data tests could strictly query the MongoDB with assertions about the data. For example, no `null` fields on incident objects (e.g. https://github.com/responsible-ai-collaborative/aiid/pull/3104), or all incidents have reports, etc.

cesarvarela:
Should we then rename current playwright folders as:

e2e-full -> e2e
e2e -> probes

And create a new `data` folder.

scottslewis:
@cesarvarela and @kepae 

In commit:  https://github.com/scottslewis/aiid/commit/199ddaeeea49efa38ebc8af0aedc99d394fd66aa

I've moved probes.ts to a new directory (e2e/probes/) and refactored the tests so that there is one const 2-dim array that defines the tests and now have only one test method.  That test method now iterates through the array to call testUrl (urlPrefix, paths) for each url prefix/paths combination defined in urls1.

Feel free to make any desired stylistic or doc changes.

PR #3161, opened 2024-10-17
Added draft probes.ts file in e2e directory

scottslewis:
Added draft probes.ts file in e2e directory as per issue

https://github.com/responsible-ai-collaborative/aiid/issues/2886


site/gatsby-site/playwright/probes/probes.ts
```diff
@@ -0,0 +1,40 @@
...
+		paths: ['/apps/newsdigest/']
+	},
+	{ 
+		title: 'Probing staging checklist page', 
+		// baseUrl: null means use value of config.gatsby.siteUrl
+		baseUrl: 'https://staging-aiid.netlify.app', 
+		paths: ['/apps/checklist/']
+	}
+].forEach(({title, baseUrl, paths}) => {
+	test(title, () => testUrls(baseUrl, paths));
```
kepae, line 39: Just checking: I believe `testUrls()` will run nested Playwright tests. What does the output of this test look like?

scottslewis: wrt testUrls, I was assuming that this import would override

import { testUrls } from '../e2e/pages-tests';

But my scoping understanding might be wrong.

What does the output of this test look like?

I haven't been able to run this test locally (although I did test/run for syntax locally without imports.   I had some problems with some config/setup/dependency, etc.    I admit I was hoping someone else could run it after fussing with playwright and existing deps/tests for too long.   I'll return to fussing perhaps today, but in following the existing instructions I couldn't get a functional environment to run this test only with something like (paraphrasing):  npx playwright probes/probes.ts...which seems correct via playwright docs.   Again, I'll fuss with it some more and open an issue if update to latest (on staging branch?) doesn't help.   Thanks.

PR #3197, opened 2024-11-01
Use bcc for email notifications

cesarvarela:


kepae:
Can we add a test here? that the email request will not result in a filled out "TO", max num recipients is 1, etc?

PR #3200, opened 2024-11-05
Feature/update entity list

### LINKED ISSUE #3048: New entity entry doesn't update entity list on forms
On the submit, submission and incident forms:
When a new entity is entered in `developer`, `deployer` or `harmed parties` it's not added to the dropdown for the other ones.

image

If you add the same `new` entity for all fields, it will only be added to the first one.

clari182:
- closes #3048 

Testing preview link: https://pr-3200--staging-aiid.netlify.app/


pdcp1 (changes requested):
The code and the UX look good. Please add/update a test

pdcp1 (approved):

PR #3208, opened 2024-11-08
Fix/contents markdown

clari182:


site/gatsby-site/src/utils/extractHeaders.js
```diff
@@ -9,7 +9,12 @@ export const extractHeaders = (content) => {
   if (content.length > 0) {
     content.forEach((block) => {
       richText = block.text.richText;
-      headers.push(...extractHeadersFromRichText(richText));
+      if (richText && richText.length > 0) {
+        headers.push(...extractHeadersFromRichText(richText));
+      }
+      if (block?.markdown && block.markdown.richText.length > 0) {
+        headers.push(...extractHeadersFromMarkdown(block.markdown.richText[0]?.text));
+      }
```
pdcp1, line 17: I think we should change these to use an `else`. If both conditions are true we are going to have two items in the headers list variable

pdcp1 (approved):

PR #3210, opened 2024-11-08
Run notification process as a script

### LINKED ISSUE #3217: Email Notification System: Action Items & Recovery Plan
## Fixes & Prevention
- [x] Confirm original bugfix for email addressing is in production (César)
- [x] Add email pipeline validation function:
  - [x] Implement ASSERT-like function for single recipient validation
  - [x] Add unit tests for validation function
- [x] Improve error reporting for process notification API
- [x] Disable notifications during forced deploys

## Review & Security
- [x] Verify Mailersend token pause functionality
- [x] Verify user data isolation from staging backups

## Cleanup & Recovery
- [x] Run migration to remove pending notifications
- [x] Re-enable notification system

cesarvarela:
closes #3217 

Refactor process notification mutation to run as a script. This allows us to bypass the 26-sec Netlify function limit.

Successful run: 
https://github.com/aiidtest/aiid/actions/runs/11785936273/job/32828184850#step:5:1

failed run (no emails when calling mailersend): https://github.com/aiidtest/aiid/actions/runs/11789126074/job/32837365540#step:5:1

#### Instructions before merging

Secrets that need to be added to the GitHub environment:

- `ROLLBAR_POST_SERVER_ITEM_ACCESS_TOKEN`
- `MAILERSEND_API_KEY`

Variables:

- `NOTIFICATIONS_SENDER`
- `NOTIFICATIONS_SENDER_NAME`
- `REALM_APP_ID`

#### architecture considerations

Instead of executing the notification process inside the API hosted in a Netlify function, we are now executing a Node script inside a GitHub Worker.

- default timeout goes from 26 secs to 6 hours
- logs are now public, including errors; we should not leak sensitive data on script crashes.
- the script shares code with the API but lives outside the API


site/gatsby-site/server/fields/common.ts
```diff
@@ -90,140 +88,82 @@ export interface UserAdminData {
...
 
-    const emailTemplateBody = templates[templateId];
+/**
+ * Fetches the access token for the MongoDB Atlas Admin API.
+ * 
+ * Note: The token is cached to speed up subsequent requests. Tokens expire after 30 minutes.
+ * 
+ * @returns {Promise<string>} A promise that resolves to the access token.
+ */
+const getAccessToken = async () => {
```
cesarvarela, line 129: This is not specifically related to the fix.

But it will help speed up the notification process and every other request.

site/gatsby-site/postBuild/processNotifications.js
cesarvarela, line 1: Looks like this file should have been deleted

.github/workflows/process-notifications.yml
```diff
@@ -1,41 +1,50 @@
...
     inputs:
-        environment:
-          description: The Github environment to load secrets from
-          type: string
-          required: true
+      environment:
+        description: The Github environment to load secrets from
+        type: string
+        required: true
+      force-deploy:
```
kepae, line 9: why do we need force deploy on this script? is this trying to inherit from the deploy workflow?

These env vars are feeling brittle; is there a way we can control this logic from the deploy workflow more directly?

site/gatsby-site/server/emails/index.ts
```diff
@@ -0,0 +1,95 @@
...
+    });
+}
+
+export const mailersendBulkSend = async (emails: EmailParams[]) => {
+
+    const mailersend = new MailerSend({
+        apiKey: config.MAILERSEND_API_KEY,
+    });
+
+    assert(emails.every(email => email.to.length == 1), 'Emails must have exactly one recipient');
```
kepae, line 43: This is I would extend this to handle both `to` + `cc` <= 1. (or == 1).

thoughts?

site/gatsby-site/server/emails/index.ts
```diff
@@ -0,0 +1,95 @@
...
+                .setTo([new Recipient(recipient.email, '')])
+                .setPersonalization(personalizations)
+                .setSubject(subject)
+                .setHtml(html);
+            //TODO: add a text version of the email
+            // .setText("Greetings from the team, you got this message through MailerSend.");
+
+            bulk.push(emailParams);
+        }
+
```
kepae, line 86: can we move the `try` to start here? The above logic is all initialization that cannot fail, correct?

I believe this would be more readable and we shouldn't lose error/exception coverage.

site/gatsby-site/server/fields/common.ts
```diff
@@ -90,140 +88,82 @@ export interface UserAdminData {
...
-        response.creationDate = new Date(userApiResponse.creation_date * 1000);
-        response.lastAuthenticationDate = new Date(userApiResponse.last_authentication_date * 1000);
-        response.disabled = userApiResponse.disabled;
-    }
-
-    return response;
-}
-
-export const getUsersAdminData = async () => {
+const usersCache: UserAdminData[] = [];
```
kepae, line 91: Document this global value please... it's getting mutated by a "get" function, which is indirect. Maybe `getAndCacheUserAdminData()` is better.

site/gatsby-site/server/fields/common.ts
```diff
@@ -90,140 +88,82 @@ export interface UserAdminData {
...
+ * 
+ * Note: The token is cached to speed up subsequent requests. Tokens expire after 30 minutes.
+ * 
+ * @returns {Promise<string>} A promise that resolves to the access token.
+ */
+const getAccessToken = async () => {
 
-    if (!emailTemplateBody) {
-        throw new Error(`Template not found: ${templateId}`);
+    // Refresh the authentication token well before expiration to avoid interruptions\
```
kepae, line 131:
```suggestion
    // Refresh the authentication token well before expiration to avoid interruptions.
```

site/gatsby-site/server/fields/common.ts
```diff
@@ -90,140 +88,82 @@ export interface UserAdminData {
...
+    if (cachedToken && tokenExpiration && Date.now() < tokenExpiration * 1000 + 5 * 60 * 1000) {
```
kepae, line 132: tokenExpiration is a duration, not DateTime? comment this length of time computed? "Return the token if it is valid for at least X duration"?

site/gatsby-site/server/fields/common.ts
```diff
@@ -90,140 +88,82 @@ export interface UserAdminData {
...
+        throw new Error(`Login failed: ${data.error}`);
+    }
 
-        await mailersendBulkSend(bulk);
+    const decoded = jwt.decode(data.access_token) as { exp: number };
 
-    } catch (error: any) {
-        error.message = `[Send Email]: ${error.message}`;
-        reporter.error(error);
+    cachedToken = data.access_token;
```
kepae, line 155: any further validation here?

site/gatsby-site/server/fields/common.ts
```diff
@@ -233,36 +173,24 @@ export const sendEmail = async ({ recipients, subject, dynamicData, templateId }
  * 
  * @throws {Error} Throws an error if an unsupported HTTP method is provided.
  */
-export const apiRequest = async ({ path, method = "GET" }: { method?: string, path: string }) => {
+export const apiRequest = async ({ path, params = {}, method = "GET" }: { method?: string, params?: Record<string, string>, path: string }) => {
```
kepae, line 176: (nitpick: we might consider abstracting all Atlas-facing stuff to its own module, so that this file can read Atlas-agnostic...)

site/gatsby-site/src/scripts/process-notifications.ts
kepae, line 1: (I'm still reading this refactor.)

kepae (changes requested):
Still in process but here are some changes and questions. Looking great!

.github/workflows/process-notifications.yml
cesarvarela: It is to skip the notifications workflow if the current deploy is a forced one.

site/gatsby-site/server/fields/common.ts
cesarvarela: Hopefully we'll drop Atlas altogether before that 😉

site/gatsby-site/server/fields/common.ts
cesarvarela: I moved the cache to the script itself to avoid keeping a global cache, as you mentioned.

site/gatsby-site/server/fields/common.ts
cesarvarela: `tokenExpiration` is the date the token expires. I added a new test for that and the caching function.

cesarvarela: site/gatsby-site/server/tests/atlas.spec.ts

kepae (approved):

PR #3223, opened 2024-11-18
Feature/playwright algolia updater

### LINKED ISSUE #3079: Migrate e2e/unit/AlgoliaUpdater.cy to playwright

clari182:
- closes #3079


site/gatsby-site/playwright/e2e/unit/AlgoliaUpdater.spec.ts
kepae, line 1: Drive-by comment: why use a stub of the mongo database if instead we can use the in-memory mongo instance? and therefore put this in the e2e-full tests?

site/gatsby-site/playwright/config.ts
```diff
@@ -11,6 +11,7 @@ const config: ConfigType = {
     E2E_ADMIN_USERNAME: process.env.E2E_ADMIN_USERNAME!,
     IS_EMPTY_ENVIRONMENT: process.env.IS_EMPTY_ENVIRONMENT ?? '',
     AVAILABLE_LANGUAGES: process.env.GATSBY_AVAILABLE_LANGUAGES ?? '',
+    MONGODB_CONNECTION_STRING: process.env.MONGODB_CONNECTION_STRING ?? '',
```
kepae, line 14: please add this as an optional type to the `ConfigType` interface above, like `AVAILABLE_LANGUAGES`. (and see with below comment -- this might be a good change to keep if its documenting an env variable that's being used)

site/gatsby-site/playwright/e2e-full/unit/AlgoliaUpdater.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+import config from '../../config';
+import { init } from '../../memory-mongo';
+
+test('Should update translations to Algolia', async ({ page }) => {
+
+  init();
+
+  const reporter = { log: sinon.stub() };
+
+  const mongoClient = new MongoClient(config.MONGODB_CONNECTION_STRING);
```
kepae, line 16: hmm, I'm not sure this is how we want to "connect" to the in-memory MongoDB instance. I suspect that the test fixture should expose a client.

The client connect logic for the in-memory instance is handled here, and we probably shouldn't let different connection logic leak:
https://github.com/responsible-ai-collaborative/aiid/blob/4b5603012c380a6beac122717d4b6b48b2df5218/site/gatsby-site/playwright/memory-mongo.ts#L85

@cesarvarela should we expose a client in the test fixture?

clari182: @cesarvarela I went back with using the mocks, but I didn't use jest and moved to `/server` as there wasn't an easy way to stub the AlgoliaUpdater with it.

kepae (approved):

PR #3258, opened 2024-12-03
Add gtag id only in prod

### LINKED ISSUE #3015: Refactor Google analytics tag tracking to separate environments
Currently, we use [gatsby-plugin-google-gtag](https://www.gatsbyjs.com/plugins/gatsby-plugin-google-gtag/) to configure our site with the given Google Analytics tag.

At the moment, this always ends up instrumenting the site in all environments (local, CI/CD, staging, production) with the same tag. To identify production traffic in the analytics we have to filter by host name to incidentdatabase.ai -- but this restricts access to some analytics features or makes them harder to use.

We should refactor so that either the tag is only present in production environments.
Alternatively, if there are better ways to separate environments for the sake of analytics, we can do that.

https://github.com/responsible-ai-collaborative/aiid/blob/f10e79b7f4a2f827366b19a7494a0898c6298c52/site/gatsby-site/gatsby-config.js#L67

clari182:
- closes #3015 

We will only track prod now

### DIFF
--- a/site/gatsby-site/playwright/e2e/unit/gtag.spec.ts
+++ b/site/gatsby-site/playwright/e2e/unit/gtag.spec.ts
@@ -2,40 +2,39 @@ import { expect } from '@playwright/test';
 import config from '../../../config';
 import { test } from '../../utils';
 
-const productionUrl = config.gatsby.siteUrl;
 const googleTrackingId = config.gatsby.gaTrackingId;
 
 test.describe('Google Analytics Tracking', () => {
-  test('Should track gtag only on production', async ({ page }) => {
+  test('Should track gtag only on production', async ({ page, runOnlyInProduction }) => {
     await page.goto('/');
-
-    const currentUrl = page.url();
-
-    const isProduction = currentUrl.startsWith(productionUrl);
 
     const hasGoogleTracking = await page.locator(`script[src*="https://www.googletagmanager.com/gtag/js?id=${googleTrackingId}"]`).count();
 
-    if (isProduction) {
-      expect(hasGoogleTracking).toBeGreaterThan(0);
+    expect(hasGoogleTracking).toBeGreaterThan(0);
 
-      const [trackingRequest] = await Promise.all([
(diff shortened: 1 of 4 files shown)


site/gatsby-site/playwright/e2e/unit/gtag.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+  test('Should track gtag only on production', async ({ page }) => {
+    await page.goto('/');
+
+    const currentUrl = page.url();
+
+    const isProduction = currentUrl.startsWith(productionUrl);
+
+    const hasGoogleTracking = await page.locator(`script[src*="https://www.googletagmanager.com/gtag/js?id=${googleTrackingId}"]`).count();
+
+    if (isProduction) {
```
pdcp1, line 18: I understand the logic here, but I'm trying to see if we can avoid different test flows in different environments.
For example, this `if` condition will evaluate to true in the `Deploy to production` workflow, but it will evaluate to false in all other environments.

site/gatsby-site/playwright/e2e/unit/gtag.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+const productionUrl = config.gatsby.siteUrl;
+const googleTrackingId = config.gatsby.gaTrackingId;
+
+test.describe('Google Analytics Tracking', () => {
+  test('Should track gtag only on production', async ({ page }) => {
+    await page.goto('/');
+
+    const currentUrl = page.url();
+
+    const isProduction = currentUrl.startsWith(productionUrl);
```
pdcp1, line 14: Are we sure that the `currentUrl` in production will be `https://incidentdatabase.ai` in the testing environment? If not, this test should be changed.

pdcp1: Note: we can confirm that the `currentUrl` during the test run is not `https://incidentdatabase.ai`, it's `localhost`, so this logic should be changed.

pdcp1 (approved):

Issue #3268, opened 2024-12-06
Translation process and localization improvements 


pdcp1:
### 1. Env variable
`GATSBY_AVAILABLE_LANGUAGES=en,es,fr,ja`

### 2. i18n config file
`site/gatsby-site/i18n/config.json`

### 3. UI translations
JSON files with the translations under the folder `site/gatsby-site/i18n/locales/[language code]`. E.g: `site/gatsby-site/i18n/locales/ja/account.json`

#### Improvements
- The [meta i18n page](https://incidentdatabase.ai/meta/i18n/) helps to identify pending translations
- We should implement fallback to English translations when a new text lacks a translation.

### 4. Site language switcher
Change `site/gatsby-site/src/components/i18n/LanguageSwitcher.js` code adding the new language

### 5. Atlas rules
Add the Atlas rules file for the `translations/reports_[language code]` collection.
#### Improvements
- We are going to leave Atlas soon ✅ 

### 6. Gatsby type definition
Add Gatsby type definition to this file: `site/gatsby-site/typeDefs.js`

### 7. Sorting option on Discover page
Add Sorting option to Discover page for each field: `site/gatsby-site/src/components/discover/SORTING_LISTS.js`

#### Improvements
- We should add these fields programmatically (e.g.: `value_ja: 'instant_search-ja_epoch_incident_date_asc'`) iterating on the `GATSBY_AVAILABLE_LANGUAGES` language list
  - https://github.com/responsible-ai-collaborative/aiid/pull/3271

### 8. Report translation MongoDB query
Add report translation query on `site/gatsby-site/src/graphql/reports.js`

#### Improvements
- Change the way we store the translations. Using just one `reports` collection with a new document field `language`.
  - https://github.com/responsible-ai-collaborative/aiid/issues/3289
  - https://github.com/responsible-ai-collaborative/aiid/pull/3302
- Change the API to use this new DB approach
  - #3315 
- Change the client graphql mutations 
  - #3315 

### 9. Blog posts translations
Not mandatory since we have translations fallback to English

### 10. Gatsby graphql queries
For example, in the `landingPage.js` file, we have `latestIncidentsReports_[language code]` query as part of the Gatsby page query.
- https://github.com/responsible-ai-collaborative/aiid/pull/3303

#### Improvements
- Remove all language dependency from Gatsby page queries, not just from the landing page.

PR #3271, opened 2024-12-09
[Translations] 4 - Refactor Discover sorting options to use a dynamic language list

pdcp1:

Also part of the language improvements in https://github.com/responsible-ai-collaborative/aiid/issues/3268

On top of:
- https://github.com/responsible-ai-collaborative/aiid/pull/3315

### Summary

The objective is to prevent further hardcoded code in the `components/discover/SORTING_LISTS.js` whenever a new language is introduced to the site.
The sorting fields are now generated dynamically based on the available languages.


site/gatsby-site/src/components/discover/SORTING_LISTS.js
```diff
@@ -1,93 +1,86 @@
...
-    value: 'instant_search-en-featured',
-    value_en: 'instant_search-en-featured',
-    value_es: 'instant_search-es-featured',
-    value_fr: 'instant_search-fr-featured',
-    value_ja: 'instant_search-ja-featured',
     label: 'Relevance',
     faIcon: faCalendarAlt,
     faClasses: 'far fa-calendar-alt',
     division: true,
+    instantKey: 'featured',
```
cesarvarela, line 12: Consider adding a comment to explain how you arrive at this value.

site/gatsby-site/src/components/discover/SORTING_LISTS.js
```diff
@@ -1,93 +1,86 @@
...
     faIcon: faCalendarAlt,
     faClasses: 'far fa-calendar-alt',
     type: 'date',
+    instantKey: 'epoch_date_submitted_asc',
   },
 ];
 
+const SORTING_LIST = [];
+
+BASE_SORTING_LIST.forEach((item) => {
```
cesarvarela, line 68: nit. Maybe a `.map` is easier to understand because it seems you are going from  `BASE_SORTINF_LIST` to  `SORTING_LIST`?

cesarvarela (commented):
Is there already a test covering this?

site/gatsby-site/src/components/discover/SORTING_LISTS.js
pdcp1: done

pdcp1:
> Is there already a test covering this?

@cesarvarela Yes, we test the sorting feature in the [discover.spec.ts](https://github.com/responsible-ai-collaborative/aiid/blob/main/site/gatsby-site/playwright/e2e/discover.spec.ts)

site/gatsby-site/src/components/discover/SORTING_LISTS.js
```diff
@@ -1,93 +1,87 @@
...
     faClasses: 'far fa-calendar-alt',
     type: 'date',
+    instantKey: 'epoch_date_submitted_asc',
   },
 ];
 
+const SORTING_LIST = BASE_SORTING_LIST.map((item) => {
+  const newItem = {
+    ...item,
+    value: `instant_search-en${item.instantKey == 'featured' ? '-' : '_'}${item.instantKey}`,
```
kepae, line 72: I guess this is how the index is queried in the code, but the fact that "featured" is prefixed with a hyphen and others are not is... subtle. If that's correct, I would just add a quick comment here.

site/gatsby-site/src/components/discover/SORTING_LISTS.js
```diff
@@ -1,93 +1,87 @@
 import { faCalendarAlt } from '@fortawesome/free-solid-svg-icons';
+import config from '../../../config';
 
-const SORTING_LIST = [
+/**
+ * @property {string} instantKey - This field is used to generate the instant search key for each language
```
kepae, line 5:
```suggestion
 * @property {string} instantKey - This auxiliary field is used to generate the instant search key for each language, and should be removed before creating the final sorting list.
```

kepae (approved):

site/gatsby-site/src/components/discover/SORTING_LISTS.js
pdcp1: done

kepae (approved):

cesarvarela (changes requested):
I think there should be at least one test changed here 🤔

pdcp1:
> I think there should be at least one test changed here 🤔

@cesarvarela I think we discussed this in the past [here](https://github.com/responsible-ai-collaborative/aiid/pull/3271#issuecomment-2540027845).
We already have a test for this feature, the idea is this chance doesn't break it.

Issue #2945, opened 2024-07-02
Fix skipped Playwright tests


pdcp1:
As part of https://github.com/responsible-ai-collaborative/aiid/issues/2940 and https://github.com/responsible-ai-collaborative/aiid/pull/2943 we skip all failing Playwright tests.

We should fix them and remove the `skip`

PR #3275, opened 2024-12-10
Fix/related by url

clari182:
- closes #3187

PR #3287, opened 2024-12-16
Fix/unskip playwright tests 2.0

clari182:
- part of #2945


site/gatsby-site/playwright/e2e-full/submit.spec.ts
```diff
@@ -625,99 +625,6 @@ test.describe('The Submit form', () => {
   }
   );
 
-    test.skip('Should show a preliminary checks message', async ({ page }) => {
```
pdcp1, line 628: Do we have another test that substitutes for this one?

clari182: Yes, `Should *not* show a list of related reports` which was added in PR https://github.com/responsible-ai-collaborative/aiid/pull/3275

pdcp1 (approved):
LGTM 👍

PR #3292, opened 2024-12-17
Email templates – Fallback to sans-serif, reduce horizontal padding

lmcnulty:
Was previously falling back to a serif font when it was the default, and it had enough padding to look weird on mobile.


kepae (approved):
(I'll be looking for the staging test email)

PR #3294, opened 2024-12-18
Migrate created_at related triggers to the API

cesarvarela:


clari182 (changes requested):
New incidents (from submissions) are not getting a `created_at` field. I think it's because of the `promoteSubmissionToReport` in `server/fields/submissions.ts` and it should be added manually in the `insertOne`s used there.

I think the same will happen for all `insertOne` written inside another mutation

cesarvarela:
Thanks, I also exposed the field in the schema. And yes, we'll have to track where insertions happen and ensure the `created_at` field is added.

clari182:
@cesarvarela I think the same issue is happening for `createVariant` and it's not adding the `created_at`

clari182 (approved):

PR #3295, opened 2024-12-18
Delete default admin user mutation and plugin

cesarvarela:
With the new Auth incoming, we should rethink our strategy. Migrating this method as it is is not worth it.


kepae:
+1. In a world where the only dependency is the DB itself and not Atlas features for user management, the answer might be a simple command-line script that elevates a given user by ID to admin.

kepae (approved):

PR #3296, opened 2024-12-18
Delete Atlas Graphql

cesarvarela:
This PR deletes the Atlas graphql config files and other related features. 

It should be safe to ship this alongside the current #3344. 

note: we can't delete everything in Atlas yet, because we need some of Atlas still working so the users migration is able to pull users from Atlas.


pdcp1 (approved):
LGTM. Just one question.
Are the `logReportHistory`, `logIncidentHistory`, and `riskResolver` functions migrated to another location? I don't see them included in this PR.

kepae (approved):

cesarvarela:
> LGTM. Just one question. Are the `logReportHistory`, `logIncidentHistory`, and `riskResolver` functions migrated to another location? I don't see them included in this PR.

Yes, they were migrated some time ago.

Issue #3289, opened 2024-12-16
[Localization improvements] MongoDB translation storage changes


pdcp1:
This issue is part of the epic https://github.com/responsible-ai-collaborative/aiid/issues/3268

## Improvements

1. Change the way we store the translations. Using just one `reports` collection with a new document field `language`.
- Collections that will be removed:
  - `translations/reports_en`
  - `translations/reports_es`
  - `translations/reports_fr`
  - `translations/reports_ja`
- Collection that will be added
  - `translations/reports` 
- About the new `translations/reports` collection
  - The collection will include translations of all reports in the system.
  - A new document field named `language` will be introduced to identify the translation language.
  
3. Change the API to use this new DB approach
4. Change the client graphql mutations to use the language code as a parameter

### PRs
- #3302 
- #3303 
- #3315
- #3271
- #3327
- #3476

PR #3302, opened 2024-12-20
[Translations] 1 - New single Reports translation collection

pdcp1:

### Summary

- Adds a migration that copy all `translations.reports_xx` ("en", "es", "fr", and "ja") documents to a new `translations.reports` collection
- Changes the `Translator` script to also insert new report translations into this new `reports` collection


site/gatsby-site/migrations/2024.12.11T14.39.38.refactor-reports-translations.js
```diff
@@ -0,0 +1,77 @@
...
+    language: 'fr',
+  }));
+
+  const transformedReportsJa = reportsJa.map((report) => ({
+    ...report,
+    language: 'ja',
+  }));
+
+  // Insert transformed documents into reports
+  const resultEn = await reportsCollection.insertMany(transformedReportsEn);
```
cesarvarela, line 46: I wonder if this will be too big to insert at once. Have you tried using the production data already?

pdcp1: Yes, I executed this migration on my personal MongoDB instance hosted in Atlas, which is a restore of the Production database, and it works fine 👌.

cesarvarela (commented):
Looks good, my only concern is that this collection will grow huge, if we translate every report it will be the original reports collection * number of languages big.

kepae:
This is true, but I think at our scale it won't matter, right? Aside, the
documents will be smaller, so more updates can be processed with the same
amount of working memory. Lastly, we are more likely to request only one
language from the DB at a time when displaying reports.

pdcp1:
I agree. I don't anticipate an issue with the collection size. It will contain small documents with just 6 fields.
In case we need to improve query performance, we can add an index on `report_number` field

kepae (approved):

pdcp1:
@kepae I think we are good to merge this PR into Staging. Let me know if you prefer to release the current staging changes first, and then this one.

cesarvarela (approved):

kepae:
Let's deploy to staging tonight, then production tomorrow when you are satisfied with the migration. Then we can rapidly stage and deploy the queued (and patiently waiting) changes. :-)

PR #3303, opened 2024-12-23
[Translations] 1 - Gatsby graphql queries refactor

pdcp1:
Part of https://github.com/responsible-ai-collaborative/aiid/issues/3289

This PR:

- Add gatsby-source-mongodb plugin for the new "reports" collection in translations database

PR #3315, opened 2025-01-14
[Translations] 3 - GraphQL API refactoring

pdcp1:
On top of:
- https://github.com/responsible-ai-collaborative/aiid/pull/3303

### Summary
- Change `FindReportWithTranslations` graphql query to accept a list of languages
- Update Report edit page (`/cite/edit.js`) to load translations using the new query
- Update Report edit page (`/cite/edit.js`) to save the reports translations into the new collection `translations/reports`


site/gatsby-site/server/types/report.ts
```diff
@@ -48,22 +50,33 @@ export const ReportType = new GraphQLObjectType({
...
+            
+                return args.languages.map((language: string) => {
+                    const translation = translations.find(t => t.language === language);
+                
+                    return translation ? {
+                        text: translation.text || "",
+                        title: translation.title || "",
+                        plain_text: translation.plain_text || "",
+                        language: language,
+                    } : {
```
kepae, line 73: Is this the best option when there is no translation for a report? I wish we could return a NOT_FOUND type error. But I'm not sure what is best from the GraphQL norm perspective.

site/gatsby-site/src/pages/cite/edit.js
```diff
@@ -92,13 +101,22 @@ function EditCitePage(props) {
...
 
-  const { config } = useLocalization();
+  useEffect(() => {
+    if (reportData?.report?.translations) {
+      const translations = reportData.report.translations.reduce((acc, translation) => {
+        acc[`translations_${translation.language}`] = pick(translation, translationsFields);
+        return acc;
+      }, {});
+
+      setReportTranslations(translations);
```
kepae, line 117: Why are the translations being passed in the old format `translations_LANG` keys? Is this temporary?

cesarvarela (commented):
I agree with kevin's comments, nice to see tests doing its job <3

site/gatsby-site/server/types/report.ts
cesarvarela: I think `null` would be better

pdcp1: I agree. Returning `null` in GraphQL when a translation is not found is considered a good practice. It's a common GraphQL pattern that indicates the absence of data.

site/gatsby-site/src/pages/cite/edit.js
```diff
@@ -92,13 +101,36 @@ function EditCitePage(props) {
...
+      const translations = availableLanguages
+        .map((c) => c.code)
+        .reduce((acc, languageCode) => {
+          // Find existing translation for this language
+          const existingTranslation = reportData.report.translations?.find(
+            (t) => t.language === languageCode
+          );
+
+          // If translation exists use its values, otherwise use empty values
+          acc[`translations_${languageCode}`] = existingTranslation
```
pdcp1, line 121: @kepae This `translations_[LANGUAGE CODE]` format is actually to be used in the `IncidentReportForm.js` form input field names. It is not used in the data base structure, is just for the form data.

site/gatsby-site/server/types/report.ts
```diff
@@ -48,22 +50,28 @@ export const ReportType = new GraphQLObjectType({
...
+                return args.languages.map((language: string) => {
+                    const translation = translations.find(t => t.language === language);
+                    
+                    return translation ? {
+                        text: translation.text,
+                        title: translation.title,
+                        plain_text: translation.plain_text,
+                        language: language,
+                    } : null;
+                }).filter(Boolean);
```
kepae, line 74: Final nitpick: I believe it's better to return the `null` here and return exactly the number of items requested, instead of filtering. The client already has to error check.
(this is different than before, which returned empty translations without status messages, so the empty strings were ambiguous)

kepae (changes requested):
One last nitpick on API behavior then good to merge.

site/gatsby-site/server/types/report.ts
pdcp1: @kepae If we don't filter the null values here, the API will return a result like this for a query of two languages:
```
{
  "data": {
    "reports": [
      {
        "report_number": 2,
        "translations": [
          null,
          null
        ]
      },
      {
        "report_number": 3,
        "translations": [
          null,
          {
            "language": "es",
            "title": "Trabajando cualquier cosa menos de 9 a 5",
            "text": "SAN DIEGO — En una típica carrera de último minuto, Jannette Navarro"
          }
        ]
      },
...
```

There is a second option:
```

  "data": {
    "reports": [
      {
        "report_number": 2,
        "translations": [
          {
            "language": "en",
            "title": null,
            "text": null
          },
          {
            "language": "es",
            "title": null,
            "text": null
          }
        ]
      },
      {
        "report_number": 3,
        "translations": [
          {
            "language": "en",
            "title": null,
            "text": null
          },
          {
            "language": "es",
            "title": "Trabajando cualquier cosa menos de 9 a 5",
            "text": "SAN DIEGO — En una típica carrera de último minuto, Jannette Navarro"
          }
        ]
      },
...
```

I personally prefer the second option, since it allows the client to see which language doesn't have translations.
Let me know which one do you prefer.

kepae: I agree with you on the second option being better, let's do that.

(Sigh. Null handling...)

pdcp1: done!

cesarvarela (approved):

PR #3316, opened 2025-01-14
Add `call-test-playwright-full` step dependency to `call-deploy`

pdcp1:
This PR avoids Staging deploys if a Playwright test fails.

Example run:
https://github.com/pdcp1/aiid/actions/runs/12776870045


kepae (approved):
I see now that this matches production. Thanks!

PR #3329, opened 2025-01-27
Feature/tags split

### LINKED ISSUE #3330: Should allow commas in new entities name (from submission form)
Currently, new developers, deployers, harmed parties and implicated systems inputs on the submission form, are treated as tags. We should allow the name to contain commas. Right now, a comma acts as `Enter`.

clari182:
- closes #3330


site/gatsby-site/src/components/forms/Tags.js
```diff
@@ -14,13 +14,20 @@ export default function Tags({
   className,
   allowNew = true,
   stayOpen = false,
+  splitChar = ',',
```
pdcp1, line 17: I think we have a type discrepancy in the field `splitChar`. Here we have a String default value `,` but in the `StepThree.js` file I see that you use a Boolean value.
I suggest to use always String type, and pass a `null` value instead of `false` for these fields.

pdcp1 (changes requested):

pdcp1 (approved):
LGTM 👍

PR #3346, opened 2025-02-01
Use single email Mailersend endpoint for auth emails

cesarvarela:
This uses Mailersend's single email endpoint for auth-related emails (sign-up, sign-in) and the bulk endpoint for notifications (new incident, new report, etc.)

It will help avoid hitting the rate limit of 15 calls per minute of the bulk endpoint, and it is how Mailersend's api expects to be used.

https://pr-3346--staging-aiid.netlify.app


site/gatsby-site/server/emails/index.ts
```diff
@@ -135,3 +135,66 @@ export const sendEmail = async ({ recipients, subject, dynamicData, templateId }
...
+        apiKey: config.MAILERSEND_API_KEY,
+    });
+
+    assert(email.to.length == 1, 'Email must have exactly one recipient');
+    assert(!email.cc, 'Should not use the "cc" field');
+
+    await mailersend.email.send(email);
+}
+
+export interface SendEmailParams {
```
kepae, line 151: Not in this PR but a next one: we could use the official MailerSend Node API. Please create an issue for this as a potential improvement.
https://github.com/mailersend/mailersend-nodejs?tab=readme-ov-file#send-an-email

kepae (approved):
Please confirm that the test coverage linting is only reporting missing tests because it's not instrumenting sendEmailTest.js?

site/gatsby-site/server/emails/index.ts
kepae: Ah, apologies, I misread the imports and didn't see the original types being imported right away. Don't do code reviews on mobile, kids.

Issue #3320, opened 2025-01-17
Make incident title a hyperlink in notification emails


kepae:
In the below email, I expected the incident title to be a clickable hyperlink. I think it would be nice if it were.

<img width="828" alt="Image" src="https://github.com/user-attachments/assets/8a43330e-a886-4528-b3eb-8069605c0d80" />

Issue #3321, opened 2025-01-21
Notification emails rendering issue on iOS Mail app


cesarvarela:
There are some overlaps, and we might want the entity tags to not break into two lines.

![Image](https://github.com/user-attachments/assets/45dfd1cc-96f5-45ff-9554-6d351981dcfe)

iOS 18.1
default Mail app

PR #3349, opened 2025-02-04
Email templates fixes

lmcnulty:
Resolves https://github.com/responsible-ai-collaborative/aiid/issues/3320.

Testing with a webkit browser, deleting the line setting an explicit height for the subtitle fixes the issue where it overlaps the email content from https://github.com/responsible-ai-collaborative/aiid/issues/3321. I hoping that explicitly setting a height and width for the image will fix the issue where it overlaps with other elements, but I'm not sure if it will – I don't have an easy way to test it on iOS.


clari182 (approved):
Looks good! I was also able to test the changes in IOS and it's not overlapping anymore. 
I tried the previous template and got an overlap, then applied your changes and no overlap!

PR #3356, opened 2025-02-07
Fix/cite tags placement

### LINKED ISSUE #3150: Inconsistent alignment of "Responded" badge in cite page header
The "responded" is vertically centered but the other items are not.

image

clari182:
- closes #3150 

Solution: 
Improve cite's header layout to remove unecessary wrapping of tags or social media icons.

Test URL for a responded incident:
https://pr-3356--staging-aiid.netlify.app/cite/321


lmcnulty:
So the design idea here was to align all of the header items with the Discover/Submit buttons when we have the space for it.

image

This is how it works on most pages on the site:

image

Putting the tags on their own line kind of throws off the balance.

image

I'd say we should either

1. Put all the tags and buttons below the title and remove the grey line below them so it doesn't look like they should be centered  *or*
2. Keep them all centered on one line, wrapping only when there's not enough space.

I think we should ask what other folks think on this one.

clari182:
Hey @lmcnulty when you get a chance can you review this PR?

pdcp1 (approved):

Issue #3368, opened 2025-02-13
Empty report texts on latest reports in the homepage


pdcp1:
![Image](https://github.com/user-attachments/assets/9f0b8aaf-106a-41ee-b139-b00ae435500b)

We found that there are some empty translations stored in the `translations/reports_xx` collections.
We should investigate the reason for these empty values and a fix for it.

![Image](https://github.com/user-attachments/assets/88bc0304-1fba-4f67-9068-c339ea169d23)

PR #3369, opened 2025-02-13
Delete all empty translations in all languages

pdcp1:
This PR deletes all empty translations in the `translations/reports_xx` collections.

Part of the fix for https://github.com/responsible-ai-collaborative/aiid/issues/3368

image


pdcp1:
> Can we also add logic to prevent empty translations from being inserted? otherwise this can happen again. Not sure how they landed in the first place.

@kepae I'm currently working on a permanent solution to avoid this issue in the future. I plan to create a new PR with this fix, but if you prefer, I can include it in this one.

kepae:
> > Can we also add logic to prevent empty translations from being inserted? otherwise this can happen again. Not sure how they landed in the first place.
> 
> @kepae I'm currently working on a permanent solution to avoid this issue in the future. I plan to create a new PR with this fix, but if you prefer, I can include it in this one.

Gotcha. Let's merge this now and clean up the data, then fix for the future.

kepae (approved):

PR #3374, opened 2025-02-14
Avoid inserting empty report translation if the Google's translate API returns empty translations

pdcp1:
This is the permanent fix for https://github.com/responsible-ai-collaborative/aiid/issues/3368

### DIFF
--- a/site/gatsby-site/src/scripts/Translator.js
+++ b/site/gatsby-site/src/scripts/Translator.js
@@ -137,6 +137,16 @@ class Translator {
     for (let i = 0; i < results.length; i++) {
       const result = results[i];
 
+      // Check if the translation is empty. The translation API sometimes returns empty strings with a 200 status code.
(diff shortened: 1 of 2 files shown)


site/gatsby-site/src/scripts/Translator.js
```diff
@@ -190,7 +197,10 @@ class Translator {
 
       const items = reports.filter((r) => r.language !== to);
 
-      const translated = await this.translateReportsCollection({ items, to });
+      let translated = await this.translateReportsCollection({ items, to });
+
+      // filter translated reports that are not empty
```
kepae, line 202:
```suggestion
      // filter translated reports that are empty
```

kepae: (I always have to think about this for a few seconds :-))

kepae (approved):
Thanks!

PR #3411, opened 2025-02-25
Feature import classifications

cesarvarela:
## Staging instructions

Run the command:

```
npm run import-classifications-csv -- --csvFile=data/MIT-Incident-classifications.csv --namespace=MIT --graphqlEndpoint=https://staging-aiid.netlify.app/api/graphql --sessionToken=521260ec-3399-4869-a602-a9a2c62044b3 --mappingFile=data/mit-mapping.json
```
Parameter explanation:

- csvFile: source .csv file
- mappingFile: file that maps between .csv fields and values and our MIT taxonomy field and values (there were some inconsistencies  here and there)
- sessionToken: session token of a user with write permissions
- namespace: namespace under which the classifications will be imported

--notes

To get the session token, you need to log in to the web app and fetch the cookie from a request to the `graphql` endpoint:
image

To generate a magic link with `admin` roles (necessary to upsert classifications):
`npm run magic-link -- --email test@something.com --callbackUrl / --roles admin`

Result:

image

Next possible steps:

- [x] - Add an option to map CSV columns to classification fields
- [x] - Add permitted values validation


site/gatsby-site/src/scripts/import-classifications-csv.ts
```diff
@@ -0,0 +1,724 @@
...
+
+/**
+ * Convert value based on MongoDB type
+ */
+function convertValueByMongoType(value: string, mongoType: string): any {
+  switch (mongoType) {
+    case 'string':
+      return value.trim();
+    default:
+      throw new Error(`Unsupported mongo_type: ${mongoType}. Supported types are: string, number, boolean, date`);
```
lmcnulty, line 282: Why does this say that string, number, and boolean are supported when the only case handled here is string?

cesarvarela: Yeah I starteed with more data types but realized we only need `string` for now(); Fixed.

lmcnulty (approved):
This looks fine. It does seem a bit long for what we're doing with it... I'm not sure that we need fancy logging, timing, a bunch of cli switches, etc. It sounded like you were eager to try out some LLMs on this, which is fine, but I think we should be careful not to generate more code than we can review or maintain.

cesarvarela:
@lmcnulty Yeah, I'm testing the waters to see what is possible. The code has a machine-generated smell that I'm not a fan of, but at the same time, these new LLMs are so fast and efficient that we can't ignore them. There is a middle ground somewhere...

PR #3413, opened 2025-02-26
Add MIT taxonomy

lmcnulty:
- Adds a migration to add the MIT taxa object with full field descriptions
- Adds the MIT taxonomy to the list on /taxonomies.
- I'm giving the taxa object a field `automatedClassifications: true` so that we can show it in the UI when required. Ultimately this should be on the classifications level, but I think this will be the fastest way to have it done this week.


kepae (commented):
This change, plus some disclaimer as we mentioned that this taxonomy is in the process of being imported based on Simon's work, and pointer to the original blog source.

kepae (approved):

Issue #3412, opened 2025-02-26
Discover button has a different shade of blue


cesarvarela:
<img width="478" alt="Image" src="https://github.com/user-attachments/assets/155d0751-b8d2-4ce7-bf05-05f9d4be3c7f" />

cesarvarela:
There is a yellowish tint in the background of some components too:
<img width="470" alt="Image" src="https://github.com/user-attachments/assets/b1bf4342-d7da-4fc9-b5d6-b50a0ac961f2" 
/>

kepae:
And other components are affecteD: checkboxes, tags in submission queue... 

<img width="523" alt="Image" src="https://github.com/user-attachments/assets/017aecda-5336-417b-be41-42e182b6e963" />

But not entity tags in cite page?

<img width="930" alt="Image" src="https://github.com/user-attachments/assets/77070723-fbbd-4be4-a7ab-70b6d2968327" />

lmcnulty:
It changed from `bg-blue-xxx` to `bg-cyan-xxx`.

![Image](https://github.com/user-attachments/assets/97fa34d0-fcb9-4ee8-b915-d43e95f59988)

`cyan` appears exactly zero places in our frontend code.

![Image](https://github.com/user-attachments/assets/6286aa5b-0914-4f51-ab25-22f7f6bf63c9)

It looks like this is because the landing page imports directly from `react-flowbite`.

<img src="https://github.com/user-attachments/assets/4b0beca3-2e53-45f7-b3ad-ef3083580ca0" width="500"/>

Whereas the entities badges on the citation page just include the styles directly.

<img src="https://github.com/user-attachments/assets/2f6fda09-b9ed-4479-b7d0-bb32373736a7" width="600"/>

So we must have updated `react-flowbite` and to a new version that uses cyan instead of blue.

PR #3446, opened 2025-03-11
Replace cyan with blue in Tailwind config

lmcnulty:
It turns out that solving https://github.com/responsible-ai-collaborative/aiid/issues/3412 properly is surprisingly complicated. Fortunately, we don't use cyan anywhere, so we can just redefine it as blue in our Tailwind config.

image

image

### DIFF
--- a/site/gatsby-site/tailwind.config.js
+++ b/site/gatsby-site/tailwind.config.js
@@ -115,6 +115,7 @@ module.exports = {
         // in a way that's not easy to reverse.
         // We don't use cyan anywhere,
         // so the easiest fix is to redefine cyan as blue.
+        // Issue: https://github.com/responsible-ai-collaborative/aiid/issues/3412
         'cyan-50': 'oklch(0.97 0.014 254.604)',
         'cyan-100': 'oklch(0.932 0.032 255.585)',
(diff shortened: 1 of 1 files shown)


site/gatsby-site/tailwind.config.js
```diff
@@ -110,6 +110,35 @@ module.exports = {
         'table-text': 'rgba(0,0,0,0.05)',
         'btn-check-active': '#565e64',
         'btn-check-border': '#51585e',
+
+        // flowbite-react replaced blue with cyan
+        // in a way that's not easy to reverse.
+        // We don't use cyan anywhere,
+        // so the easiest fix is to redefine cyan as blue.
```
kepae, line 117:
```suggestion
        // flowbite-react replaced blue with cyan
        // in a way that's not easy to reverse.
        // We don't use cyan anywhere,
        // so the easiest fix is to redefine cyan as blue.
        // Issue: https://github.com/responsible-ai-collaborative/aiid/issues/3412
```

kepae: Inline links save time from investigating git logs. :-)

kepae (approved):
Nice save! See if there's a flowbite repo issue to stir up or add a +1?

Issue #3439, opened 2025-03-06
incident Stats card  lists taxonomies not visible


cesarvarela:
GMF is not being displayed here, but shown in the list of Applied Taxonomies:

<img width="1122" alt="Image" src="https://github.com/user-attachments/assets/cd677206-20f9-48c8-a4f0-cff3a81abdd9" />

PR #3451, opened 2025-03-12
Fix taxonomies with unpublished classifications showing in incident stats card

lmcnulty:
Resolves https://github.com/responsible-ai-collaborative/aiid/issues/3439


cesarvarela (changes requested):
Please add or update an existing test!

PR #3469, opened 2025-03-24
Update script to only push specific incidents

cesarvarela:


clari182 (approved):
Code looks good, works as expected

PR #3470, opened 2025-03-25
Extract algolia step to workflow step

cesarvarela:
This extracts the current Algolia push process from the Gatsby node to a new step in the deploy workflow and also adds some improvements to the subset environment variable, which allows us to push the Discover index to a free Algolia account.


.github/workflows/deploy.yml
```diff
@@ -134,6 +134,18 @@ jobs:
...
+      - name: Update Algolia indexes
+        if: inputs.netlify-context == 'production' && inputs.netlify-alias == ''
+        run: npm run algolia-update:ci
+        working-directory: site/gatsby-site
+        env:
+          GATSBY_ALGOLIA_APP_ID: ${{ vars.GATSBY_ALGOLIA_APP_ID }}
+          ALGOLIA_ADMIN_KEY: ${{ secrets.ALGOLIA_ADMIN_KEY }}
+          GATSBY_AVAILABLE_LANGUAGES: ${{ vars.GATSBY_AVAILABLE_LANGUAGES }}
+          MONGODB_TRANSLATIONS_CONNECTION_STRING: ${{ secrets.MONGODB_TRANSLATIONS_CONNECTION_STRING }}
+          GATSBY_SITE_URL: ${{ inputs.site-url || vars.SITE_URL }}
```
pdcp1, line 146: Is this `GATSBY_SITE_URL` variable needed for the Algolia update process?

pdcp1 (commented):
@cesarvarela Do you have a working run to take a look?

.github/workflows/deploy.yml
cesarvarela: It is because the config file checks for every variable to be defined, we could set it to dummy though

pdcp1: Got it. Could you set this to `dummy`?

pdcp1 (approved):

pdcp1:
@cesarvarela please resolve conflicts before merge it to Staging

PR #3484, opened 2025-03-31
Fix blocked navigation between pages on submission form

lmcnulty:


pdcp1 (changes requested):
Please add tests for this new behaviour.

Also, it will be better if the required fields are marked in red when the submission fails with _"Was not able to create the report, please review the form and try again."_

### Steps to reproduce it:
1. https://pr-3484--staging-aiid.netlify.app/apps/submit
2. Click on `Add more info`
3. Click on `Submit` and confirm
4. A `Was not able to create the report, please review the form and try again.` toast is displayed and redirected to Step 1
5. The required fields should be marked in red

PR #3506, opened 2025-04-11
Embed button implementation only

cesarvarela:
Tests being added in https://github.com/responsible-ai-collaborative/aiid/pull/3507


site/gatsby-site/src/utils/LookupIndex.js
```diff
@@ -1,13 +1,42 @@
...
+
+    return parsedURL.host + parsedURL.pathname;
+  } catch (e) {
+    return url;
+  }
+}
+
 class LookupIndex {
-  constructor({ client, filePath }) {
+  constructor({ client, filePath, optimized = false }) {
```
kepae, line 32: Please document this class, especially if there is no documentation in-line or for the helpers. This lookup and the optimization are -- as many good optimizations -- clever.

site/gatsby-site/static/embed.js
```diff
@@ -0,0 +1,155 @@
...
+  function isValidURL(str) {
+    try {
+      new URL(str);
+      return true;
+    } catch {
+      return false;
+    }
+  }
+
+  function hashString(str) {
```
kepae, line 41: This is the catch with using a custom hash: If this function served via the static JS or the other identical function used to generate the map changes, we break all links. Please somehow leave a comment indicating this in both places, since we likely can't/don't want to import anything here.

site/gatsby-site/static/embed.js
```diff
@@ -0,0 +1,155 @@
...
+
+      const buttonContainer = document.createElement('div');
+
+      buttonContainer.style.cssText = styles.buttonContainer;
+
+      for (const id of incidentIds) {
+        const link = document.createElement('a');
+
+        link.href = `${baseUrl}/cite/${id}`;
+        link.textContent = `See it on the AIID #${id}`;
```
kepae, line 130: The user should be able to override this/define this text. Follow-up PR if you like?

kepae (changes requested):

kepae (approved):
thanks!

kepae:
One last thought -- the serving URL of this script. Any pros/cons of using the root domain? Should we consider any of:

`/static/embed.min.js`
`/assets/embed.min.js`
`/js/embed.min.js`

PR #3516, opened 2025-04-14
Fix test for right sidebar "Contents" scroll-to section on click on prismic blog post

pdcp1:
This Pr fixes https://github.com/responsible-ai-collaborative/aiid/actions/runs/14448446064/job/40518611426#step:8:234

### DIFF
--- a/site/gatsby-site/playwright/probes/navigation.spec.ts
+++ b/site/gatsby-site/playwright/probes/navigation.spec.ts
@@ -79,7 +79,7 @@ test.describe('Navigation', () => {
       await expect(async () => {
         const subject = await page.locator('h2:has-text("🗄 Trending in the AIID")');
         const boundingBox = await subject.boundingBox();
-        expect(boundingBox?.y).toBeCloseTo(0, 30);
+        expect(boundingBox?.y).toBeCloseTo(0, 0.20);
       }).toPass();
     }
   });


site/gatsby-site/playwright/probes/navigation.spec.ts
```diff
@@ -79,7 +79,7 @@ test.describe('Navigation', () => {
       await expect(async () => {
         const subject = await page.locator('h2:has-text("🗄 Trending in the AIID")');
         const boundingBox = await subject.boundingBox();
-        expect(boundingBox?.y).toBeCloseTo(0, 30);
+        expect(boundingBox?.y).toBeCloseTo(0, 0.20);
```
kepae, line 82: drive-by: why 0.20? IIUC, this should be the number of digits after a decimal, and therefore a natural number.
30 might have been way too precise for comparing pixels.
https://playwright.dev/docs/api/class-genericassertions#generic-assertions-to-be-close-to

pdcp1: Yes, exactly. For example, if we use `toBeCloseTo(0, 5)`, the number should have a difference of less than `0.000005`. It's too precise for comparing pixel counts.

kepae (approved):

PR #3518, opened 2025-04-15
Add classifications to Incident type

cesarvarela:
This allows us to query Incidents' Classifications like this:

```
{
  incident {
    classifications {
      namespace
      attributes {
        value_json
        short_name
      }
    }
  }
}
```


pdcp1 (commented):
The code looks good. Should we add a test for this new classifications Incident type?

cesarvarela:
@pdcp1 it is there in the updated Incidents test fixture.

pdcp1 (approved):

PR #3519, opened 2025-04-15
[HOTFIX] Delete all incident translations. All of them are incorrect in production

pdcp1:
We should merge this into `main` to fix the data and unblock the scheduled deploy.


cesarvarela (approved):
Nice

PR #3528, opened 2025-04-18
Feature/submit issue incident

### LINKED ISSUE #3501: Option to choose "Issue" or "Incident" in the submit form
Right now, the submit form always acts as if we are entering a new Incident. It would be helpful to add a `toggle` option to hide `Incident datas` when Issue is selected. 
More info here: 

> Right now, when someone goes to + Submit and opens the New Incident Report form, I am wondering if we should have the ability to toggle the form at that stage between Issue/Incident. I think it should default to Incident, but if someone switches to Issue, then it could remove all of the unnecessary fields. Currently, someone could theoretically fill out all the details for an incident, but if I review it and determine it is an issue, all of that information would be lost. 

https://raicollab.slack.com/archives/C03L4FLEL2Y/p1744214611126409?thread_ts=1740689920.100029&cid=C03L4FLEL2Y

clari182:
- closes #3501 

https://pr-3528--staging-aiid.netlify.app/apps/submit/

image

Defaults to "Incident". When "Issue" is selected step 3 is not needed


pdcp1 (changes requested):
The code looks good.
I found some missing translations on the "Report type" component:
https://pr-3528--staging-aiid.netlify.app/es/apps/submit/
image

pdcp1 (approved):

PR #3541, opened 2025-04-29
Create 2025.04.25T16.31.55.migrate-incident-88.js

clari182:


site/gatsby-site/migrations/2025.04.25T16.31.55.migrate-incident-88.js
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    // Get reports associated with incident 88
+    const incident88Reports = incident88.reports;
+
+    // Find the last incident to determine the new incident_id
+    const lastIncident = await incidents.find({}).sort({ incident_id: -1 }).limit(1).next();
+
+    // Calculate new incident_id
+    const incident_id = lastIncident ? lastIncident.incident_id + 1 : 1;
```
pdcp1, line 30: I think we should throw an error and cancel the migration if the `lastIncident` is not defined instead of set the `incident_id` as number `1`

pdcp1 (changes requested):

pdcp1 (changes requested):

cesarvarela (commented):
Shouldn't we use the duplicate incident tool instead?

pdcp1 (approved):

PR #3572, opened 2025-05-14
Fix modal scroll

### LINKED ISSUE #3571: Can't scroll page after some modals are closed
And this is the reason: https://github.com/themesberg/flowbite-react/issues/829

For example, it happens with the Citation Info button. And I found it was happening with the modal in the upcoming #3537

cesarvarela:
closes #3571


clari182 (commented):
Should we add some tests for this and check that the body doesn't have overflow:hidden after closing modals?

cesarvarela:
@clari182  tests added!

PR #3599, opened 2025-06-03
Add AI Briefing subscription toggle for admins and editors

### LINKED ISSUE #3598: Enable briefing subscription for editors and admin

clari182:
- closes #3598


site/gatsby-site/src/components/UserSubscriptions.js
```diff
@@ -28,9 +28,15 @@ const UserSubscriptions = () => {
...
-  const { data, loading } = useQuery(FIND_USER_SUBSCRIPTIONS, {
-    variables: { filter: { userId: { EQ: user.id } } },
-  });
+  const { data, loading } = useQuery(
+    FIND_USER_SUBSCRIPTIONS,
+    {
+      variables: { filter: { userId: { EQ: user.id } } },
+    },
+    {
+      fetchPolicy: 'network-only',
```
cesarvarela, line 37: why this?

cesarvarela (commented):
One question before approving

site/gatsby-site/src/components/UserSubscriptions.js
clari182: I found a bug, that if you go to your account - toggle the subscription to receive new incidents email, leave the page and go back to account, the toggle is not updated as the new info is not fetch

cesarvarela: oh thats because the apollo cache doesn't know how to update the subscriptions schema:

https://github.com/responsible-ai-collaborative/aiid/blob/main/site/gatsby-site/src/contexts/UserContext.tsx#L37

Maybe you can give that a try, that's the "correct" way of fixing this, you need to tell it how to update the cache for the Subscriptions type, if it gets to complicated we can keep the `network- only` fix, but this solution makes the site faster. The apollo client dev tools help with this.

cesarvarela (approved):

Issue #3590, opened 2025-05-29
Staging: Page blanking on certain incidents


lmcnulty:
For example: https://staging-aiid.netlify.app/cite/869/

The error is:

```
TypeError: Cannot read properties of null (reading '0')
On Citation.js

const editorFirstNameInitial = first_name[0] + '.';
```

Pablo found that "it seems it fails when the first editor doesn’t have a first name"

PR #3600, opened 2025-06-03
Restore Sean as default editor

lmcnulty:
Currently the page crashes if no editor is found, which happens in staging because it's missing users from the database https://github.com/responsible-ai-collaborative/aiid/issues/3590. Previously the default editor displayed in the citations was Sean. Although ideally we should have the actual editor associated with each incident, we don't want the page to crash in situations like the above. So this changes it back.


cesarvarela:
we decided to update the restore staging script instead

PR #3625, opened 2025-06-11
Fix checklist risk updates

lmcnulty:
The issue was that `value.risks` in `updateRisk()` was coming up as an empty array, so updates were overwriting previous updates. Instead the array of risks has to be passed as a parameter to the function.


pdcp1 (approved):
LGTM

Issue #3643, opened 2025-06-24
Incident 835 page crashes, affects incidents 834 + 836 (orig: Broken link for incident 835)


mh2171:
When navigating to incident 835 via Table View > Incident ID = 835, I get a blank page when clicking on the incident link.

kepae:
This is actually a broken behavior on the side page for incident 835, and likely not to do with the table.

https://incidentdatabase.ai/cite/835/

It also breaks with incidents 834 and 836 -- which I suspect query incident 835 as well as part of the navigation/fetching.

Great catch! We're on it.

PR #3644, opened 2025-06-24
Delete empty translations

pdcp1:
We found that some reports have empty translations that cause this error:
- https://github.com/responsible-ai-collaborative/aiid/issues/3643

```
Error in function Citation in ./src/components/cite/citationTypes/Citation.js:27
Cannot read properties of undefined (reading 'submitters')

./src/components/cite/citationTypes/Citation.js:27
```

> 27 |   let submitterCite = getFormattedName(docs[0]['submitters'][0]);
     |                                               ^
  28 |
  29 |   const [retrievalString, setRetrievalString] = useState('');
  30 |

### DIFF
--- a/site/gatsby-site/migrations/2025.06.24T12.00.00.delete-empty-translations.js
+++ b/site/gatsby-site/migrations/2025.06.24T12.00.00.delete-empty-translations.js
@@ -1,5 +1,3 @@
-const languages = require('../i18n/config.json');
-
 /**
  *
  * @param {{context: {client: import('mongodb').MongoClient}}} context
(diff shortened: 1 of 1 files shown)


cesarvarela (approved):
+1 to Kevin's question

pdcp1:
@kepae The empty translations make the report not meet the “isCompleteReport” condition (on `site/gatsby-site/src/utils/variants.js`):
```
module.exports.isCompleteReport = (report) => {
  return (
    report.title &&
    report.title != '' &&
    report.url &&
    report.url != '' &&
    report.source_domain &&
    report.source_domain != ''
  );
};
```
So the incident loads with no reports on it. And since docs is the collection of the incident reports, docs[0] is undefined, because docs is empty.
In conclusion, the exception is because the incident report list is empty, but the origin of this problem is the empty report translation.

PR #3747, opened 2025-09-17
Reorder submit tests

lmcnulty:
The tests that are currently timing out on staging are all close to the top – I want to see what happens if we reorder them – will the ones moved to the top time out instead?


smcgregor:
I think these timeouts are now resolved.

PR #3787, opened 2025-12-05
close API unless logged in

smcgregor:


smcgregor:
Mitigated another way.

PR #3795, opened 2025-12-11
Push to Production – 2025-12-11

smcgregor:
* Increase window length on rate_limit


smcgregor:
@kepae I confirmed a few Discover queries work before it starts dropping with 429s. This PR should fail some tests with a 429 error.

kepae (approved):

PR #3798, opened 2025-12-12
Remove push trigger from deploy-only workflow

kepae:
Otherwise is executes on every push. Mea culpa.


kepae:
It's also not obvious that an empty `push:` object will be interpreted as "run on every push." :-)

PR #3806, opened 2026-01-25
Update report count in search box

smcgregor:


smcgregor:
I checked the failing test and it is the one that is passing in staging after skipping it. Merging this...

PR #3807, opened 2026-01-25
Skip CSETv0 fields list rendering test

smcgregor:
Skip test for rendering CSETv0 fields due to header issue.


smcgregor:
@kepae merging this to staging since it seems to have resolved the build testing issue and I have a few other changesets staging through now.

PR #3808, opened 2026-01-25
Add Time Magazine to the landing page

smcgregor:


smcgregor:
I am making some of these changes just so I can rebuild Prismic and see how it looks, but these are also good changes to have...

smcgregor:
Tests pass and this is a minor update of wording. Merging!

PR #3847, opened 2026-04-15
workflow to cleanup translations

smcgregor:
After months of no translation issues, we had a translation API outage and empty strings landed in prod. This causes [integrity tests to fail](https://github.com/responsible-ai-collaborative/aiid/actions/runs/24434077890/job/71384371092). This workflow deletes the offending translations.

### DIFF
--- a/.github/workflows/cleanup-translations.yml
+++ b/.github/workflows/cleanup-translations.yml
@@ -1,0 +1,53 @@
+name: Cleanup Invalid Translations
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Environment to clean up (production or staging)"
+        required: true
+        default: "production"
+        type: choice
+        options:
+          - production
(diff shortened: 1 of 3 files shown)


.github/workflows/cleanup-translations.yml
```diff
@@ -0,0 +1,53 @@
+name: Cleanup Invalid Translations
+
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Environment to clean up (production or staging)"
+        required: true
+        default: "production"
```
kepae: consider defaulting to staging, even though it does have dry run

kepae (approved):
one comment! LGTM

.github/workflows/cleanup-translations.yml
smcgregor: Done.

PR #3848, opened 2026-04-15
update security version of deps

smcgregor:
Update all dependencies to their current security patch without updating major/minor versions. This needs to be thoroughly tested in staging.


smcgregor:
GitHub had a transient build error in staging. I am going to merge this into staging and roll it back in the event the build still fails

PR #3859, opened 2026-04-18
add email chunking

smcgregor:
Fixes notifications moving forward and clears out the backlog of notifications by not sending them.


kepae (approved):

smcgregor:
There is a possibility that emails will send at each production build since emails are not transactional (i.e., if one of the chunks fails after the other chunks succeed).

Let's merge this, then the migration will clear the queue and between the next two production builds post deployment we will want to check to confirm the prior production build cleared the queue after sending.

PR #3860, opened 2026-04-19
add updated incidents listing to the briefing email

smcgregor:
By request from @datherton09

### DIFF
--- a/site/gatsby-site/src/scripts/process-briefing-notifications.ts
+++ b/site/gatsby-site/src/scripts/process-briefing-notifications.ts
@@ -102,6 +102,20 @@ async function notificationsToBriefingIncidents(context: Context) {
   lastWeek.setUTCDate(now.getUTCDate() - 7); // Move to 7 days ago
   lastWeek.setUTCHours(15, 0, 0, 0); // Set to start of the day
 
+  const newIncidentIds = incidentList.map(i => i.id);
+
+  const updatedIncidentsRaw = await incidentsCollection.find({
+    date_modified: { $gte: lastWeek },
+    incident_id: { $nin: newIncidentIds },
+  }).sort({ date_modified: -1 }).limit(100).toArray();
+
+  const updatedIncidents = updatedIncidentsRaw.map(i => ({
+    id: i.incident_id,
+    title: i.title,
+    url: `${config.SITE_URL}/cite/${i.incident_id}`,
+    date_modified: i.date_modified instanceof Date ? i.date_modified.toISOString().split('T')[0] : i.date_modified,
+  }));
+
   const lastWeekDate = lastWeek.toISOString().split('T')[0]; // 'YYYY-MM-DD'
   const nowDate = now.toISOString().split('T')[0]; // 'YYYY-MM-DD'
   const lastWeekISOString = lastWeek.toISOString().split('.')[0] + 'Z';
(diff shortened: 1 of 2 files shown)


site/gatsby-site/src/scripts/process-briefing-notifications.ts
```diff
@@ -182,14 +196,14 @@ async function notificationsToBriefingIncidents(context: Context) {
...
-  const hasContentToSend = incidentList.length > 0 || newBlogPosts.length > 0 || updates.length > 0;
+  const hasContentToSend = incidentList.length > 0 || updatedIncidents.length > 0 || newBlogPosts.length > 0 || updates.length > 0;
   const shouldSendEmail = recipients.length > 0 && hasContentToSend;
 
   if (shouldSendEmail) {
     result = recipients.length;
   }
 
-  console.log(`Found ${incidentList.length} incidents, ${newBlogPosts.length} new blog posts, and ${updates.length} updates to send to ${recipients.length} users found between ${lastWeekDate} - ${nowDate}`);
+  console.log(`Found ${incidentList.length} incidents, ${updatedIncidents.length} recently edited incidents, ${newBlogPosts.length} new blog posts, and ${updates.length} updates to send to ${recipients.length} users found between ${lastWeekDate} - ${nowDate}`);
```
kepae:
```suggestion
  console.log(`Found ${incidentList.length} incidents, ${updatedIncidents.length} updated incidents, ${newBlogPosts.length} new blog posts, and ${updates.length} updates to send to ${recipients.length} users found between ${lastWeekDate} - ${nowDate}`);
```

kepae (approved):
One nitpick and LGTM but I wouldn't let this out of staging without checking it. It's hard to verify what the HTML template of the email will do just reading!

site/gatsby-site/src/scripts/process-briefing-notifications.ts
```diff
@@ -102,6 +102,20 @@ async function notificationsToBriefingIncidents(context: Context) {
   lastWeek.setUTCDate(now.getUTCDate() - 7); // Move to 7 days ago
   lastWeek.setUTCHours(15, 0, 0, 0); // Set to start of the day
 
+  const newIncidentIds = incidentList.map(i => i.id);
+
+  const updatedIncidentsRaw = await incidentsCollection.find({
+    date_modified: { $gte: lastWeek },
```
kepae, line 108: Revoking approval -- I'm thinking about this more.
I think there needs to be review of what will trigger an "updated" incident. In the past, we kicked off a huge amount of churn because migrations "updated" the content of hundreds of incidents.

As I read it now, if you touch a bunch of incident in a data-cleaning migration, this may or may not also cause the the date_modified field to be updated, and result in large amounts of unintentional churn, which would be Bad.

kepae (changes requested):

site/gatsby-site/src/scripts/process-briefing-notifications.ts
smcgregor: I thought about that...is the limit to 100 not enough of a limit?

kepae: Hmm... what's more useful to a subscriber? 10-20 vs 100?

I think 10-20 would cover editing activity but 100 feels like it could only serve to share accidental churn. Or?

smcgregor: 15 it is!

PR #3872, opened 2026-04-26
Feature single notification email

smcgregor:
This pull request made extensive use of Claude Code and will take time to qualify/merge. It provides for one MailerSend bulk request per build instead of one-per-notification-type. The rate-limit-prone fan-out is gone.        
                                                                                                                           
  Files                                                                                                                    
                  
  - New: server/emails/templates/Notifications.ts — single Liquid template with four conditional sections (newIncidents,   
  entityEvents, incidentUpdates, submissionsPromoted).
  - Modified: server/emails/templates/index.ts — registers Notifications; drops the 6 obsolete entries.                    
  - Modified: server/emails/index.ts — SendBulkEmailParams.recipients[] now accepts a per-recipient dynamicData that merges
   over the shared one inside the loop, so one bulk call can carry different content per user.                             
  - Rewritten: src/scripts/process-notifications.ts — single pass: load all 5 notification types in one query, batch-fetch 
  incidents/reports/entities/subscriptions, build a per-user UserDigest, mark processed, send one bulk, revert on failure. 
  - Modified: server/tests/notifications.spec.ts — 7 assertions updated to expect the consolidated shape.
  - Deleted: EntityIncidentUpdated.ts, IncidentUpdate.ts, NewEntityIncident.ts, NewIncident.ts,                            
  NewReportAddedToAnIncident.ts, SubmissionApproved.ts.                                                                    
                                                                                                                           
  Behavioral changes worth flagging before deploy                                                                          
                  
  - A user who's both a "new-incidents" subscriber and the submitter of a just-promoted incident now receives one email    
  containing both sections, where they previously got two.
  - The error message for failed sends changed from [Process Pending Notifications: <Type>]: ... to [Process Pending       
  Notifications]: ... (no per-type sub-prefix), since the bulk send isn't tied to one type. Anything grepping logs for the 
  old prefixes will need an update.
  - Submission-promoted recipients still get their own digest even if they have no other subscriptions — that section is   
  enough on its own.


smcgregor:
I edited https://staging-aiid.netlify.app/cite/1253 and added a new incident to staging. I am merging this to staging to see if I receive the expected notification.

PR #3908, opened 2026-05-12
Security Updates

smcgregor:
Apollo4 is at EOL. Long live Apollo5. I tested all these changes and I believe they work. I will still be prepared to roll them back.


smcgregor:
@kepae I think I can just push forward on this and be prepared to rollback

PR #3923, opened 2026-06-02
don't hit the API endpoint for the complete database when viewing the table

smcgregor:
This should save a ton of MongoDB egress since it is blowing through the AWS lambda limits and failing.


smcgregor:
This may have cut build times by ~15 minutes 🤔

PR #3935, opened 2026-06-11
Add cross-taxonomy visualization feature

kaiorta-hash:
# Add cross-taxonomy visualization feature

## Summary

Adds a new analytics page at `/apps/cross-taxonomy` for exploring patterns across
AIID taxonomy fields (CSETv1, GMF, MIT) and incident entities (developers, deployers,
harmed parties).

It provides four guided views plus a freeform explorer:

- **By Developer / By Deployer** — pick an entity, see a dashboard of how incidents
  involving it break down across sectors, failure modes, affected demographics,
  intent, risk domains, and time.
- **By Technology** — same idea, keyed on a GMF AI technology.
- **By Affected** — a single tab with a mode toggle: structured demographic basis
  (CSETv1) or named harmed-party groups.
- **Custom Explorer** — choose any two taxonomy fields as X/Y axes and cross-reference
  them as a bar, stacked-bar, line, or scatter chart, with an optional filter.

Every chart supports bar/line/donut types, a `%` normalization toggle, and an
**Incident IDs** view that lists links back to the contributing `/cite` pages, so any
number on a chart traces back to its source incidents.

## Screenshots

Screenshot 2026-06-11 at 2 31 23 PM
Screenshot 2026-06-11 at 2 31 55 PM
Screenshot 2026-06-11 at 2 32 23 PM

## Scope & self-containment

This feature is deliberately self-contained for easy review:

- **No new dependencies** (no `package.json` / lockfile changes).
- **No existing files changed except one 6-line nav entry** in `config.js`.
- All new code lives under `site/gatsby-site/`:
  - `src/pages/apps/cross-taxonomy.js` — page + data load + URL state
  - `src/utils/crossTaxonomy.js` — pure data-processing layer
  - `src/components/visualizations/{ChartCard,CrossTaxonomyChart,GuidedAnalysisTab,SearchableSelect}.js`
  - `src/utils/__tests__/crossTaxonomy.test.js` + `jest.utils.config.js`
- All processing is client-side; no backend changes.

## Testing

- **57 unit tests** cover the pure data layer (`crossTaxonomy.js`), including the
  tricky data quirks: semicolon-separated multi-values (`"Google; DeepMind"`),
  MIT numeric-prefix stripping (`"3. Misinformation"` → `"Misinformation"`),
  boolean → Yes/No normalization, and `value_json` JSON-parsing.
- Run with: `npx jest --config jest.utils.config.js src/utils/__tests__/crossTaxonomy.test.js`
  (a separate node-env config so it runs without MongoDB/DOM).
- UI was verified manually against a full local DB snapshot.

## Responses to review feedback

**Commit history:** squashed to a single clean commit containing only the feature files.

**Claude Code prompt / time in app:** this was built with Claude Code across many
iterative sessions, so there isn't a single prompt I can quote — it was developed
incrementally through many prompts that refined behavior and visuals over time. I've
spent significant time running the app locally against a full data snapshot, clicking
through every tab and chart type to validate the output and tune the visuals.

**Level of human code review:** I have personally reviewed all of the code and
understand each design decision and its tradeoffs. The design and product direction were
mine — I decided what to build, which visualizations to include, and how the interactions
should behave — and the implementation was AI-assisted, with me reviewing and validating
each part. I wrote the unit tests to lock down the data-layer edge cases. For full
transparency, the notable design decisions and known limitations are listed below.

## Design decisions (intentional)

- **Two Billboard.js wrappers.** `ChartCard` uses a small hand-rolled wrapper
  (`bb.generate` in `useLayoutEffect` + guarded `destroy`) to avoid a React 18 Strict
  Mode unmount crash in Billboard's official React wrapper; `CrossTaxonomyChart` uses the
  official wrapper where it only remounts on user action and is safe.
- **Synthetic "Time" taxonomy.** A fake `Time` namespace with a `Year` field is derived
  client-side from each incident's date and injected into the same arrays as real
  classifications, so time flows through every data function with no special-casing.
- **Two data paths, one output shape.** Entity-based and classification-based filtering
  both return an identical `{ counts, total, incidentCount, incidentIdsByValue }` shape,
  so a single `ChartCard` renders both without knowing the source.

## Known limitations

- The incidents query uses a hardcoded `limit: 9999`. This is fine at the current
  incident count but would need pagination at larger scale.
- The page issues three large unpaginated `no-cache` queries on load (taxa, all
  classifications, all incidents) — relevant to the GraphQL rate-limiting already being
  considered.
- Dynamic chart titles are translated at runtime via `t(variable)`, so they won't be
  picked up by automated i18n string extraction; full localization would need manual key
  registration. Static UI chrome uses `<Trans>`.
- Automated tests cover the data layer only; the React components and page are verified
  manually.
- The field selector is a custom combobox with partial (not full WAI-ARIA) combobox
  semantics.
- Extra Custom Explorer graph panels are not URL-shareable (only the first panel writes
  to URL params).

## Open design questions

**Navigation placement.** The `config.js` change adds Cross-Taxonomy as its own
top-level nav item right after "Taxonomies". It could alternatively be folded into the
existing Taxonomies page as a second view — happy to go either way based on maintainer
preference.

## Future directions (feedback welcome)

A few enhancements I'm considering and would value maintainer input on before building:

- **Grouping / filtering the field selector.** The field picker currently lists every
  categorical field in one flat list, which is a lot to scan, and some fields only have
  one or two annotated incidents. I'd like to group fields by descriptors (optgroup-style)
  and optionally hide or de-emphasize very-low-coverage fields so the most useful
  dimensions surface first. Open question: should low-coverage fields be hidden, or just
  labeled with their incident count?
- **Reference/overlay series.** An option to draw a comparison line behind a chart — e.g.
  the overall distribution as a faint baseline behind the currently filtered selection —
  so users can see how a slice compares to the database as a whole.
- **Selectable time granularity.** The synthetic Time taxonomy currently derives only the
  year (first four characters of each incident's date). Making the bucket selectable
  (year / month / full date) would let users zoom the temporal resolution.

### DIFF
--- a/site/gatsby-site/config.js
+++ b/site/gatsby-site/config.js
@@ -87,6 +87,12 @@ const config = {
       { title: 'List View', label: 'list', url: '/summaries/incidents/', items: [] },
       { title: 'Entities', label: 'entities', url: '/entities/', items: [] },
       { title: 'Taxonomies', label: 'taxonomies', url: '/taxonomies/', items: [] },
+      {
+        title: 'Cross-Taxonomy Visualizations',
+        label: 'cross-taxonomy',
+        url: '/apps/cross-taxonomy/',
+        items: [],
+      },
       { title: 'Random Incident', label: 'random', url: '/random/', items: [] },
       { title: 'Submit Incident Reports', label: 'submit', url: '/apps/submit/', items: [] },
       { title: 'Risk Checklists', label: 'checklists', url: '/apps/checklists/', items: [] },
--- a/site/gatsby-site/jest.utils.config.js
+++ b/site/gatsby-site/jest.utils.config.js
@@ -1,0 +1,11 @@
+/**
+ * Minimal Jest config for running pure utility function tests.
+ * Separate from the main jest.config.ts which requires MongoDB setup.
+ */
+module.exports = {
+  testEnvironment: 'node',
+  transform: {
+    '\\.[jt]sx?$': ['babel-jest', { presets: ['@babel/preset-env'] }],
+  },
+  testMatch: ['**/src/utils/__tests__/**/*.test.js'],
+};
--- a/site/gatsby-site/package.json
+++ b/site/gatsby-site/package.json
@@ -144,6 +144,7 @@
     "serve": "netlify dev --target-port 9000 --port 8000 -d public --offline",
     "test:e2e": "npx playwright test playwright/e2e-full --ui",
     "test:e2e:ci": "DEBUG=pw:webserver npx playwright test $TEST_FOLDER --shard=$SHARD_INDEX/$SHARD_TOTAL",
+    "test:utils": "jest --config jest.utils.config.js",
     "test:api": "jest --setupFiles dotenv/config --runInBand --forceExit",
     "test:api:ci": "jest --runInBand --forceExit",
     "codegen": "graphql-codegen --config codegen.ts",
--- a/site/gatsby-site/src/components/visualizations/ChartCard.js
+++ b/site/gatsby-site/src/components/visualizations/ChartCard.js
@@ -1,0 +1,454 @@
+import React, { useMemo, useState, useLayoutEffect, useRef } from 'react';
+import { Card } from 'flowbite-react';
+import bb, { bar, line, donut } from 'billboard.js';
+import { Trans, useTranslation } from 'react-i18next';
+
+// Color palette matching the AIID taxonomy page style
+const COLORS = [
+  '#3b82f6', // blue
+  '#f59e0b', // amber
+  '#10b981', // emerald
+  '#ef4444', // red
+  '#8b5cf6', // violet
+  '#6b7280', // gray
+  '#ec4899', // pink
+  '#14b8a6', // teal
+  '#f97316', // orange
+  '#06b6d4', // cyan
+  '#84cc16', // lime
+  '#a855f7', // purple
+  '#78716c', // stone
+  '#e11d48', // rose
+  '#0ea5e9', // sky
+  '#d946ef', // fuchsia
+  '#eab308', // yellow
+  '#22c55e', // green
+  '#64748b', // slate
+  '#dc2626', // red-dark
+];
+
+// Thin wrapper instead of @billboard.js/react, whose destroy() can run against
+// a DOM node React already removed (Strict Mode) and throw. Regenerates only
+// when chartKey changes.
+export function BillboardChart({ options, chartKey }) {
+  const elRef = useRef(null);
+
+  useLayoutEffect(() => {
+    const el = elRef.current;
+
+    if (!el) return;
+    const instance = bb.generate({ bindto: el, ...options });
+
+    return () => {
+      try {
+        instance.destroy();
+      } catch (_) {
+        /* node already detached by React; nothing to clean up */
+      }
+    };
+  }, [chartKey]); // options are stable within a given key
+
+  return <div ref={elRef} />;
+}
+
+const TYPE_OPTIONS = [
+  { value: 'bar', label: 'Bar' },
+  { value: 'line', label: 'Line' },
+  { value: 'donut', label: 'Donut' },
+];
+
+export default function ChartCard({
+  title,
+  subtitle,
+  counts,
+  chartType = 'bar',
+  maxItems = 15,
+  incidentCount,
+  sortBy = 'count',
+  incidentIdsByValue,
+}) {
+  const { t } = useTranslation();
+
+  const [activeType, setActiveType] = useState(chartType);
+
+  const [showPct, setShowPct] = useState(false);
+
+  // When true, the chart is replaced by a list of the contributing incidents,
+  // grouped by the same values shown on the chart.
+  const [showIds, setShowIds] = useState(false);
+
+  const hasData = !!counts && counts.size > 0;
+
+  const canNormalize = activeType !== 'donut' && incidentCount > 0 && !showIds;
+
+  const hasIds = incidentIdsByValue && incidentIdsByValue.size > 0;
+
+  const sortedEntries = useMemo(() => {
+    if (!counts) return [];
+    const entries = Array.from(counts.entries());
+
+    if (sortBy === 'key') {
+      entries.sort((a, b) => a[0].localeCompare(b[0]));
+      // Keep the end of the ascending sort so the most recent years survive
+      // the cap.
+      return entries.slice(-maxItems);
+    }
+    entries.sort((a, b) => b[1] - a[1]);
+    return entries.slice(0, maxItems);
+  }, [counts, maxItems, sortBy]);
+
+  const typeToggle = (
+    <div className="flex gap-1 shrink-0 flex-wrap justify-end">
+      {TYPE_OPTIONS.map(({ value, label }) => (
+        <button
+          key={value}
+          type="button"
+          disabled={!hasData}
+          onClick={() => {
+            setActiveType(value);
+            setShowIds(false);
+            if (value === 'donut') setShowPct(false);
+          }}
+          className={`text-xs px-2 py-0.5 rounded border transition-colors ${
+            !hasData
+              ? 'border-gray-100 text-gray-300 cursor-not-allowed'
+              : activeType === value && !showIds
+              ? 'bg-blue-50 border-blue-300 text-blue-700 font-medium'
+              : 'border-gray-200 text-gray-400 hover:text-gray-600 hover:border-gray-300'
+          }`}
+        >
+          {t(label)}
+        </button>
+      ))}
+      <button
+        type="button"
+        onClick={() => canNormalize && setShowPct((p) => !p)}
+        className={`text-xs px-2 py-0.5 rounded border transition-colors ${
+          !canNormalize
+            ? 'border-gray-100 text-gray-300 cursor-not-allowed'
+            : showPct
+            ? 'bg-blue-50 border-blue-300 text-blue-700 font-medium'
+            : 'border-gray-200 text-gray-400 hover:text-gray-600 hover:border-gray-300'
+        }`}
+      >
+        %
+      </button>
+      {hasIds && (
+        <button
+          type="button"
+          onClick={() => setShowIds((s) => !s)}
+          className={`text-xs px-2 py-0.5 rounded border transition-colors ${
+            showIds
+              ? 'bg-blue-50 border-blue-300 text-blue-700 font-medium'
+              : 'border-gray-200 text-gray-400 hover:text-gray-600 hover:border-gray-300'
+          }`}
+        >
+          {t('Incident IDs')}
+        </button>
+      )}
+    </div>
+  );
+
+  if (!counts || counts.size === 0) {
+    return (
+      <Card>
+        <div className="flex items-start justify-between gap-2">
+          <div>
+            <h3 className="text-base font-semibold">{t(title)}</h3>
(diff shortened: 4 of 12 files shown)


smcgregor:
Thanks for the pull request, this is relevant and useful. One quick question before diving into the details: Are you confident that the queries being issued to the DB are minimizing DB load? There are two dimensions to this. (1) are the queries receiving only data that is necessary for the plot; and (2) are the queries made against indices on the server side that can return information efficiently? This will determine important elements of feature cost and resource contention.

kaiorta-hash:
Hi,

How the page fetches data:

On load, three queries fire in parallel: one for the taxonomy schemas, one for all classification annotations, and one for all incidents with their entity fields. The taxa query is tiny and not a concern. The other two are worth noting.

Classifications query:

We reuse the shared FIND_CLASSIFICATION query from src/graphql/classifications.js with an empty filter (filter: {}), which tells the database to return all ~1,782 classification documents. The server's GraphQL layer does use getMongoDbProjection to limit the returned fields to exactly what the query selects, however there are two issues:

The first is that the shared query selects notes and reports { report_number }, neither of which this feature uses. That's a pre-existing over-fetch baked into the shared query. I can't change it without touching code outside the feature scope.

The second is more actionable: the publish: true filter currently runs client-side in JavaScript after all documents are already downloaded. Unpublished draft annotations are transferred across the wire and then discarded in the browser. I could push that filter to the server:

variables: { filter: { publish: { EQ: true } } }

This would reduce the payload for non-admin users and is an improvement I can make within the feature scope.

Because we're fetching the full collection with no selective WHERE clause, indices on individual fields don't come into play on the initial fetch. A server-side publish: true filter would benefit from an index on publish if one exists on the collection, but the primary win is just reducing data transferred.

Incidents query:

The projection here is tight, we request only incident_id, date, and the three entity-ref fields. Each incident document in MongoDB is much larger than that (title, description, reports, edit history), so the projection keeps the payload lean despite fetching all ~1,425 incidents.

The limit: 9999, skip: 0 pattern with no filter is a full collection read, which means no index benefit regardless of what's indexed. That said, this is a pre-existing pattern used across the codebase for bulk data needs and isn't something we introduced. The rate limiter flagged in the PR description is the right mitigation for the long-term growth concern here.

What I'd propose to change:

The one concrete improvement within scope is moving the publish filter server-side on the classifications query. Everything else is either a pre-existing codebase pattern or a fundamental requirement of the feature -- arbitrary cross-taxonomy analysis requires all classification data client-side since we can't know in advance which field combinations the user will select.

I considered per-query lazy fetching but it would be net-worse for an exploration tool where users are expected to try multiple field combinations. Each new selection would add a DB round trip, whereas the current approach pays the cost once on load and all subsequent interactions are instant client-side computation. The data volume at roughly 1,800 classification documents with field projection is small enough that eager loading is the appropriate tradeoff. If the dataset grows significantly, lazy loading by namespace (fetch CSETv1 documents on first CSETv1 field selection, cache them, etc.) would be a reasonable future migration path.

Happy to make that filter change before merge if that addresses the concern. Let me know if you'd like to discuss further.

Thanks,
Kai

smcgregor:
Can you characterize how you (the human) have instrumented and examined the performance impacts of the pull request? The LLM assessments in this regard are not very reliable.

I am thinking we may want to require a login for this feature so we can avoid getting hammered, but I would prefer for it to be open if possible.

smcgregor:
For more context, training coding agents to produce more efficient code is only a recent development: https://swefficiency.com/

kaiorta-hash:
Answering both questions above — DB load and human-verified performance.

## What changed

Since this was last discussed, the data layer was reworked: **all data now loads once at build time** via the page's Gatsby query and ships as static files. The three runtime queries discussed in June no longer exist — the DB is read once per *deploy*, not once per *visitor*, so query indices and rate limiting are no longer relevant to this page's load profile.

Two concrete things in this pass:

- **Build query now excludes `*_Annotator` classification namespaces** (never chartable — the UI already filtered them client-side, but they were still being downloaded). Cut the page-data payload 47%: 7.0 → 3.7 MB raw, 0.70 → 0.44 MB gzipped.
- **Found and fixed a per-interaction re-download.** The site-wide `QueryParamProvider` routes URL param updates through Gatsby's `navigate()`, which re-fetches a page's entire page-data.json whenever its query string changes — so every dropdown selection, tab switch, or chart-type toggle was re-downloading the full ~377 kB payload. (`/apps/discover` has the same underlying pattern, but its page-data is 0.3 kB so it's invisible there — this feature's much larger payload is what made it show up in the Network panel.) Replaced with a small `history.pushState`-based hook scoped to this page; URLs, share links, and back/forward behavior are unchanged, but interactions now cost zero network. Screenshots below.

Given zero runtime load, I'd suggest the page can stay open rather than login-gated. One trade-off to flag: data is now as fresh as the last deploy, not live which I think is the right call for a curated dataset like this.

## How I instrumented performance (by hand, not the agent)

Production build only (`gatsby clean && gatsby build`, served with `gatsby serve`) — dev-server numbers are meaningless for this, since dev ships uncompressed assets plus hot-reload machinery no visitor gets. Chrome DevTools Network + Performance tabs, `/taxonomies/` as the control page throughout (an existing page already served with no login gate), cache disabled for load tests and enabled for interaction tests.

| Metric | `/apps/cross-taxonomy/` | `/taxonomies/` (control) |
|---|---|---|
| Runtime DB/API requests (full load + interaction session) | **0** | 0 |
| Network cost per interaction (dropdown/tab/chart change) | **~0** — 0.4 kB 304 revalidations only. Was ~377 kB per click before the fix above | n/a |
| Own page-data.json (gzipped) | **376 kB** | 759 kB |
| Cold load, total transferred | **3.2 MB** | 5.8 MB |
| Load time @ Slow 4G throttle, cold cache | **4.54 s** (deep link, incl. a rendered chart) | 4.56 s |
| INP / CLS while toggling chart types | **85 ms / 0** | — |
| Main-thread busy time, 14 s interaction recording | **~545 ms (~96% idle)** | — |

The per-interaction refetch above was found *from* this instrumentation — watching the Network panel while clicking through the UI, not from a static read of the code.

## Bug/correctness pass, same session

- Chart values independently recomputed from the raw build-time data (not the UI's own output) — labels, counts, and incident IDs matched exactly for every case checked.
- ~30 scripted checks across all 5 tabs, every chart type, URL edge cases (stale/garbage params, out-of-range modes, unknown fields), panel add/remove, share-link round-trip.
- Production-build console output compared against `/taxonomies/`: identical, zero feature-specific errors, zero hydration mismatches on hard-loaded deep links.
- Fixed along the way: a dropdown that couldn't be reopened by clicking after a selection, a chart axis title colliding with rotated tick labels, per-chart sample-size warnings that previously only checked the tab-level total (so a chart built on 3 incidents could sit under an unflagged "N = 222" header), and a truncation cap that silently dropped values with no indication.

## Notes for maintainers

- `config.js`'s nav entry is a fallback only — production reads the sidebar from Prismic, so a Prismic sidebar item is still needed for the page to be discoverable.
- No dependency changes in this pass either; still self-contained to the files already listed above.

smcgregor:
@kaiorta-hash Can you characterize how you (the human) have reviewed and updated this pull request?

kaiorta-hash:
Hey Sean,

I take it you're not exactly sold on AI assisted contributions, I get where that's coming from. I've seen here at Berkeley that if I don't learn to use these tools well I'll fall behind, so I've been trying to figure out how to use them responsibly instead of just avoiding them. Perhaps the way I'm using it right now is a little too over-reliant, and I apologize for that.

To be transparent, I have two big projects going on that have made it hard to put a ton of time into this, and there was a lot of the underlying knowledge for this feature that I simply did not have going in. But I really do care about AI safety and ethics and I love what you all do here, so I put in the time and skills I had, and this is what came out of it.

Before I get into the code side of things, I want to be clear about the feature itself. The entire design was mine, the tabs, every feature, what gets compared against what, how the visualizations should work, all of it. I went through a lot of back and forth until it matched what I actually wanted, not just whatever came out first. I'm genuinely proud of how it turned out and I think it's a strong addition to the site.

This last push was my call from start to finish too, what changed, what got tested, what shipped, and what got cut. I actually pulled a couple things out of this update on purpose, a scatter chart type that worked but wasn't fully baked, and an ARIA tabs implementation that was only half done, because I'd rather cut something than ship it half finished. The performance numbers in my last comment were me at the keyboard, I built the production bundle myself and worked through Chrome DevTools with /taxonomies/ as the control the whole way through. The per interaction refetch I mentioned, I caught that myself, I noticed the same page data request kept showing up in my Network tab while I was clicking around and that's what led to finding the bug, it wasn't something any automated check flagged. The actual implementation and most of the scripted functional tests were the agent's work, which I reviewed by checking results and design tradeoffs rather than reading every line.

That confidence in the design doesn't extend the same way to the code. Most of what's in this diff is display, markup and chart config. The part that's actually load bearing logic is maybe 600 lines total and I understand that part well enough to explain it. Where I'm not going to pretend to have depth is the UI implementation itself, the React and Billboard.js side of it. Honestly, I'm not going to go learn all of that for a side project I'm doing because I care about the cause, not because it needs to be on my resume. It's not the line of work I see myself going into and my time and mental resources are already being stretched. (this is not me trying to be dismissive, I'm just trying to help you understand my context to this project)

I went back and did one more self check on my own after all that, and caught two small bugs in how the URL state hook was handling browser history, both fixed now. I also rebuilt the whole branch against the newer dependency versions currently on main, since billboard.js has moved a few minor versions since I started this, just to confirm nothing breaks on merge. Everything still passes. I've spent a good amount of time clicking through the site myself too, trying different tabs and chart types, and haven't run into anything obviously broken, though I wouldn't call that exhaustive across every browser or edge case.

I'm a cognitive science major, not a developer, so I'm not going to pretend I gave the UI code the kind of review someone with your background would. I'd rather be upfront about that than oversell what I did. If there's anything specific you want me to check, test, or change, let me know and I'll do it.

smcgregor:
Hi @kaiorta-hash. AI assisted contributions are exceptional! :) They have everything I would ask for (documentation, tests, etc.), but they also make it difficult to know how much additive human thought lives on top of the AI. I am capable of prompting an AI to rework the entire codebase on a daily basis, but it is the human layer's capacity to oversee those changes and fix problems when they arise that is the gating criteria. So this part is what I was looking for: `I'm a cognitive science major, not a developer, so I'm not going to pretend I gave the UI code the kind of review someone with your background would. I'd rather be upfront about that than oversell what I did. If there's anything specific you want me to check, test, or change, let me know and I'll do it.` This tells me I should treat your PR as a highly sophisticated prototype that will be mine to oversee if/when I merge to production. Does this make sense?

kaiorta-hash:
Hi @smcgregor. 

Yes, that makes sense. 

I want to ask directly: what's driving your hesitation right now? Is it the ongoing burden of maintaining code you can't fully vouch for, the current state of the UI implementation itself, or something else? If there's a specific piece you'd want reviewed, tested, or cut before this is mergeable, tell me and I'll work on it.

I'd also rather know now if this isn't a good fit for production as is, so I can plan accordingly. Either way, thanks for taking the time to go through this with me.

Kai

smcgregor:
The functionality looks great. If no human has loaded the code into their mental state and reconciled it with the rest of the codebase, then I have to do that before merging. This also likely needs to wait until after #4058 lands due to traffic concerns and the high likelihood this will hammer the API in the absence of some controls.

kaiorta-hash:
Sounds good, keep me in the loop if anything changes. If you have any sense of rough timing once #4058 lands, that'd help me plan for recruiting on my end, no pressure either way.

PR #3942, opened 2026-06-16
feature : excel export pipeline

jayparmar16:
## Summary

Adds an automated pipeline that produces a downloadable Excel export of the full AIID dataset and updates the `/research/snapshots` page to surface it.

## Why

Researchers, journalists, and analysts who want to explore the database often do not want to write GraphQL queries or work with raw MongoDB dumps. A single Excel file with incidents, reports, entities, and taxonomy classifications joined together covers that use case without any API or database knowledge required.

### New: `site/excel-export-pipeline/`

A Python pipeline with 6 modules, each mapping to a documented maintenance step:

| Module | What it does |
|---|---|
| `download.py` | Scrapes the snapshots page, downloads the latest `.tar.bz2`, extracts BSONs |
| `schema_check.py` | Validates expected columns are present before processing starts |
| `load_data.py` | Loads 5 BSON collections into DataFrames, flattens taxonomy attributes |
| `clean.py` | Normalises each source; taxonomy-specific hooks registered in `CLEANING_HOOKS` |
| `build_dataset.py` | Left-joins all taxonomies onto the incident spine (one row per incident) |
| `export_excel.py` | Writes a styled 5-sheet workbook |

Output: **5 sheets** -- Incidents, Reports, Entities, Data Dictionary, Coverage Map.

Config is YAML-driven (`config.yaml`). Column mappings, taxonomy namespaces, style groups, and output paths are all configurable without touching code. See [`MAINTENANCE_GUIDE.md`](site/excel-export-pipeline/MAINTENANCE_GUIDE.md) for how to add columns, taxonomies, or cleaning hooks.

### New: GitHub Actions workflow

`.github/workflows/excel-export-pipeline.yml` runs every Monday at 10:00 UTC and on manual dispatch. It builds the workbook and uploads it to Cloudflare R2 as `AIID_Excel_Export-YYYYMMDD.xlsx`.

Required repository secrets:
- `CLOUDFLARE_R2_ACCOUNT_ID`
- `CLOUDFLARE_R2_WRITE_ACCESS_KEY_ID`
- `CLOUDFLARE_R2_WRITE_SECRET_ACCESS_KEY`
- `CLOUDFLARE_R2_BUCKET_NAME`

### Updated: `/research/snapshots` page

- Two-column layout: snapshot downloads on the left, Excel export on the right. Stacks vertically on mobile.
- `createBackupsPage.js` now always renders (with an empty state) even when R2 is unreachable, and separates Excel exports (`AIID_Excel_Export-*.xlsx`) from database backups (`backup-*`).
- `backups.js` shows build date, file size, and download link for each export.

### Updated: `site/db-backup/bin/cloudflare_operations.py`

Added an optional `--content_type` CLI argument (default: `application/x-bzip2`). Excel uploads require `application/vnd.openxmlformats-officedocument.spreadsheetml.sheet`.

## How to test locally

```bash
python -m venv .venv
source .venv/bin/activate        # Windows: .venv\Scripts\Activate.ps1
pip install -r site/excel-export-pipeline/requirements.txt
python site/excel-export-pipeline/main.py
# Output written to: output/AIID_Excel_Export.xlsx
```

The pipeline downloads the latest public snapshot on first run and caches it locally under `snapshot_dir` (set in `config.yaml`). Subsequent runs reuse the cached archive.

## Example output

An example workbook generated using the pipeline can be found here: https://github.com/jayparmar16/aiid/releases/tag/ExcelExportSample


smcgregor:
This is getting closer, but it still needs changes.

The blocker is the screen scraping (BeautifulSoup) and downloading logic that introduces failure/friction points. What if the backup logic and file structure changes? What if the backup fails to generate one week? What if the excel generation completes before the weekly backup is generated? How do we date the excel artifact when it is generated from a potentially stale database backup artifact? It is also a lot more complicated than just using the [standard connection](https://github.com/responsible-ai-collaborative/aiid/blob/main/.github/workflows/db-daily-backup.yml) to MongoDB and downloading the data when generating the Excel file. That doesn't have the same staleness and maintenance issues.

Change requested: Either make the excel generation happen immediately after generating the full snapshot or (**preferred**) look at the DB daily backup script and use the same method to get the database live data directly.

Finally, there is a merge conflict that needs to be resolved before this will be mergeable.

So one must-change (don't screen scrape) and one process need (resolve merge conflict).

jayparmar16:
@smcgregor Thanks for pointing it out! I have addressed the changes in the commits above:

- This change moves the Excel export pipeline away from scraping the public website and downloading the weekly snapshot archive, and instead it reads the data straight from MongoDB. The pipeline now connects to the aiidprod database using the `MONGODB_CONNECTION_STRING` secret, which is the same Atlas connection string the existing backup workflow already relies on, and that value is supplied through the environment so it is never stored in the config file.

- The above implementation has been verified by running the pipeline against local memory mongo (`export MONGODB_CONNECTION_STRING="mongodb://127.0.0.1:4110/")`

-  The `src/download.py` module was removed along with its BeautifulSoup, requests and lxml dependencies, and dnspython was added so the Atlas connection string can be resolved.

- The `src/load_data.py` module was rewritten to query the incidents, reports, entities, duplicates and classifications collections directly, while the cleaning, joining and export logic stayed exactly the same.

-  The `config.yaml` file gained a mongo section and lost the old snapshot section, and the GitHub Actions workflow now passes the Mongo secret into the build step.

- The schema check was also reworked as I observed that the runs were failing due to EXIT CODE 2 (Schema Checks). It still compares the mappings declared in `config.yaml` against the columns that actually come back from the live collections, and it now reports any drift as warnings rather than failing, so a mapped column that has gone missing is simply left out of the output while a brand new column that is not yet mapped is flagged as a candidate to add. After printing those warnings the run carries on as normal. This behaviour makes sense for an upstream database that keeps evolving, because new taxonomies and columns are added over time and existing fields are occasionally renamed or removed, and in all of those cases the export keeps working and just tells you what changed.

- both `README.md` and `MAINTENANCE_GUIDE.md` were updated so the documentation matches all of the above.

- I have addressed the merge conflict pertaining to `site/db-backup/README.md`.

smcgregor:
Thanks @jayparmar16! I will aim to process and merge over the next week.

site/gatsby-site/src/templates/backups.js
```diff
@@ -66,41 +75,56 @@ const Backups = ({ pageContext }) => {
...
+                  {(value['Size'] / 1000000).toFixed(2)} MB &middot;{' '}
+                  <Link to={value['Url']}>{value['Key']}</Link>
+                </li>
+              ))}
+          </ul>
+
+          <h2 className="mb-0 order-4 md:order-2">Excel Export</h2>
+          <p className="order-5 md:order-4">
+            A curated Excel file, updated weekly, that combines all incident records with their
+            taxonomy classifications (CSETv0, CSETv1, GMF, MIT). Suitable for analysis and research;
```
smcgregor: It is mostly CSETv1 and MIT...

Maybe: `An Excel file, updated weekly, that combines all numbered incidents with many of their taxonomy classifications. Suitable for analysis and research; updated every Monday. No coding required — download and open directly in Excel or Google Sheets. Not all relationships can be modeled in spreadsheet form. If the spreadsheet does not provide the data you require, we recommend working with the MongoDB snapshots.`

site/gatsby-site/src/templates/backups.js
```diff
@@ -66,41 +75,56 @@ const Backups = ({ pageContext }) => {
...
+                  {format(new Date(value['CreationDate']), 'yyyy-MM-dd hh:mm a')} &middot;{' '}
+                  {(value['Size'] / 1000000).toFixed(2)} MB &middot;{' '}
+                  <Link to={value['Url']}>{value['Key']}</Link>
+                </li>
+              ))}
+          </ul>
+
+          <h2 className="mb-0 order-4 md:order-2">Excel Export</h2>
+          <p className="order-5 md:order-4">
+            A curated Excel file, updated weekly, that combines all incident records with their
```
smcgregor: @jayparmar16 todo: check on GMF columns and what is missing.

site/gatsby-site/page-creators/createBackupsPage.js
```diff
@@ -4,47 +4,47 @@ const { S3Client, ListObjectsV2Command } = require('@aws-sdk/client-s3');
...
+    const backups = allObjects
+      .filter((obj) => obj.Key.startsWith('backup-'))
+      .sort((a, b) => (a.Key < b.Key ? 1 : a.Key > b.Key ? -1 : 0));
+
+    createSnapshotsPage(createPage, backups, excelExports);
+  } catch (error) {
+    console.warn(
+      `[createBackupsPage] R2 listing failed: ${error.message}. Creating page with empty data.`
+    );
+    createSnapshotsPage(createPage);
```
smcgregor: Let's reject and raise the error rather than silently fail in the event an unexpected error occurs.

smcgregor:
Ready to merge after the inline comments are addressed. I need to read all the docs as a final check.

site/gatsby-site/src/templates/backups.js
jayparmar16: Updated

site/gatsby-site/src/templates/backups.js
jayparmar16: Latest change adds two extra columns Potential AI Technology and Potential Technical Failure to the excel export. I have left out Potential AI Goal since that column is 90%+ blank/null. @smcgregor If you would still like me to add it, happy to do so.

site/gatsby-site/page-creators/createBackupsPage.js
jayparmar16: Updated in latest commit.

jayparmar16:
> Ready to merge after the inline comments are addressed. I need to read all the docs as a final check.

The comments have been addressed. Below is another change made for better code maintenance:

- The constants that used to live at the top of export_excel.py were moved into `config.yaml`, hence the font, the banner and stripe and border colors, the per column widths, the data dictionary descriptions and the coverage map text are all configuration now. 

@smcgregor You will notice multiple inline changes to `export_excel.py` due to the above change and are targeted at using the keys from `config.yaml`. Hence, a cursory review would suffice as I have verified the pipeline completion and excel export post the change.

site/gatsby-site/src/templates/backups.js
smcgregor: For completeness sake, it is best to include everything, even if it is sparse.

smcgregor:
I reviewed the new changes. Approved! :)

If you could open one more pull request to add the one GMF missing column, that would be great. Make sure you branch from staging against since I am going to squash.

smcgregor (approved):

PR #3986, opened 2026-07-26
Move API requests behind user accounts

smcgregor:
Functionality that requires API access will display a "login" message if the user is not logged in. Otherwise the site should work as normal. Most pages still render without needing to log in although some functionality is gated. Most notably, this removes the capacity to submit incidents without being logged in.

**Why?** We are getting a tremendous amount of traffic in total and somewhere around half of that traffic is of dubious origin. This change set allows us to identify and contact people that may be integrating the database without much grace (e.g., just hammering the AIID without a cache between us). It also gives us the ability to restrict access in the future if we go down that path.

This is a giant change set that requires a line-by-line review and some time in staging to ensure everything is working properly. I made extensive use of agentic coding tools in its production.


smcgregor:
Superseded by #4058

Issue #4056, opened 2026-09-19
Move API requests behind user accounts


smcgregor:
See work in progress pull request #3986. I produced it with heavy use of agentic coding tools and I need to do more extensive human testing after using agentic coding tools to update it to track with recent changes to the staging branch.

PR #4058, opened 2026-09-21
API lockdown

smcgregor:
Changeset to lock down the API.

> **Summary.** Every root field of the GraphQL API at `/api/graphql` now requires a logged-in
> account. A request without a session is refused with HTTP 401 / `API_LOGIN_REQUIRED`; an
> account an admin has blocked is refused with HTTP 403 / `API_ACCESS_BLOCKED`. Each served
> request is counted per account per UTC day in `customData.api_usage`, admins can read usage
> and block accounts from `/admin`, and the front end explains the requirement in place (no
> redirect) wherever a feature reads or writes through the API. Statically built incident,
> report and entity pages remain readable without an account. `API_REQUIRE_LOGIN=false` is a
> rollback switch. Full design notes: `site/docs/API_ACCESS.md`.
>
> **Why.** Roughly half of API traffic is unattributable automation. Requiring an account gives
> every request an identity that can be counted, contacted, and if necessary blocked, instead
> of guessing from IP and user-agent heuristics.
>
> **What this supersedes.** This is PR #3986 rebased onto current `staging` and reviewed line by
> line, closing #4056. Changes relative to #3986:
>
> - **Merged with staging's server-side hardening.** `entities/edit.js` now checks the login
>   requirement first (login notice) and then staging's `incident_editor` role check ("Not
>   enough permissions"); `rules.ts` keeps staging's fragment-aware `notQueriesAdminData` and
>   `types/user.ts` keeps the anonymous-safe `adminData` resolver. Generated GraphQL types were
>   regenerated (`npm run codegen`) against the merged schema.
> - **Block state is admin-or-self only** (new). `api_access_blocked`, `_at` and `_reason` were
>   readable by every logged-in user (`FindUsers` is used by the history pages). A
>   `graphql-middleware` layer on the `User` type (`server/userFieldVisibility.ts`) resolves
>   them as `null` for anyone but an admin or the account itself, on every path a `User` is
>   reached through (including `incident { editors }`); `notFiltersByApiAccessFields` in
>   `server/rules.ts` refuses a non-admin filtering or sorting `user`/`users` on those fields,
>   which would otherwise leak who is blocked. Done as middleware rather than field resolvers
>   because `generateMutationFields` drops resolved fields from `UserSetType`, which would
>   have left admins unable to set a block.
> - **Similar-incident cards no longer call the API** (new). Each card fetched the incident
>   (for a translated title) and the parent incident (to show the flag control) on mount — one
>   or two requests per card for every reader. Translated titles now come with the page's
>   build-time data (`page-creators/createCitationPages.js` attaches them from
>   `allMongodbTranslationsIncidents`), so a logged-out reader of `/es/cite/…` still sees
>   Spanish titles, which the gate would otherwise have broken. The flag control is shown to
>   any logged-in user (matching the server's `flagIncidentSimilarity: allow`), not only editors
>   as #3986 had it, and is withheld from logged-out readers.
> - **Landing/submit quick-add form** (`QuickAddForm`) and the **email unsubscribe page**
>   (`/unsubscribe`) show the login notice when logged out instead of a failing button. Quick
>   add was previously open to anonymous visitors (`insertOneQuickadd: allow`); unsubscribe
>   already required ownership server-side, so only the message changes.
> - **e2e suite adapted.** The test-only `query()` helper in `playwright/utils.ts` now
>   authenticates as the seeded admin test user unless a spec passes its own `Cookie`
>   (about a hundred call sites use it to read back what the UI wrote). Specs that exercised
>   anonymous flows on now-gated pages (submit form, history pages, news digest,
>   classifications, variants, dynamic cite, admin, checklist by id, quick add, flagging,
>   live-data toggle, submissions queue) log in first; the anonymous behaviour of each is
>   covered by `playwright/e2e/apiAccess.spec.ts`. `e2e/api/graphql.spec.ts` asserts the 401.
> - **Server tests added:** gate answers with `API_LOGIN_REQUIRED` even where a role rule would
>   also refuse (proves the middleware order), and six tests for block-state visibility.
> - **Translations** (es/fr/ja) for the new UI strings; two stale comments fixed.
>
> **Operational notes.** Run the migration `2026.07.26T12.00.00.add-api-usage-indexes.ts`
> (`npm run db:migrator up`) at deploy; it creates the unique `{userId, date}` index the usage
> upsert relies on. Blocks take up to 15 s to apply (per-process cache). Access denials are not
> reported to Sentry. This is a breaking change for anonymous API consumers; the README
> documents the requirement and how to send the session cookie.


smcgregor:
I am going to qualify this PR in staging, then add API tokens on top to support programmatic access on accounts.

